PR #24, opened 2019-11-06
Component for exporting data from list components

xkostka2:
* Added generic component for exporting data from list components in formats JSON, CSV, XLS
* Resilves #10

### DIFF
--- a/apps/admin-gui/src/app/admin/components/attr-def-list/attr-def-list.component.html
+++ b/apps/admin-gui/src/app/admin/components/attr-def-list/attr-def-list.component.html
@@ -1,6 +1,7 @@
-<div class="card mt-2" [hidden]="dataSource.filteredData.length === 0 || definitions.length === 0">
+<div class="card mt-2" [class.hide-table]="exporting" [hidden]="dataSource.filteredData.length === 0 || definitions.length === 0">
   <div class="card-body">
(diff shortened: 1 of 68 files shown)


apps/admin-gui/src/app/admin/components/attr-def-list/attr-def-list.component.html
```diff
@@ -56,6 +57,8 @@
   </div>
 </div>
 
+<mat-spinner *ngIf="exporting" class="ml-auto mr-auto"></mat-spinner>
+
```
xkureck: Is it really necessary to have everywhere spinner?

xkostka2: While downloading data, third party exporter library I am using here renders all the pages one by one right in the UI, which does not look good. Because of that, I am hiding the table and putting there spinner so the user knows something is going on.

apps/admin-gui/src/app/shared/components/table-options/table-options.component.scss
```diff
@@ -0,0 +1,3 @@
+.position-right {
+  float: right;
```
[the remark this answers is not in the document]
Vojtech-Sassmann: @xkostka2 It all seems ok and this change will make it look even better. This way we can remove the gap above the tables.

PR #47, opened 2019-11-13
load principal in APP_INITIALIZE after authentication

### LINKED ISSUE #36: AdminGui - load principal in APP_INITIALIZE after authentication
* The Perun principal should be loaded in app_initialize before any component can access it. It should be loaded into AuthzResolverService.

xkureck:
* load principal in APP_INITIALIZE after authentication
resolves #36


DanoFecko:
If those 2 console.logs should be there, consider me approving

PR #56, opened 2019-11-15
translation fix for notificator service

DanoFecko:
fixing this issue: 
image

I think bug came with change in configs, when translation service couldn't get language choice soon enough to get results for notificator. 

This is just quick fix, maybe @xkureck or @Vojtech-Sassmann will find other way

### DIFF
--- a/apps/admin-gui/src/app/core/services/common/notificator.service.ts
+++ b/apps/admin-gui/src/app/core/services/common/notificator.service.ts
@@ -1,6 +1,6 @@
-import {EventEmitter, Injectable, Output} from '@angular/core';
-import {NotificationData} from '../../../shared/models/NotificationData';
-import {TranslateService} from '@ngx-translate/core';
+import { EventEmitter, Injectable, Output } from '@angular/core';
+import { NotificationData } from '../../../shared/models/NotificationData';
+import { TranslateService } from '@ngx-translate/core';
 import { RPCError } from '@perun-web-apps/perun/models';
 
 @Injectable({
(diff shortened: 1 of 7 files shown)


Vojtech-Sassmann:
@DanoFecko thanks for finding the bug. I think the real problem is that the loading of the principal has been moved to the APP_INITIALIZE phase. This setting of which language to use is already used in `app.component.ts` in the constructor. The problem is, that the App component is loaded after the initialize phase. The correct solution would probably be moving the code from app.component.ts to the very begging of the method, we are using in APP_INITIALIZE. We should also probably use only one method in APP_INITIALIZE since Lucka mentioned that there is no guarantee of the order in which those methods are run. 

To summarize: I think we merge `loadAppDefaultConfig` and `loadAppInstanceConfig` into one method. This method would do all of the things we need to initialize and in the correct order. 
1) It should set the translation service.
2) It should download the default configuration.
3) After that, it should download the instance configuration.
4) Finally, it should load the Perun principal. 

So it should look something like this: 
```
  loadConfig() : Promise<void> {
    // set translation
    translate.setDefaultLang('en');
    translate.use('en');
    
    return this.loadAppDefaultConfig().then(
      this.loadAppInstanceConfig
    ).then(
      this.loadAdditionalData
    )
  }
```

the `loadConfig` method should be the only one which we provide in app.module as APP_INITIALIZE

Vojtech-Sassmann:
I think we do not need to merge this hotfix since this bug is not in the production yet, so we can solve it the right way.

DanoFecko:
i will fix this on Tuesday

DanoFecko:
@Vojtech-Sassmann It's working now

apps/admin-gui/src/app/core/services/common/notificator.service.ts
```diff
@@ -24,13 +22,33 @@ export class NotificatorService {
...
+        return this.defaultAction = value;
+      });
+    } else {
+      return this.defaultAction;
+    }
+  }
+
+  getDefaultRpcMessage(): string {
+    if (this.defaultRpcMessage === undefined) {
+      this.translate.get('NOTIFICATOR.NOTIFICATION.DEFAULT_RPC_ERROR_MESSAGE').subscribe(value => {
```
Vojtech-Sassmann: You should use here synchronous `instant` instead of the async `get`. The return in the lamba doesn't work for the whole method, it is bounded to the given lambda.

PR #59, opened 2019-11-19
User detail - organizations

### LINKED ISSUE #39: User detail - organizations
* Implement page for user detail that will display user's organizations. The column with checkboxes should be probably hidden since I cannot think of a use for it on this page.
* You should reuse the table that is already created and extend it with a filed of `hiddenColumns` that would determine which columns should not be displayed.

xkostka2:
* Implemented tab organizations in user detail/profile
* Resolves #39


apps/admin-gui/src/app/shared/components/vo-select-table/vo-select-table.component.ts
```diff
@@ -25,7 +25,10 @@ export class VoSelectTableComponent implements OnChanges, AfterViewInit, OnInit
   selection: SelectionModel<Vo>;
 
   @Input()
-  isVoAdmin: boolean;
+  showCheckbox = false;
+
+  @Input()
+  showRecent = false;
```
xkureck: I would suggest to delete this variables and replace it with
`@Input()
  displayedColumns: string[];`
This would make the component more reusable and easy to add new column.

xkureck: It would also reduce the amount of code in this component.

PR #74, opened 2019-11-28
added script to package.json 'start-dev-admin-gui'

DanoFecko:
* script: "ng serve admin-gui --public-host=http://localhost:4200 --disable-host-check"
* when someone want to run the Admin Gui, he/she can just run the script


Vojtech-Sassmann:
Good idea. I would only suggest to use - instead of _ to make it similar to the other commands.

DanoFecko:
done

PR #77, opened 2019-11-28
Fix of the add-edit notification dialog

xkureck:
* fix of editing/creating notification


Vojtech-Sassmann:
Hi @xkureck. In cases of fixes like this, it is good to write a commit message with a description of the problem you have solved. Why has it happened, what is the solution etc. This all should be clear from the commit message without the need to look at the changed lines.

PR #84, opened 2019-12-11
Invite members to group/vo

### LINKED ISSUE #75: Invite members to group/vo
* The implementation of `AddMemberDialogComponent` now only supports inviting people who have an existing user in perun. 
* This must be extended. In case the selected `MemberCandidate` does not have a richUser, `sendInvitation ( int voId, String name, String email, String language )` this must be called instead.
* The invitation for a group is not implemented at all and it must work the same way.
* Also, when the user presses the `invite` button, he must also choose a language of the invitation.

xkureck:
* invitation to group implemented
* adding member to vo now also support adding people from external sources
* resolves #75

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/add-member-dialog/add-member-dialog.component.html
+++ b/apps/admin-gui/src/app/shared/components/dialogs/add-member-dialog/add-member-dialog.component.html
@@ -39,7 +39,7 @@
   <mat-spinner class="mr-auto ml-auto" *ngIf="processing"></mat-spinner>
   <div mat-dialog-actions *ngIf="!processing">
     <button
-      mat-flat-button
+      mat-stroked-button
       class="ml-auto"
       (click)="onCancel()">
       {{'DIALOGS.ADD_MEMBERS.CANCEL' | translate}}
--- a/apps/admin-gui/src/app/shared/components/dialogs/add-member-dialog/add-member-dialog.component.html
+++ b/apps/admin-gui/src/app/shared/components/dialogs/add-member-dialog/add-member-dialog.component.html
@@ -6,6 +6,7 @@
       <mat-form-field class="mr-2 search-field adjust-width">
         <label>
           <input matInput autocomplete="false" placeholder="{{'DIALOGS.ADD_MEMBERS.DESCRIPTION' | translate}}"
+                 (keyup.enter)="onSearchByString()"
                  [(ngModel)]="searchString">
         </label>
       </mat-form-field>
@@ -29,6 +30,7 @@
       <app-members-candidates-list
         [selection]="selection"
         [members]="members"
+        [type]="data.type"
         *ngIf="members.length !== 0">
       </app-members-candidates-list>
     </div>
@@ -39,7 +41,6 @@
     <button
       mat-flat-button
       class="ml-auto"
-      color="warn"
       (click)="onCancel()">
       {{'DIALOGS.ADD_MEMBERS.CANCEL' | translate}}
     </button>
--- a/apps/admin-gui/src/app/shared/components/members-candidates-list/members-candidates-list.component.ts
+++ b/apps/admin-gui/src/app/shared/components/members-candidates-list/members-candidates-list.component.ts
@@ -7,7 +7,7 @@ import {
   getCandidateEmail,
   getExtSourceNameOrOrganizationColumn,
   parseUserEmail,
-  parseVo
+  parseVo, parseName
 } from '@perun-web-apps/perun/utils';
 
 @Component({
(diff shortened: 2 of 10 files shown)


libs/perun/models/src/lib/Candidate.ts
```diff
@@ -10,11 +10,4 @@ export interface Candidate {
...
   titleBefore: string;
   titleAfter: string;
-  fullName: string;
-  displayName: string;
-  email: string;
-  attribute: string;
-  logins: string;
-  objectType: string;
-  status: string;
 }
```
DanoFecko: I would extend user with candidate, like it is in backend

apps/admin-gui/src/app/shared/components/dialogs/add-member-dialog/add-member-dialog.component.html
```diff
@@ -39,7 +41,6 @@ <h1 mat-dialog-title>{{'DIALOGS.ADD_MEMBERS.TITLE' | translate}}</h1>
     <button
       mat-flat-button
```
DanoFecko: please change this to mat-stroked-button

apps/admin-gui/src/app/shared/components/members-candidates-list/members-candidates-list.component.ts
```diff
@@ -83,4 +91,97 @@ export class MembersCandidatesListComponent implements OnChanges, AfterViewInit
...
+      return logins;
+    }
+
+
+	}
+
+	getLoginsForRichUser(user: RichUser): string {
+    let logins = '';
+    for (const userAttribute of user.userAttributes) {
+      if (userAttribute.friendlyName.substring(0, 15) === 'login-namespace') {
```
Vojtech-Sassmann:
```suggestion
      if (userAttribute.friendlyName.startsWith('login-namespace')) {
```

apps/admin-gui/src/app/shared/components/members-candidates-list/members-candidates-list.component.ts
```diff
@@ -83,4 +91,97 @@ export class MembersCandidatesListComponent implements OnChanges, AfterViewInit
...
+    let logins = '';
+    for (const prop in candidate.attributes) {
+      if (candidate.attributes.hasOwnProperty(prop)) {
+        if (prop.indexOf('urn:perun:user:attribute-def:def:login-namespace:') !== -1) {
+          if (candidate.attributes[prop] != null) {
+            if(logins.length > 0){
+              logins += ", ";
+            }
+            // parse login namespace
+            const parsedNamespace = prop.substring(49);
```
Vojtech-Sassmann: This looks like a magic number. It would be nice to have it in a constant so one can tell what it represents.

xkureck: Honestly I have no clue what does that mean. I just copied that part from old gui. So maybe ask Pavel, what this is all about.

Vojtech-Sassmann: I think it is the length of some attribute's namespace. Don't even say that you copied some code without knowing what it does :smile:

libs/perun/utils/src/lib/perun-utils.ts
```diff
@@ -346,3 +346,108 @@ export function getAttribute(attributes: Attribute[], attrName: string) : Attrib
...
+
+export function unescapeDN(string: string): string {
+
+  return decodeURIComponent(string.replace(/\\x/g, '%'));
+
+}
+
+export function translateIdp(name: string): string {
+
+  const orgs: Map<string, string> = new Map();
```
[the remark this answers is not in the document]
xkureck: so what are you suggesting? where to store the map?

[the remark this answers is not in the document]
Vojtech-Sassmann: If you can't come up with a solution on how to store it, you can at least rewrite this to switch (not sure if js can work with string in a switch) or to if-else. This is just really unnecessary to create a map just to call it once.

apps/admin-gui/src/app/shared/components/members-candidates-list/members-candidates-list.component.ts
```diff
@@ -38,14 +47,36 @@ export class MembersCandidatesListComponent implements OnChanges, AfterViewInit
...
 
-      this.dataSource.sortingDataAccessor = (richMember, property) => {
+      this.dataSource.sortingDataAccessor = (memberCandidate, property) => {
         switch (property) {
+          case 'status':
+            return memberCandidate.member ? memberCandidate.member.status : '';
           case 'fullName':
-            return parseFullName(richMember.richUser);
+            if (memberCandidate.richUser) {
+              return memberCandidate.richUser.firstName.toLowerCase() + ' ' +
```
Vojtech-Sassmann: There is actually a possibility that the firstName and lastName can be empty. In that case, it causes an error when calling the toLowerCase() function. Is there any reason why should you transform the names to lower case?

apps/admin-gui/src/app/shared/components/members-candidates-list/members-candidates-list.component.ts
```diff
@@ -38,14 +47,36 @@ export class MembersCandidatesListComponent implements OnChanges, AfterViewInit
...
+              memberCandidate.richUser.lastName.toLowerCase();
+            } else {
+              return memberCandidate.candidate.firstName.toLowerCase() + ' ' +
+              memberCandidate.candidate.lastName.toLowerCase();
```
Vojtech-Sassmann: Also, these values can be empty as well.

apps/admin-gui/src/app/shared/components/members-candidates-list/members-candidates-list.component.ts
xkureck: well because Adam Adamek is on top of the table and aadam adamek is on last page of the table(yes a lot of names are in lower case)

PR #86, opened 2019-12-12
Manual validation of invalid members

### LINKED ISSUE #82: Manual validation of invalid members
* Sometimes in Perun the validation of new members fails. The GUI should support manual validation of invalid members.
* The best way is to replace the invalid icon with a button containing the invalid icon.
* After pressing this button, a `setStatus` method with VALID status would be called for a given member.
* After hovering on the button it should display 'try to validate' label.

xkureck:
* add button to invalid icons, so users with 'INVALID' status can be manualy set to 'VALID'
resolves #82


apps/admin-gui/src/app/shared/components/dialogs/change-member-status-dialog/change-member-status-dialog.component.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  cancel() {
+    this.dialogRef.close();
+  }
+
+  submit() {
+    this.memberService.setStatus(this.data.member.id, 'VALID').subscribe( () => {
+      this.translate.get('DIALOGS.CHANGE_STATUS.SUCCESS').subscribe( success => {
+        this.notificatorService.showSuccess(success);
```
xkostka2: I suggest adding loading to this method as it can take a while to process the request.

PR #87, opened 2019-12-13
Remove resource

xkureck:
* add possibility to remove multiple resources from Vo, remove resource button in Vo now works
* previously you could remove only one resource from facility, now you can remove multiple

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/remove-resource-dialog/remove-resource-dialog.component.ts
+++ b/apps/admin-gui/src/app/shared/components/dialogs/remove-resource-dialog/remove-resource-dialog.component.ts
@@ -6,7 +6,6 @@ import { RichResource } from '@perun-web-apps/perun/models';
 import { ResourcesService } from '@perun-web-apps/perun/services';
 
 export interface RemoveResourceDialogData {
-  facilityId: number;
   resources: RichResource[];
 }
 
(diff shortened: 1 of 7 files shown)


apps/admin-gui/src/app/shared/components/dialogs/remove-resource-dialog/remove-resource-dialog.component.ts
```diff
@@ -36,15 +36,19 @@ export class RemoveResourceDialogComponent implements OnInit {
...
-    this.resourcesService.removeResource(this.data.resources[0].id).subscribe( () => {
+    if (this.data.resources.length === 0) {
       this.translate.get('DIALOGS.REMOVE_RESOURCES.SUCCESS').subscribe(successMessage => {
         this.loading = false;
         this.notificator.showSuccess(successMessage);
         this.dialogRef.close(true);
       });
-    });
+    } else {
+      this.resourcesService.removeResource(this.data.resources[0].id).subscribe( () => {
```
Vojtech-Sassmann: You need to handle here also the possible error. If the deletion of some resource fails, it doesn't stop loading and it doesn't close the dialog.

xkureck: well if something fails, the error notification is shown( that is something that we want)
then it complete deleting other resources (that is also something we want)
so when the error occurs you want to be loading = false or do you want to close the dialog?

Vojtech-Sassmann: You are wrong. If some call fails, this method will not be called again and it will do nothing (except showing the notification). It will not try to delete other resources and the spinner will keep spinning. If it fails, you should close the dialog and refresh the table. This way the user will see updated data if some of the resources were actually deleted.

PR #127, opened 2020-02-04
Application form save

### LINKED ISSUE #104: Application form - save button
* The goal of this task is to reimplement the saving of changes made for application forms. Now, any single change is saved. The implementation should allow multiple changes to be made and then save all of them at once. The way it is implemented now, some users could see an unfinished application form.

xkureck:
resolves #104 
There is bug when adding new application form item. Need to fix that and it should be done.


Vojtech-Sassmann:
```
setNewItems(newItems : ApplicationFormItem[]) {
    this.applicationFormItems = Object.assign([], newItems);
}
```
Add this method to the `VoSettingsApplicationFormComponent` component and use it everywhere where you assing new items.
```
this.applicationFormItems = formItems
``` 
change to: 
```
this.setNewItems(formItems)
```
You can probably also remove the custom change detection in the list component.

PR #130, opened 2020-02-06
ADMIN: ATTRIBUTES: bug fix, entityless attributes paginator

### LINKED ISSUE #116: Entityless attribute keys pagination
* When editing an attribute definition, the pagination on the page with keys is not working.

DanoFecko:
resolves #116

### DIFF
--- a/apps/admin-gui/src/app/shared/components/entityless-attribute-keys-list/entityless-attribute-keys-list.component.ts
+++ b/apps/admin-gui/src/app/shared/components/entityless-attribute-keys-list/entityless-attribute-keys-list.component.ts
@@ -64,8 +64,7 @@ export class EntitylessAttributeKeysListComponent implements OnChanges, OnInit {
   isAddButtonDisabled = false;
 
   ngOnChanges(changes: SimpleChanges) {
-    this.dataSource = new MatTableDataSource<[string, Attribute]>(this.records);
-    this.setDataSource();
+    this.ngOnInit();
   }
 
   ngOnInit() {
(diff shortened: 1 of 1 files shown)


apps/admin-gui/src/app/shared/components/entityless-attribute-keys-list/entityless-attribute-keys-list.component.ts
```diff
@@ -121,11 +120,13 @@ export class EntitylessAttributeKeysListComponent implements OnChanges, OnInit {
...
     this.records.push(rec);
-    this.ngOnChanges(undefined);
+    this.dataSource.data = this.records;
+    this.setDataSource();
     this.selection.clear();
     this.selection.select(rec);
     this.isAddButtonDisabled = true;
-    this.paginator.lastPage();
+    // .lastPage() has to be delayed, otherwise it wont go to last page if number of records modulo page size is 1
+    setTimeout(() => this.paginator.lastPage(), 100);
```
Vojtech-Sassmann: I think it makes sense to add the new record to the start of the list. Also, I am a little worried about this solution which is using the timeout.

DanoFecko:
I wish i knew what was the problem. I usually it only takes to change OnInit to AfterViewInit when paginator is broken, but here it didn't work. So i tried everything, untill this worked.

apps/admin-gui/src/app/shared/components/entityless-attribute-keys-list/entityless-attribute-keys-list.component.ts
DanoFecko: Done, there's no timeout anymore, and added attribute is on top (although after save backend puts him at the end)

PR #157, opened 2020-02-27
Remove dots from button labels

esterbando:
* I removed dots from buttons labebels.


apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -63,7 +63,7 @@
   },
   "USER_MANAGEMENT": {
     "TITLE": "Users",
-    "INPUT_PLACEHOLDER": "Enter user's name, login or email...",
```
Vojtech-Sassmann: Please do not remove these dots, they are not used on a button.

PR #214, opened 2020-04-30
Attributes list improvement

xkostka2:
* Attributes list can now be correctly set as readonly
* Translations were moved to SHARED_LIB part
* Attributes list now supports hiding columns
* Newly added functionalities were integrated to user-profile

### DIFF
--- a/apps/admin-gui/src/assets/i18n/en.json
+++ b/apps/admin-gui/src/assets/i18n/en.json
@@ -1216,13 +1216,6 @@
       "RESOURCES_TAGS_LIST": {
         "EDIT_SUCCESS": "Resource tag was successfully edited."
       },
-      "ATTRIBUTES_LIST": {
-        "FILTER": "Filter by id, name, value or description",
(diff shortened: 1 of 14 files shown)


apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1330,6 +1323,12 @@
             "SUBMIT_BUTTON": "Delete",
             "CANCEL_BUTTON": "Cancel"
           }
+        },
+        "ATTRIBUTES_LIST": {
+          "ID": "Id",
+          "NAME": "Name",
+          "VALUE": "Value",
+          "DESCRIPTION": "Description"
```
Vojtech-Sassmann, line 1331: There are some locations that you must change when you move the localizations (eg, user-settings-attributes.component.html uses `SHARED.COMPONENTS.ATTRIBUTES_LIST.FILTER`. Please, find all these locations and fix them.

PR #215, opened 2020-05-04
Disable routing feature in lists

xkostka2:
* Disable routing input parameter has been added for resources-list, vo-select-table and groups-list
* InDialog option in groups-list was replaced by disableRouting option
* Changes were also integrated to components using those features

### DIFF
--- a/libs/perun/components/src/lib/groups-list/groups-list.component.html
+++ b/libs/perun/components/src/lib/groups-list/groups-list.component.html
@@ -37,14 +37,12 @@
       <ng-container matColumnDef="name">
         <th mat-header-cell *matHeaderCellDef
             mat-sort-header>{{'SHARED_LIB.PERUN.COMPONENTS.GROUPS_LIST.TABLE_GROUP_NAME' | translate}}</th>
-        <td mat-cell *matCellDef="let group"
-            [routerLink]="inDialog ? null : ['/organizations', group.voId, 'groups', group.id]">{{group.name}}</td>
+        <td mat-cell *matCellDef="let group">{{group.name}}</td>
       </ng-container>
       <ng-container matColumnDef="description">
         <th mat-header-cell *matHeaderCellDef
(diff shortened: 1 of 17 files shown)


libs/perun/components/src/lib/groups-list/groups-list.component.html
```diff
@@ -62,6 +60,7 @@
       <tr
         mat-row
         *matRowDef="let group; columns: displayedColumns;"
+        [routerLink]="disableRouting ? null : ['/organizations', group.voId, 'groups', group.id]"
         class="dark-hover-list-item">
```
Vojtech-Sassmann, line 64: Please, duplicate the `tr` element, where one instance has the routerLink set and the other one doesn't. Then by using the NgIf decide, which one to show. Because with this implementation, if the routing is disabled, the element is still "clickable" and it doesn't look good.

libs/perun/components/src/lib/resources-list/resources-list.component.html
```diff
@@ -51,7 +51,7 @@
       <tr
         mat-row
         *matRowDef="let resource; columns: displayedColumns;"
-        [routerLink]="['/facilities', resource.facilityId, 'resources', resource.id]"
+        [routerLink]="disableRouting ? null : ['/facilities', resource.facilityId, 'resources', resource.id]"
```
Vojtech-Sassmann, line 54: same here

libs/perun/components/src/lib/vo-select-table/vo-select-table.component.html
```diff
@@ -41,7 +41,7 @@
       <tr
         mat-row
         *matRowDef="let vo; columns: displayedColumns;"
-        [routerLink]="['/organizations', vo.id]"
+        [routerLink]="disableRouting ? null : ['/organizations', vo.id]"
```
Vojtech-Sassmann, line 44: same here

PR #219, opened 2020-05-12
Handling error when perunPrincipal.user is null

### LINKED ISSUE #113: Handle not existing users login
* There is a possibility, that someone is trying to access Perun with an identity which is not in Perun. In that case, Perun returns `null` user in Perun principal. If that happens, the gui has to notify the users about that.

DanoFecko:
Resolves #113 

* sorry for double commit, it was a lot easier to keep to old one here


libs/perun/services/src/lib/init-auth.service.ts
```diff
@@ -40,8 +43,12 @@ export class InitAuthService {
     return this.authzService.getPerunPrincipal()
       .toPromise()
       .then(perunPrincipal => {
-        this.storeService.setPerunPrincipal(perunPrincipal);
-        this.authResolver.init(perunPrincipal);
+        if (perunPrincipal.user === null) {
+          this.dialog.open(UserDontExistDialogComponent, {height: '80px'});
```
Vojtech-Sassmann, line 47: Why is there a height limit? This height causes a scrollbar to appear.

DanoFecko: without it, dialog had lot of empty space under text. It didn't show scrollbar on my screen, but i didn't check smaller resolutions.

DanoFecko: i will play a bit with it and find solution

Vojtech-Sassmann: To reduce the padding or margin of an element, just change the CSS. Or, you can even create a new class with only those CSS attributes that you want to use.

Vojtech-Sassmann: This should not be hard at all. It is just a matter of a couple of minutes to write a CSS class and apply it.

DanoFecko:
@Vojtech-Sassmann I fixed by placing the text to the dialog content instead of title. I also disabled closing the dialog by clicking on the background and somehow I had .css file in component instead of .scss, so I changed that too.

PR #233, opened 2020-05-28
Service configuration enhancment

### LINKED ISSUE #211: Service configuration
* The goal of this task is to finish the implementation of the service configuration page.
* A user should be able to choose a relevant resource and based on that, it should display its Vo attributes and it should display a selection to choose a group that is assigned to this resource. After that, it should display a member selection. It should be possible to choose a member of the vo. 
* Some suggested improvements:
  * The selects should not be only disabled when the selection is not possible(e.g. choose a member without selecting a resource) but it should be completely hidden from the user.
  * The tabs should have in their labels the type of the entity displayed (e.g. Resource, Member).
* Also, it should be possible to give the user an option to either display all set attributes of the selected entities or to display all required attributes by some service.

DanoFecko:
Resolves #211 

* added member selection
* added possibility to edit attributes
* added entity tags after entity name (in tab header)
* selection will be hidden now when there are no options in
* attribute list component now filter core attributes, by itself (it's not needed to filter them before sending attribute array to list component)
* edit attribute component was moved to 'libs'


Vojtech-Sassmann:
Also, please fix the lint problem.

package.json
```diff
@@ -37,66 +37,66 @@
...
-    "@angular/compiler-cli": "~9.1.0",
-    "@angular/language-service": "~9.1.0",
-    "@nrwl/cypress": "^9.1.2",
+    "@angular/compiler-cli": "^9.1.9",
+    "@angular/language-service": "^9.1.9",
+    "@nrwl/cypress": "^9.3.0",
     "@nrwl/jest": "9.2.2",
     "@nrwl/workspace": "9.2.2",
-    "@openapitools/openapi-generator-cli": "^1.0.13-4.2.2",
+    "@openapitools/openapi-generator-cli": "^1.0.13-4.3.1",
```
Vojtech-Sassmann, line 87: Please, try if after this update the generated libraries are ok. If not, downgrade to the original version.

DanoFecko:
@Vojtech-Sassmann Everthing seems to be working, I fixed what you found and I made some other tweaks

PR #236, opened 2020-06-03
Export/import of attributes

Vojtech-Sassmann:
* Added new feature that enables the import of attribute definitions
(with rights) and the export.
* To export an attribute click on it in the admin section and press the
Copy for export button.
* To import an attribute, go to the admin attributes page and press the
import button and paste the copied content.

### DIFF
--- a/apps/admin-gui/src/app/admin/pages/admin-page/admin-attributes/admin-attributes.component.ts
+++ b/apps/admin-gui/src/app/admin/pages/admin-page/admin-attributes/admin-attributes.component.ts
@@ -9,6 +9,7 @@ import { filterCoreAttributesDefinitions } from '@perun-web-apps/perun/utils';
 import { AttributeDefinition, AttributesManagerService } from '@perun-web-apps/perun/openapi';
 import { PageEvent } from '@angular/material/paginator';
 import { TABLE_ADMIN_ATTRIBUTES, TableConfigService } from '@perun-web-apps/config/table-config';
+import { AttributeImportDialogComponent } from '../../../../shared/components/dialogs/attribute-import-dialog/attribute-import-dialog.component';
 
 @Component({
   selector: 'app-admin-attributes',
@@ -83,5 +84,17 @@ export class AdminAttributesComponent implements OnInit {
   pageChanged(event: PageEvent) {
     this.pageSize = event.pageSize;
     this.tableConfigService.setTablePageSize(this.tableId, event.pageSize);
+  }
+
+  onImport() {
+    const dialogRef = this.dialog.open(AttributeImportDialogComponent, {
+      width: '700px'
+    });
+
+    dialogRef.afterClosed().subscribe(value => {
+      if (value === true) {
+        this.refreshTable();
+      }
+    })
   }
 }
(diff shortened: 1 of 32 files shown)


apps/admin-gui/src/app/admin/pages/admin-page/admin-attributes/admin-attributes.component.ts
```diff
@@ -9,6 +9,8 @@ import { filterCoreAttributesDefinitions } from '@perun-web-apps/perun/utils';
 import { AttributeDefinition, AttributesManagerService } from '@perun-web-apps/perun/openapi';
 import { PageEvent } from '@angular/material/paginator';
 import { TABLE_ADMIN_ATTRIBUTES, TableConfigService } from '@perun-web-apps/config/table-config';
+import { AttributeImportDialogComponent } from '../../../../shared/components/dialogs/attribute-import-dialog/attribute-import-dialog.component';
+import { Clipboard } from '@angular/cdk/clipboard';
```
DanoFecko, line 13: unused import

apps/admin-gui/src/app/shared/components/dialogs/attribute-export-dialog/attribute-export-dialog.component.html
```diff
@@ -0,0 +1,6 @@
+<h1 mat-dialog-title>Export attribute definition</h1>
```
DanoFecko, line 1: missing translation

apps/admin-gui/src/app/shared/components/dialogs/attribute-export-dialog/attribute-export-dialog.component.ts
```diff
@@ -0,0 +1,21 @@
+import { Component, Inject, OnInit } from '@angular/core';
+import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
+import { RemoveGroupResourceDialogData } from '../remove-group-resource-dialog/remove-group-resource-dialog.component';
```
DanoFecko, line 3: unused import

DanoFecko:
Also, you could change height of the 'edit attr def' window so there won't be scroll bar with no services assigned. It's just few pixel missing. (see in pic)

image

Vojtech-Sassmann:
Hi @xkureck, I have fixed some of the problems you mentioned. However, I don't understand what you mean by `textfield instead of input`. If you meant `textarea`, I don't think it is necessary here, since this field is just for pasting the copied text.

apps/admin-gui/src/app/admin/pages/admin-page/admin-attributes/admin-attributes.component.ts
Vojtech-Sassmann: Removed.

apps/admin-gui/src/app/shared/components/dialogs/attribute-export-dialog/attribute-export-dialog.component.html
Vojtech-Sassmann: Actually, I forgot to remove this dialog, since it is not used at all. So I have removed it.

apps/admin-gui/src/app/shared/components/dialogs/attribute-export-dialog/attribute-export-dialog.component.ts
Vojtech-Sassmann: This dialog has been removed completely.

Vojtech-Sassmann:
> Also, you could change height of the 'edit attr def' window so there won't be scroll bar with no services assigned. It's just few pixels missing. (see in pic)
> 
> image

I don't think it is a good idea to set a fixed height since the content height is not always the same, and also, the heights of the display devices are not all the same. It might just cause more trouble.

Vojtech-Sassmann:
> Implementation works for me. I just found a few little details:
> 
> * Notification text seems too generic, even though it is used only here
> * After I display EditAttributeDefinitionDialog the copy button has automatically hover effect

* I don't think there is anything wrong with the notification text.
* The button has been moved so this is no longer a problem.

PR #247, opened 2020-06-08
Unify dialog configuration

### LINKED ISSUE #245: Unify dialog configurations
* Now, when a dialog is opened, we create for it a unique configuration object. This makes it difficult to edit this configuration for all dialogs. 
* The solution is to create a new method in `perun-utils.ts` that would return some basic configuration that would be possible to customize on all used places.
* The basic configuration should have the `disableClose` property set to true. This might prevent users from accidentally closing dialogs.

bodnara:
Resolves #245

* added easy way of setting default dialog config
* added method to get default dialog config
* changed all dialogs to use default config (already existing config parameters  were unchanged)


Vojtech-Sassmann:
Just please rebase the PR to fix the conflicts.

PR #253, opened 2020-06-09
Application overview line break

### LINKED ISSUE #246: AdminGui - applications overview word-break
* When viewing applications, there might be some long word values (eg SSH keys). Now, they are not displayed very nicely the content is stretched to the right side.
* This could be fixed by applying the `word-break: break-all` style to the value column.

bodnara:
Resolves #246

* Fixed displaying of long words in application detail

### DIFF
--- a/apps/admin-gui/src/app/vos/components/application-detail/application-detail.component.html
+++ b/apps/admin-gui/src/app/vos/components/application-detail/application-detail.component.html
@@ -68,7 +68,7 @@
       </ng-container>
       <ng-container matColumnDef="value">
         <th mat-header-cell *matHeaderCellDef></th>
-        <td mat-cell *matCellDef="let userData"> {{userData.value}} </td>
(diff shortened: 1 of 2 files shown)


apps/admin-gui/src/app/vos/components/application-detail/application-detail.component.html
```diff
@@ -68,7 +68,7 @@ <h2 class="page-subtitle">
       </ng-container>
       <ng-container matColumnDef="value">
         <th mat-header-cell *matHeaderCellDef></th>
-        <td mat-cell *matCellDef="let userData"> {{userData.value}} </td>
+        <td mat-cell *matCellDef="let userData" style="word-break: break-all"> {{userData.value}} </td>
```
Vojtech-Sassmann, line 71: Please, try to avoid inlining style. You can create some CSS class in the `styles.scss` which then can be reused across the whole application. You can name it something like `break-all-words`.

PR #254, opened 2020-06-09
Vo short name

### LINKED ISSUE #252: Vo short name
* The goal of this task is to add the vo shortname to the table on the vo select page table.
* Also, it should be displayed somewhere on the vo detail page, probably somewhere next to the title.

bodnara:
Resolves #252 

* added short name display in organization selection
* added short name display in organization detail

### DIFF
--- a/apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-detail-page.component.html
+++ b/apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-detail-page.component.html
@@ -10,6 +10,12 @@
       &nbsp;#{{vo.id}}
     </span>
   </h1>
+  <h1 class="page-subtitle">
+    Short name:
+    <span class="vo-text">
+    {{vo.shortName}}
+  </span>
+  </h1>
   <app-animated-router-outlet></app-animated-router-outlet>
 </div>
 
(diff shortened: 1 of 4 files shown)


apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-detail-page.component.html
```diff
@@ -6,6 +6,10 @@ <h1 class="page-title">
     <a class="vo-link" [routerLink]="['/organizations/', vo.id]">
       {{vo.name}}
     </a>
+    <br>
+    <span class="vo-text" style="margin-left: 6.5em">
```
Vojtech-Sassmann, line 10: Please, never write a style directly to an element like this. You should always prefer using some CSS classes instead. And I don't even like the position of the short name. It is nice, that it is aligned with the full name, but it will not work on all devices. It would be better if the short name would be positioned under the full name and on the left side. Also, the ID should be left where it was.

apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-detail-page.component.html
```diff
@@ -9,6 +9,10 @@ <h1 class="page-title">
     <span class="text-muted">
       &nbsp;#{{vo.id}}
     </span>
+    <br>
+    <span class="vo-text">
+      {{vo.shortName}}
+    </span>
```
xkureck, line 15: tbh this looks weird. Try something like: (please check slack, couldnt it copy here)

PR #274, opened 2020-06-12
User profile update

xkostka2:
* Change email now works correctly
* Added attribute validation into profile page
* Fixed removing external identities


xkostka2:
As for the failed request to change email, it is normal if it happens while running locally. It should work on devel.

PR #283, opened 2020-06-16
User detail fix

### LINKED ISSUE #281: User detail displays data of logged user instead of selected user
* User detail in admin section displays data of currently logged user instead of selected user

xkostka2:
* User detail now correctly displays data of the selected user
* Resolves #281


xkureck:
This problem is already solved in Dano's PR: users organizations and groups in Admin section

PR #321, opened 2020-07-02
Tables improvement

### LINKED ISSUE #295: Tables oveflowing
Remake all tables. Table content must not overflow the table container.(eg. when the screen width is small).
Also make cusor as movable if row of table have some link.

xkostka2:
* Table content now does not overflow the container, in case screen width is too small to display the table content, horizontal scrollbar appears
* Tables now indicate whether something happens after clicking on a row by changing the shape of cursor
* Table rows highlight on hover
* Resolves #295


apps/admin-gui/src/app/shared/components/facility-select-table/facility-select-table.component.html
```diff
@@ -1,47 +1,49 @@
 <div class="card mt-3" [class.hide-table]="exporting" [hidden]="dataSource.filteredData.length === 0">
   <div class="card-body">
     <perun-web-apps-table-options (exportEnd)="exporting = false" (exportStart)="exporting = true" [exporter]="exporter"></perun-web-apps-table-options>
+    <div class="overflow-auto">
```
xkureck, line 4: On this table you forget to change the cursor.

PR #340, opened 2020-07-07
Corrected probable error in Readme file

saksham93:
In Readme,
The command for running  consolidator is as below:
**nx** serve consolidator --disable-host-check --baseHref=/krb/nic/ --deployUrl=/krb/nic/ --publicHost=localhost:4200

In my opinion, it should be:
**ng** serve consolidator --disable-host-check --baseHref=/krb/nic/ --deployUrl=/krb/nic/ --publicHost=localhost:4200

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ ng serve admin-gui --public-host=http://localhost:4200 --disable-host-check
 Command to serve the consolidator:
 
 ```shell script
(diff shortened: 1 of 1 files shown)


Vojtech-Sassmann:
Hi @saksham93, since we are using the Nx in our project (https://nx.dev/angular) this command is actually ok.

PR #345, opened 2020-07-08
Change user's password

### LINKED ISSUE #303: Change user's password
Add new functionality to gui so that user can change his password. User should choose the namespace and then our gui should redirect him into the application that is changing password(ask Pavel Zlámal for that application). User should not change his password directly in our gui.

xkostka2:
* Added page where users can change their password
* Password reset component is located in shared lib to be used by other apps
* Resolves #303


libs/perun/components/src/lib/password-reset/password-reset.component.html
```diff
@@ -1,20 +1,20 @@
-<h1 class="page-title">{{'PASSWORD_RESET.TITLE' | customTranslate | translate}}</h1>
+<h1 class="page-title">{{'SHARED_LIB.PERUN.COMPONENTS.PASSWORD_RESET.TITLE' | customTranslate | translate}}</h1>
 <div class="card mt-2 table-size-fix">
   <div class="card-body">
     <table [dataSource]="dataSource" class="w-100" mat-table matSort matSortActive="value"  matSortDirection="asc"
```
xkureck, line 4: for this table you should probably implement the overflow ass well and set the some min margin as you did in all other tables

libs/perun/components/src/lib/password-reset/password-reset.component.html
```diff
@@ -1,20 +1,20 @@
-<h1 class="page-title">{{'PASSWORD_RESET.TITLE' | customTranslate | translate}}</h1>
+<h1 class="page-title">{{'SHARED_LIB.PERUN.COMPONENTS.PASSWORD_RESET.TITLE' | customTranslate | translate}}</h1>
```
xkureck, line 1: Also is here neccessarry to have there these double pipes for customTranslate and translate?

xkostka2: At the moment yes, mainly because of the user-profile app where is this component used.

PR #348, opened 2020-07-09
Perun Admin users

### LINKED ISSUE #319: Perun Admin - users
After typing some name to filter the table of users shows. The checkbox column is here for nothing. Dont show that column and instead create new column, where would be shown if the user is Person or Service(it could be showed by some icons).

bodnara:
Resolves #319 

* removed checkbox column from users table
* added icon to indicate if user is service or person in users table

### DIFF
--- a/apps/admin-gui/src/app/shared/components/users-list/users-list.component.ts
+++ b/apps/admin-gui/src/app/shared/components/users-list/users-list.component.ts
@@ -29,21 +29,19 @@ export class UsersListComponent implements OnChanges {
   private sort: MatSort;
 
   @Input()
-  hideColumns: string[] = [];
-
-  @Input()
   selection = new SelectionModel<RichUser>(true, []);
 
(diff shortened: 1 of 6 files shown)


apps/admin-gui/src/app/shared/components/users-list/users-list.component.ts
```diff
@@ -37,13 +37,16 @@ export class UsersListComponent implements OnChanges {
   @Input()
   inDialog: boolean;
 
+  @Input()
+  inAdmin = false;
+
```
xkureck, line 42: I definitely do not want another property in this list. If you want to hide some column do it with property [hideColumns].
When you touch it I want from you to actually redesign this so that users-list take instead of hideColumns - displayColumns like it is done in ApplicationsListComponent. So when situation like this happens and you add another column you dont have to change 1549645 files to hide the new column.

PR #353, opened 2020-07-10
Side menu - resource rework

### LINKED ISSUE #341: Sidemenu - resource "subpart"
In the sidemenu the resource part is now showed under the facility subpart only.
We want to also show it under the VO subpart. So the resource subpart could be shown on facility and vo subpart. When you are accessing resource from organization or group or member the resource subpart should be shown right under the vo subpart and when we want to access resource from facility subpart the resource should be shown under the facility subpart.
Please also make sure that the all links from vo/group/member are accesing the right route so the resource is shown under the vo subpart.

xkostka2:
* Resource now shows in side menu under entity it has been opened from
* Resolves #341

### DIFF
--- a/apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-detail-page.component.ts
+++ b/apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-detail-page.component.ts
@@ -1,9 +1,14 @@
-import {Component, OnInit} from '@angular/core';
-import {ActivatedRoute} from '@angular/router';
-import {fadeIn} from '@perun-web-apps/perun/animations';
-import {SideMenuService} from '../../../core/services/common/side-menu.service';
-import {SideMenuItemService} from '../../../shared/side-menu/side-menu-item.service';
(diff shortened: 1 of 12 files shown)


apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-detail-page.component.ts
```diff
@@ -32,12 +38,21 @@ export class ResourceDetailPageComponent implements OnInit {
...
         this.resource = resource;
 
-        this.facilityManager.getFacilityById(resource.facilityId).subscribe(facility => {
-          const facilityItem = this.sideMenuItemService.parseFacility(facility);
-          const resourceItem = this.sideMenuItemService.parseResource(resource);
+        if (this.route.parent.snapshot.url[0].path === 'facilities') {
+          this.facilityManager.getFacilityById(resource.facilityId).subscribe(facility => {
+            const facilityItem = this.sideMenuItemService.parseFacility(facility);
+            const resourceItem = this.sideMenuItemService.parseResource(resource, false);
+            console.log('From resource detail');
```
xkureck, line 45: just remove this log and it should be fine

PR #357, opened 2020-07-15
ISSUE 337: editing resource list

esterbando:
Resolves#337

I added column 'vo name' to resource list and hide that column where it doesn't make sense.

### DIFF
--- a/libs/perun/components/src/lib/resources-list/resources-list.component.ts
+++ b/libs/perun/components/src/lib/resources-list/resources-list.component.ts
@@ -47,7 +47,7 @@ export class ResourcesListComponent implements AfterViewInit, OnChanges {
 
   private sort: MatSort;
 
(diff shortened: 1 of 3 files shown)


libs/perun/components/src/lib/resources-list/resources-list.component.ts
```diff
@@ -93,6 +93,9 @@ export class ResourcesListComponent implements AfterViewInit, OnChanges {
   }
 
   ngAfterViewInit(): void {
+    if (window.location.href.indexOf("/organization") === -1){
```
Vojtech-Sassmann, line 96: Although this implementation might work, it is not a good way of doing this. A better way is to define a new input property called `hideColumns` and use this to specify, which columns should be hidden. Then on a place where this component is used and you want to hide some column, you can pass to this property the desired value.

This is already implemented in some of our components. For example, you can take a look at the `GroupsListComponent`.

PR #358, opened 2020-07-16
User profile update

xkostka2:


apps/admin-gui/src/assets/config/defaultConfig.json
```diff
@@ -9,7 +9,8 @@
     "oauth_post_logout_redirect_uri": "",
     "oauth_redirect_uri": "http://gui-dev.org/api-callback",
     "oauth_silent_redirect_uri": "http://gui-dev.org/silent-refresh.html",
-    "oauth_load_user_info": false
+    "oauth_load_user_info": false,
+    "scope": "openid profile perun_api perun_admin"
```
Vojtech-Sassmann, line 13:
```suggestion
    "oauth_scopes": "openid profile perun_api perun_admin"
```
I would suggest renaming this property so it has the same format like the other OAuth properties.

PR #370, opened 2020-07-22
Group name secondary regex

### LINKED ISSUE #361: Group name secondary regex
* Perun supports a secondary regex for group names. The GUI should support this as well. 
* The secondary regex has to be configurable in the GUI config. 
* When validating the group name, the default regex validation should be performed first, then the validation for the secondary regex should be done (if defined)
* There should be also an option to configure an error message, which should be displayed if any of these two validations would fail.
* For inspiration, one can look at the new changes in the old GUI: https://github.com/CESNET/perun/pull/2795

xkostka2:
* When creating group, additional name validation regex can be used
* The new validation regex and custom error message are set in the GUI config
* Create group dialog code was adjusted
* Resolves #361

### DIFF
--- a/apps/admin-gui/src/assets/config/defaultConfig.json
+++ b/apps/admin-gui/src/assets/config/defaultConfig.json
@@ -26,6 +26,8 @@
   ],
   "pwd_reset_base_url": "https://perun-dev.cesnet.cz/fed/pwd-reset/",
   "skip_oidc": false,
+  "groupNameSecondaryRegex": "^[-a-zA-Z.0-9_ ]+$",
+  "groupNameErrorMessage": "Name cannot be empty and can contain only characters a-z, A-Z, numbers, spaces, dots, '_' and '-'",
   "footer_perun_web_web": "https://perun-aai.org/",
   "footer_perun_team_web": "https://perun-aai.org/team",
   "footer_privacy_policy_web": "https://perun-aai.org/",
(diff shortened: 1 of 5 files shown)


apps/admin-gui/src/assets/config/defaultConfig.json
```diff
@@ -26,6 +26,8 @@
   ],
   "pwd_reset_base_url": "https://perun-dev.cesnet.cz/fed/pwd-reset/",
   "skip_oidc": false,
+  "groupNameSecondaryRegex": "(...).*",
+  "groupNameErrorMessage": "Name has to consist from at least 3 characters",
```
Vojtech-Sassmann, line 30:
```suggestion
  "groupNameErrorMessage": "Name can contain only a-z, A-Z, numbers, spaces, dots, '_' and '-'",
```

apps/admin-gui/src/assets/config/defaultConfig.json
```diff
@@ -26,6 +26,8 @@
   ],
   "pwd_reset_base_url": "https://perun-dev.cesnet.cz/fed/pwd-reset/",
   "skip_oidc": false,
+  "groupNameSecondaryRegex": "(...).*",
```
Vojtech-Sassmann, line 29:
```suggestion
  "groupNameSecondaryRegex": "^[-a-zA-Z.0-9_ ]+$",
```

PR #371, opened 2020-07-23
Attributes value improvement

### LINKED ISSUE #327: Attributes bug
With virtual attributes the value of the attributes is not showed.(when attributes are too long) And they cannot be edited as other normal attributes. 
Change this behavious so that these attributes can be edited(thus can be copied and viewed by all value), but these attributes cannot be checked(in the table)

xkostka2:
* Added indicator whether attribute value overflows
* Read-only values can be viewed in dialog
* Little refactor was done
* Resolves #327


libs/perun/components/src/lib/attributes-list/attribute-value/attribute-value-integer/attribute-value-integer.component.ts
```diff
@@ -19,7 +23,20 @@ export class AttributeValueIntegerComponent {
 
   @Output() sendEventToParent = new EventEmitter();
 
+  ngOnInit() {
+    if(!this.readonly){
+      this.readonly = new IsVirtualAttributePipe().transform(this.attribute);
```
Vojtech-Sassmann, line 28: It might be worth extracting the code from the pipe to the perun-utils.ts file, so you don't need to create a pipe.

libs/perun/components/src/lib/attributes-list/attribute-value/attribute-value-integer/attribute-value-integer.component.css
```diff
@@ -0,0 +1,7 @@
+.cursor-pointer {
+  cursor: pointer;
+}
+
+.overflow-ellipsis {
+  text-overflow: ellipsis;
+}
```
Vojtech-Sassmann, line 7: I think this could be declared in the main styles file, so it can be reused.

PR #372, opened 2020-07-23
Edits for facility, resource, vo, group implemented

### LINKED ISSUE #328: Change object name/description in detail of object
In Facility/Resource/Group/Vo detail there should be (probably right next to the name of entity) some edit icon which would pop some edit dialog. Should work the same as in old gui.

xpavlic:
- Resolves #328 

- edit buttons with dialogs added to facility, resource, vo and group pages


xkostka2 (approved):
image

Implementation seems fine. I would just suggest a little refactor, while refreshing, call only a relevant part of ngOnInit method.

xkureck (approved):

PR #373, opened 2020-07-23
User service identities

### LINKED ISSUE #367: Perun admin - user - settings - service identities implement
Impelement page with service identities. Should be the same as in old gui

bodnara:
Resolves #367 

* added User Service Identities page to Perun Admin
* sometimes when redirecting to page on same 'level' would cause service method errors
  so I added a in between page to reset their subscriptions, so far this is only used with user-list
* updated openapi


apps/admin-gui/src/app/shared/components/dialogs/add-user-service-identity/add-user-service-identity.component.html
```diff
@@ -0,0 +1,53 @@
...
+      (click)="onSearchByString()">
+      <mat-icon>search</mat-icon>
+    </button>
+    <mat-spinner *ngIf="loading" class="ml-auto mr-auto">
+    </mat-spinner>
+    <div *ngIf="!loading">
+      <app-users-list
+        (page)="pageChanged($event)"
+        [pageSize]="pageSize"
+        [displayedColumns]="['user', 'select', 'id', 'name', 'email', 'logins', 'organization']"
```
xkureck, line 23: same here with column user

xkostka2 (approved):

xkureck (approved):

PR #380, opened 2020-07-24
Groups changes

### LINKED ISSUE #322: Vo - groups changes
Please use slide toggle from angular materials to change the view from Tree view to list view. On one side of the toggle should be written tree view and on other side list view. It still should be remember which view user prefers.
Tree view:
- add button "copy to clipboard" which should copy to clipboard name of group. It should be right next to synchronize button.
- Three dots button - under it there are hidden two functionality. On wide screen dont use three dots button and instead there should be buttons for this two functionality next to synchronize button. (so in sum there should be 4 buttons).
- on small screen - eg. mobile. the three dots button should contain all four buttons and their functionality

List view: 
- the functionality should/could be be the same

also add functionality to the button "update name/description"

xkostka2:
* All buttons located in each row have been moved to one component, the display format is determined by the window size
* Switching between tree and list view is now done by slide toggle
* Resolves #322


apps/admin-gui/src/app/shared/shared.module.ts
```diff
@@ -429,12 +428,12 @@ import { AddMemberToResourceDialogComponent } from './components/dialogs/add-mem
...
     AddFacilityOwnerDialogComponent,
     RemoveFacilityOwnerDialogComponent,
-    GroupSyncDetailDialogComponent,
     NotFoundPageComponent,
     CreateServiceDialogComponent,
     DeleteServiceDialogComponent,
     UserResourcesComponent,
-    AddMemberToResourceDialogComponent
+    AddMemberToResourceDialogComponent,
+    NotFoundPageComponent
```
xkureck, line 436: already here

apps/admin-gui/src/app/vos/pages/group-detail-page/group-subgroups/group-subgroups.component.ts
```diff
@@ -115,4 +126,21 @@ export class GroupSubgroupsComponent implements OnInit {
     this.pageSize = event.pageSize;
     this.tableConfigService.setTablePageSize(this.tableId, event.pageSize);
   }
+
+  onMoveGroup(group: GroupFlatNode | Group) {
+    console.log('Group - ' + group);
```
xkureck, line 131: I guess the logs are no more needed

libs/perun/components/src/lib/edit-group-dialog/edit-group-dialog.component.html
```diff
@@ -0,0 +1,34 @@
...
+      <mat-form-field class="w-100">
+        <label class="w-100">
+          <input [formControl]="nameControl" matInput placeholder="{{'DIALOGS.EDIT_GROUP.NAME' | translate}}">
+        </label>
+      </mat-form-field>
+      <mat-form-field class="w-100">
+        <label class="w-100">
+          <textarea [formControl]="descriptionControl" class="overflow-hidden" mat-autosize matInput placeholder="{{'DIALOGS.EDIT_GROUP.DESCRIPTION' | translate}}"></textarea>
+        </label>
+      </mat-form-field>
```
xkureck, line 15: There is no mat-error message when the name/description of the group is empty.
Also I think there would also be needed things from your other PR where there should be some message about group name regex.

libs/perun/components/src/lib/group-menu/group-menu.component.ts
```diff
@@ -25,4 +31,12 @@ export class GroupMenuComponent implements OnInit {
     console.log('Emit menu - ' + this.group);
```
xkureck, line 31: this log can be also deleted I think

libs/perun/components/src/lib/group-menu/group-menu.component.html
```diff
@@ -1,13 +1,57 @@
-<button (onMenuClose)="whenClosed.emit()" [disabled]="disabled" [mat-menu-trigger-for]="groupMenu" mat-icon-button>
-  <mat-icon>more_vert</mat-icon>
-</button>
-<mat-menu #groupMenu="matMenu" >
-  <button mat-menu-item (click)="onMoveGroup()">
+<div *ngIf="displayButtons" class="buttons-container">
+  <button [cdkCopyToClipboard]="group.name" [matTooltipPosition]="'above'" class="ml-auto" color="accent"
```
Vojtech-Sassmann, line 2: I don't think there is any reason why this button should be more visible than the others. I would just make it black as well.

apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-groups/vo-groups.component.html
```diff
@@ -11,18 +11,20 @@ <h1 class="page-subtitle">{{'VO_DETAIL.GROUPS.TITLE' | translate}}</h1>
     [placeholder]="'VO_DETAIL.GROUPS.FILTER'"
     (filter)="applyFilter($event)">
   </app-debounce-filter>
-  <mat-checkbox color="primary" class="ml-3" [(ngModel)]="showGroupList" (change)="removeAllGroups()" #checkbox>
-    {{'VO_DETAIL.GROUPS.LIST_VIEW' | translate}}
-  </mat-checkbox>
+  <mat-slide-toggle #toggle (change)="removeAllGroups()" [(ngModel)]="showGroupList" class="ml-3" color="primary"
```
Vojtech-Sassmann, line 14: When you want to add some space between two elements, prefer to set the right margin of the element on the left. If you would do it this way, then when it is displayed on small devices where the whole content cannot fit on a single line, all elements will be nicely aligned on the left side.
before
after (1)

apps/admin-gui/src/app/shared/shared.module.ts
xkureck: you have NotFoundPageComponent in declarations twice

PR #381, opened 2020-07-24
Entities information

### LINKED ISSUE #375: Entities information
To our entities(vo, group, resource, facility,..) show right under their name related information as id, description...(inspire in old gui). The style of how it should be showed to user should be something like it is shown "short name" in vo management.

bodnara:
Resolves #375 

* added additional information to entities titles

### DIFF
--- a/apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-detail-page.component.html
+++ b/apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-detail-page.component.html
@@ -9,6 +9,31 @@
     <span class="text-muted">
       &nbsp;#{{resource.id}}
     </span>
+    <br>
+    <span class="text-muted mr-5 entity-info">
+      {{'RESOURCE_DETAIL.FACILITY' | translate}}:
+      <a class="resource-link" [routerLink]="['/facilities', resource.facilityId]">
+        {{resource.facility.name}}
+      </a>
+      <span class="text-muted">
+        &nbsp;#{{resource.facilityId}}
+      </span>
+    </span>
+    <span class="text-muted mr-5 entity-info break-line">
+      {{'RESOURCE_DETAIL.VO' | translate}}:
+      <a class="resource-link" [routerLink]="['/organizations', resource.vo.id]">
+        {{resource.vo.name}}
+      </a>
+      <span class="text-muted">
+        &nbsp;#{{resource.voId}}
+      </span>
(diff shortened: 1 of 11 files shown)


apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-detail-page.component.html
```diff
@@ -9,6 +9,20 @@ <h1 class="page-title">
     <span class="text-muted">
       &nbsp;#{{resource.id}}
     </span>
+    <br>
+    <span class="text-muted mt-1 entity-info">
+      {{'RESOURCE_DETAIL.VO' | translate}}:
+      <span class="resource-text">
+        {{resource.voId}} / {{resource.vo.name}}
```
xkureck, line 16: I would like the voId by represented as &nbsp;#{{resource.voId}} and after the name of the vo

apps/admin-gui/src/app/vos/pages/member-detail-page/member-detail-page.component.html
```diff
@@ -9,6 +9,13 @@ <h1 class="page-title mb-1">
     <span class="text-muted">
       &nbsp;#{{member.id}}
     </span>
+    <br>
+    <span class="text-muted mt-1 entity-info">
+      {{'MEMBER_DETAIL.USER_ID' | translate}}:
+      <span class="member-text">
+        {{member.userId}}
```
xkureck, line 16: I would like the id of user be clickable and lead to admin section - user overview(lead into user detail)

apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-detail-page.component.html
```diff
@@ -9,6 +9,20 @@ <h1 class="page-title">
...
+      </span>
+    </span>
+    <br>
+    <span class="text-muted mt-1 entity-info">
+      {{'RESOURCE_DETAIL.DESCRIPTION' | translate}}:
+      <span class="resource-text">
+        {{resource.description}}
+      </span>
+    </span>
```
xkureck, line 25: I would like another information about the facility, should look like info about vo
both facility and vo should be clickable and lead into the facility/vo details
also try to somehow make to one row - or just  try to use as less height as possible, because I think that this "sideinfo" takes too much space, if thats not looking good just do one information one row.

PR #410, opened 2020-08-08
User profile - Breadcrumb update

xkostka2:
* New component for breadcrumb navigation was added as replacement for old library


apps/user-profile/src/assets/i18n/en.json
```diff
@@ -226,17 +228,19 @@
             "CANCEL_BUTTON": "Cancel"
           },
           "DELETE_DIALOG": {
-            "TITLE": "Delete attribute value",
+            "TITLE": "Delete attribue value",
```
xkureck, line 231: *attribute

PR #412, opened 2020-08-10
Member expiration improvements

### LINKED ISSUE #382: Set member expiration date and show and set expiration date for member's groups
Right now we are only showing membership status and expiration date(for vo) in the members overview. (and tbh not in very nice way)
Now we want to change member expiration date. Find out where would this feature would be positioned in our gui(so it make sense). 
Also now we want to show somewhere expiration date and status of member in his groups. With that we also want to change the member status in group. Find out where would this feature would be positioned in our gui. For that Vojta/Slavek advice would be nice I guess

xkostka2:
* Added dialog template for setting membership expiration
* Membership statuses VALID/INVALID were changed to ACTIVE/INACTIVE
* On hover over a status icon, the expiration date is displayed (for group memberships, if group expiration is not available, Vo expiration is shown)
* Added button to change member's expiration while in member detail
* Resolves #382

### DIFF
--- a/apps/admin-gui/src/app/shared/pipes/transform-member-status.pipe.ts
+++ b/apps/admin-gui/src/app/shared/pipes/transform-member-status.pipe.ts
@@ -1,0 +1,18 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'transformMemberStatus'
+})
+export class TransformMemberStatusPipe implements PipeTransform {
+
+  transform(memberStatus: string, memberGroupStatus?:string): string {
+    if(memberStatus.toLowerCase() === 'valid' && (!memberGroupStatus || memberStatus.toLowerCase() === 'valid')){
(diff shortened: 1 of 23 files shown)


apps/admin-gui/src/app/shared/pipes/transform-member-status.pipe.ts
```diff
@@ -0,0 +1,18 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'transformMemberStatus'
+})
+export class TransformMemberStatusPipe implements PipeTransform {
+
+  transform(memberStatus: string, memberGroupStatus?:string): string {
+    if(memberStatus.toLowerCase() === 'valid' && (!memberGroupStatus || memberStatus.toLowerCase() === 'valid')){
```
Vojtech-Sassmann, line 9: Just a tip: you can reuse here the function from utils that do the same thing.

PR #418, opened 2020-08-13
Authorization home section

### LINKED ISSUE #398: Authorization implemented - home section
What needs to be done:

xpavlic:
Resolves #398

- Added authorization check to dashboard if needed, also removed unnecessary check.

- Added authorization for attributes list


xpavlic:
I and @xkostka2 discussed policy where SELF: USER policy role is included and judged, that check for this policy is unnecessary, but I'm not 100% sure if we are right.

Vojtech-Sassmann:
Hi @xpavlic, I shouldn't probably be part of reviewers for these authorization changes unless you have some complex problems. Therefore I am removing myself from the reviewers.

xkureck (approved):

PR #421, opened 2020-08-20
Vo authorization

### LINKED ISSUE #393: Authorization implemented - vo section
What needs to be done:

bodnara:
Resolves #393 

* Added authorzation for:
    * Vo overview
    * Vo sidemenu
    * Vo members - add/remove/invite buttons, routing, based on this some columns were hidden (eg. checkbox)
    * Vo groups - create/delete buttons, routing, moving group, hiding columns in both tree and list view
    * Vo resources - 
                    * Overview and sidemenu
                    * Resource list - remove button, routing, hiding columns
                    * Resource tags - create/delete buttons, hiding columns
    * Vo settings - 
                  * Attributes - create/delete/save buttons, hiding columns
                  * Membership - save button
                  * Managers   - create/delete buttons, hiding columns
                  * Application form - settings/add/save/copy/clear buttons, hiding columns
                  * Notifications - add/remove/copy buttons, email footer and notification editing, hiding/disabling columns
                  * Ext. sources - add/remove buttons, hiding columns

### DIFF
--- a/apps/admin-gui/src/app/shared/side-menu/side-menu-item.service.ts
+++ b/apps/admin-gui/src/app/shared/side-menu/side-menu-item.service.ts
@@ -423,10 +423,15 @@ export class SideMenuItemService {
       label: 'MENU_ITEMS.VO.OVERVIEW',
       url: [`/organizations/${vo.id}`],
       activatedRegex: '/organizations/\\d+$'
+    }, {
+      label: 'MENU_ITEMS.VO.ATTRIBUTES',
+      url: [`/organizations/${vo.id}/attributes`],
+      activatedRegex: '/organizations/\\d+/attributes$'
     });
 
+
     // Members
-    if (this.authResolver.isThisVoAdminOrObserver(vo.id)) {
+    if (this.authResolver.isAuthorized('getCompleteRichMembers_Vo_List<String>_policy', [vo])) {
       links.push({
         label: 'MENU_ITEMS.VO.MEMBERS',
         url: [`/organizations/${vo.id}/members`],
@@ -435,8 +440,7 @@ export class SideMenuItemService {
     }
 
     // Groups
-    if (this.authResolver.isThisVoAdminOrObserver(vo.id)
-      || this.authResolver.isGroupAdminInThisVo(vo.id)) {
+    if (this.authResolver.isAuthorized('getAllRichGroupsWithAttributesByNames_Vo_List<String>_policy', [vo])) {
       links.push({
         label: 'MENU_ITEMS.VO.GROUPS',
         url: [`/organizations/${vo.id}/groups`],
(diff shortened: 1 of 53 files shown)


xkureck (changes requested):
image
After clicking on attributes the error occurs.
as Group admin of Cool group1:
image
cannot delete the group, the remove button is missing and the checkboxes as well
As resource admin on vo einfra:
image
The group section is showing which is fine, but I personally wouldnt let the resource admin go to the detail of the group, so disable routing please.
Top group creator looks fine.
As Vo observer:
image
The Vo observer cant see the resource states?
image
Missing some tooltip to tell the user why he cant save it.
image
not authorized exception
image
same with resend
In settings - managers the user in table have routing set to true, even when you are only vo-observer and should have the rights to look to the user detail.
image
No information why i cant save the footer.
image
same here
also 
image
There is another button "Verify" for applications that are waiting for mail approve, the button must be available only for perun admins

New information: the vo attributes can be accessed even if you are just a vo member

xkureck (changes requested):
image
missing authorization to edit button of the vo
as Vo observer:
image
no tooltip of why i cant select the attribute
image
the checkbox column in table is missing which is right, but the button for safe and remove are still here(disabled). Can we remove them completely? or I would rather have the checkboxes here and the tooltip of why the checkbox is disabled
image
No tooltip why i cannot delete the group (i am vo observer but at the same time the group manager of some groups)
Can you also please repair the bug when you are in the list view of the groups and you want to click on the menu, but instead you get routed to the group detail?

xkureck (changes requested):
image
The bug where you get from clicking of some manager in vo, that routes you into user detail is fixed in another branch?

apps/admin-gui/src/app/shared/side-menu/side-menu-item.service.ts
```diff
@@ -527,10 +527,15 @@ export class SideMenuItemService {
       label: 'MENU_ITEMS.VO.OVERVIEW',
       url: [`/organizations/${vo.id}`],
       activatedRegex: '/organizations/\\d+$'
+    }, {
+      label: 'MENU_ITEMS.VO.ATTRIBUTES',
+      url: [`/organizations/${vo.id}/settings/attributes`],
```
xkureck, line 532: wrong url

apps/admin-gui/src/app/vos/pages/vo-select-page/vo-select-page.component.html
```diff
@@ -3,10 +3,10 @@
   </perun-web-apps-back-button>
   <h1 class="page-title">{{'VO_MANAGEMENT.TITLE' | translate}}</h1>
   <perun-web-apps-refresh-button (refresh)="refreshTable()" *ngIf="!loading"></perun-web-apps-refresh-button>
-  <button mat-flat-button color="accent" class="mr-2" (click)="onCreateVo()" *ngIf="!loading && isVoAdmin">
+  <button mat-flat-button color="accent" class="mr-2" (click)="onCreateVo()" *ngIf="!loading && isAuthorized">
     {{'VO_MANAGEMENT.CREATE_VO' | translate}}
   </button>
-  <button mat-flat-button color="warn" (click)="onRemoveVo()" *ngIf="!loading && isVoAdmin" [disabled]="selection.selected.length === 0">
+  <button mat-flat-button color="warn" (click)="onRemoveVo()" *ngIf="!loading && isAuthorized" [disabled]="selection.selected.length === 0">
```
xkureck, line 9: please use for this other policy than createVo (somethink like deleteVo_Vo_policy would be better)

xkureck (changes requested):
image
Same problem as in members, the routing to attributes would trigger the setting item in sidemenu

libs/perun/components/src/lib/group-menu/group-menu.component.ts
```diff
@@ -22,6 +23,19 @@ export class GroupMenuComponent {
 
   @Output() changeNameDescription: EventEmitter<void> = new EventEmitter<void>();
 
+  @Output() whenClosed: EventEmitter<boolean> = new EventEmitter<boolean>();
```
xkureck, line 26: for what it is here?

apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-attributes/vo-attributes.component.ts
```diff
@@ -53,9 +53,13 @@ export class VoSettingsAttributesComponent implements OnInit {
   tableId = TABLE_ATTRIBUTES_SETTINGS;
   pageSize: number;
 
+  createAuth: boolean;
+  deleteAuth: boolean;
+  editAuth: boolean;
```
xkureck, line 58: I hope this piece of code will make sense when it will be rebased or merge together with commit member:authorization?

xkureck (changes requested):
image
The delete somehow bugged again

xkureck (approved):

PR #424, opened 2020-08-25
Authorization Facility section

### LINKED ISSUE #395: Authorization implemented - facility section
What needs to be done:

bodnara:
Resolves #395 

Added authorization for:

* Overview - items
* Sidemenu - items
* Resources - Add/Remove buttons, routing
* Allowed groups - routing (routing is enabled/disabled specifically for each item in table)
* Service status - buttons
* Service config - this should be working properly once attribute-list changes are approved
* Service destinations - buttons
* Hosts - buttons, routing, host detail should work properly with attribute-list update
* Security teams - buttons (buttons are disabled for now)
* Attributes - ..attribute-list update
* Owners - buttons
* Managers - buttons


xkostka2 (changes requested):
image

Please, disable outline on-click when routing is disabled.

xkureck (changes requested):
I am completely missing authorization on facility host detail page I think

xkureck (approved):

xkureck (approved):

PR #430, opened 2020-08-27
Move group bug fix

### LINKED ISSUE #413: move members group
When I want to move members group under another group, I get a HttpErrorResponse.

**To Reproduce**
1. Go to List of Groups in some VO
2. Click on move group icon for members group
3. Select any other group
4. Click on move
5. See error

**Expected behavior**
The error occurred because the members group cannot be moved. Therefore, it would be good to have the move icon disabled for this group.

### LINKED ISSUE #414: move group under non existing group
When I choose to move a group under another group and type a destination group name that does not exist, then the moving group is shifted as a top-level group.

**To Reproduce**
1. Go to the list of groups in some Vo
2. Click on the move icon for some sub-group
3. Choose move under a group option
4. type some random string
5. Click on move
6. See that the group was moved as a top-level group

**Expected behavior**
When I type a non-existing group name, It shouldn't be possible to click on the move button, or after clicking on it some error message should be displayed.

bodnara:
Resolves #413  and Resolves #414 

* disabled option to move 'members' group as it cannot be moved
* fix bug when moving group under non-existing one would move the group to top level

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/move-group-dialog/move-group-dialog.component.html
+++ b/apps/admin-gui/src/app/shared/components/dialogs/move-group-dialog/move-group-dialog.component.html
@@ -23,13 +23,16 @@
       <input
         matInput
         placeholder="{{'DIALOGS.MOVE_GROUP.GROUP_SELECT' | translate}}"
+        required
+        (change)="selectedGroup = null"
         [matAutocomplete]="groupSelectAutocomplete"
         [formControl]="otherGroupsCtrl">
       <!--suppress AngularInvalidExpressionResultType -->
       <mat-autocomplete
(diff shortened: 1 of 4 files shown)


xkureck (changes requested):
I just noticed another bug
image
The button is disabled for group members, which is not what we want, we want only to disable the move group button.
Also we want to disable button update name/description for group members.
Also If I dont select any specific group there could be some tooltip why I cant click on the move.

xkureck (changes requested):
image
ok, now in the three dots the move and update buttons are disabled but when showing it this way the change description button is not disabled

xkureck (changes requested):
image
Ok but now the three dots button is disabled for members(both tree and list view)

xkureck (approved):

PR #432, opened 2020-08-31
Facility allowed users

### LINKED ISSUE #429: facility section - allowed users
To facility section implement new page named allowed user that will display table of users that have access to this facility. Use method `getAssignedUsers`.

bodnara:
Resolves #429 

* added page that displays users with access to facility


apps/admin-gui/src/app/shared/components/users-list/users-list.component.html
```diff
@@ -74,7 +74,8 @@
         <tr *matHeaderRowDef="displayedColumns" mat-header-row></tr>
         <tr
           *matRowDef="let user; columns: displayedColumns;"
-          [perunWebAppsForceRouterLink]="inDialog ? null : ['/admin/users', user.id]"
+          [perunWebAppsForceRouterLink]="(inDialog || disableRouting) ? null : ['/admin/users', user.id]"
```
xkureck, line 77: maybe this could be reworked in the users-list so that we dont have two indicators for the indication if the routing in the list is enabled. Rework it so there is only disableRouting.

xkureck (changes requested):

xkureck (approved):

PR #433, opened 2020-08-31
Resource assgined members

### LINKED ISSUE #428: resource section - add assigned Members
to the resource section add new page for assigned members
use method from resourceManager `getAssignedMembers` and show in table assigned members and their status

bodnara:
Resolves #428 

* added page for assigned members in resource detail


apps/admin-gui/src/app/shared/side-menu/side-menu-item.service.ts
```diff
@@ -711,4 +677,55 @@ export class SideMenuItemService {
...
+      {
+        label: 'MENU_ITEMS.RESOURCE.ASSIGNED_SERVICES',
+        url: [`/facilities/${resource.facilityId}/resources/${resource.id}/services`],
+        activatedRegex: '/facilities/\\d+/resources/\\d+/services$'
+      }];
+
+    if(this.authResolver.isAuthorized('getAssignedMembers_Resource_policy', [resource])){
+      links.push({
+        label: 'MENU_ITEMS.RESOURCE.ASSIGNED_MEMBERS',
+        url: [`/facilities/${resource.facilityId}/resources/${resource.id}/members`],
```
xkureck, line 702: for url you must use {baseUrl} because the it wont work in the vo-resource side

libs/perun/pipes/src/lib/member-status-tooltip.pipe.ts
```diff
@@ -8,8 +8,13 @@ import { parseMemberStatus } from '@perun-web-apps/perun/utils';
...
 
   transform(member: RichMember, showGroupStatuses: boolean): string {
-    const memberExpiration = member.memberAttributes.find(att => att.friendlyName === 'membershipExpiration');
-    const groupExpiration = member.memberAttributes.find(att => att.friendlyName === 'groupMembershipExpiration');
+    let memberExpiration = null;
+    let groupExpiration = null;
+
+    if(member.memberAttributes !== null){
+      memberExpiration = member.memberAttributes.find(att => att.friendlyName === 'membershipExpiration');
+      groupExpiration = member.memberAttributes.find(att => att.friendlyName === 'groupMembershipExpiration');
```
xkureck, line 16: Actually in this pipe and in the list it is showed the status on the group, but we dont care about group expiration there, we care about resource status, if the user is not expired on the resource
if this can be implemented you should make some research

xkureck:
yesterday boys had some fun so @vyskocilpavel  @balcirakpeter thanks for the help with review
also please rebase

libs/perun/components/src/lib/members-list/members-list.component.ts
```diff
@@ -76,8 +79,11 @@ export class MembersListComponent implements OnChanges, AfterViewInit {
   setDataSource() {
     this.displayedColumns = this.displayedColumns.filter(x => !this.hideColumns.includes(x));
     if (!!this.dataSource) {
-      this.dataSource.sort = this.sort;
+      this.dataSource.filterPredicate =
+        (data: RichMember, filter: string) => parseFullName(data.user).toLowerCase().includes(filter.toLowerCase());
+      this.dataSource.filter = this.filter;
```
xkureck, line 84: the filter method is not working as it should, the filtering should work also when filtering according to id

xkureck (changes requested):

xkureck (approved):

PR #435, opened 2020-08-31
Group Authorization

### LINKED ISSUE #394: Authorization implemented - group section
What needs to be done:

xpavlic:
Resolves #394
Authorization for group section added.

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.ts
+++ b/apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.ts
@@ -1,7 +1,11 @@
 import { Component, Inject, OnInit } from '@angular/core';
 import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
-import {openClose, tagsOpenClose} from '@perun-web-apps/perun/animations';
-import { ApplicationMail, RegistrarManagerService, VosManagerService } from '@perun-web-apps/perun/openapi';
+import { openClose, tagsOpenClose } from '@perun-web-apps/perun/animations';
+import {
+  ApplicationMail,
+  GroupsManagerService,
+  RegistrarManagerService
+} from '@perun-web-apps/perun/openapi';
 import { GuiAuthResolver } from '@perun-web-apps/perun/services';
 
 export interface ApplicationFormAddEditMailDialogData {
@@ -27,7 +31,9 @@ export class AddEditNotificationDialogComponent implements OnInit {
   constructor(private dialogRef: MatDialogRef<AddEditNotificationDialogComponent>,
               private registrarService: RegistrarManagerService,
               @Inject(MAT_DIALOG_DATA) public data: ApplicationFormAddEditMailDialogData,
-              private authResolver: GuiAuthResolver) { }
+              private authResolver: GuiAuthResolver,
+              private groupsService: GroupsManagerService) {
+  }
 
   applicationMail: ApplicationMail;
   showTags = false;
@@ -42,10 +48,14 @@ export class AddEditNotificationDialogComponent implements OnInit {
     this.applicationMail = this.data.applicationMail;
     this.theme = this.data.theme;
 
-    if(this.data.voId){
+    if (this.data.groupId) {
+      this.groupsService.getGroupById(this.data.groupId).subscribe(group => {
+        this.editAuth = this.authResolver.isAuthorized('group-addMail_ApplicationForm_ApplicationMail_policy', [group]);
+      });
+    } else if (this.data.voId) {
       const vo = {
         id: this.data.voId,
-        beanName: "'Vo"
+        beanName: 'Vo'
       };
 
       this.editAuth = this.authResolver.isAuthorized('vo-addMail_ApplicationForm_ApplicationMail_policy', [vo]);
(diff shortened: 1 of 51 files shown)


xkureck (changes requested):
I guess attributes was made by Adam so when you rebase the attributes should work as it should?
image
Because right now I cant see any tooltip why I can select the attributes
image
As a vo observer I can see the checkboxes - that is really weird
image
vo observer: I should see button create subgroup
image
Shouldnt see that
image
wrong tooltip when I want to click on checkbox and also I can click on create even if I dont have nothing checked, I know that it is not part of authorization but please fix that
image
Also please fix bug when I want to get into the detail of application of the group the sidemenu get crazy and it get me into the vo application, which is wrong, I want to stay in the group applications

libs/perun/components/src/lib/groups-list/groups-list.component.html
```diff
@@ -1,8 +1,11 @@
-<div [class.hide-table]="exporting" [hidden]="groups.length === 0 || dataSource.filteredData.length === 0" class="card mt-2">
+<div [class.hide-table]="exporting" [hidden]="groups.length === 0 || dataSource.filteredData.length === 0"
+     class="card mt-2">
   <div class="card-body">
-    <perun-web-apps-table-options (exportEnd)="exporting = false" (exportStart)="exporting = true" [exporter]="exporter"></perun-web-apps-table-options>
+    <perun-web-apps-table-options (exportEnd)="exporting = false" (exportStart)="exporting = true"
+                                  [exporter]="exporter"></perun-web-apps-table-options>
```
xkureck, line 5: one "thing" one row please

libs/perun/components/src/lib/groups-list/groups-list.component.html
```diff
@@ -1,8 +1,11 @@
...
     <div class="overflow-auto">
-      <table #exporter="matTableExporter" [dataSource]="dataSource" class="w-100" mat-table matSort matSortActive="id" matSortDirection="asc" matSortDisableClear
+      <table #exporter="matTableExporter" [dataSource]="dataSource" class="w-100" mat-table matSort matSortActive="id"
+             matSortDirection="asc" matSortDisableClear
```
xkureck, line 8: one "thing" one row please

libs/perun/components/src/lib/groups-list/groups-list.component.html
```diff
@@ -15,15 +18,18 @@
...
-                          [aria-label]="checkboxLabel(row)"
-                          [checked]="selection.isSelected(row)"
-                          [disabled]="(row.name === 'members' && disableMembers) || disableSelect(row.id)"
-                          matTooltip="{{'SHARED_LIB.PERUN.COMPONENTS.GROUPS_LIST.DISABLED_HINT' | translate}}"
-                          [matTooltipDisabled]="!disableSelect(row.id)"
-                          color="primary">
-            </mat-checkbox>
+            <span
+              [matTooltip]="(authType === 'create-relation-dialog' ? 'SHARED_LIB.PERUN.COMPONENTS.GROUPS_LIST.CREATE_RELATION_AUTH_TOOLTIP' :
+               'SHARED_LIB.PERUN.COMPONENTS.GROUPS_LIST.ALREADY_MEMBER_TOOLTIP') | translate"
```
xkureck, line 23: better code readability - translate in in the ts file and here you get only variables

xkureck (changes requested):
image
Hide the outline when the routing is disabled

bodnara (changes requested):
Editing of email footer not working for any role
image

Same goes for editing of notifications
image

As Vo observer, I cannot open notification detail. I assume that this role should be able to see content of the notification only editing should be disabled.
image

After adding member the dialog is not closing. Perhaps you could fix it even if it is not part of authorization.
image

As group-admin after creating subgroup I need to refresh page to gain authorization rights over this group.
Before refresh:
image
After refresh:
image

As group admin I don't have rights to add/remove resources. I am not sure if this is expected behavior, maybe I am missing some information.
image

xpavlic:
image
this is expected behaviour, because you also need RESOURCESELFSERVICE role.
image

apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.ts
```diff
@@ -45,10 +45,16 @@ export class AddEditNotificationDialogComponent implements OnInit {
...
       };
 
       this.editAuth = this.authResolver.isAuthorized('vo-addMail_ApplicationForm_ApplicationMail_policy', [vo]);
+    } else if (this.data.groupId) {
+      const group = {
+        id: this.data.groupId,
+        beanName: "Group"
+      };
+      this.editAuth = this.authResolver.isAuthorized('group-addMail_ApplicationForm_ApplicationMail_policy', [group]);
     }
```
xkureck, line 58:
```suggestion
      if (this.data.groupId) {
      this.groupsService.getGroupById(this.data.groupId).subscribe(group => {
        this.editAuth = this.authResolver.isAuthorized('group-addMail_ApplicationForm_ApplicationMail_policy', [group]);
        console.log('edit auth');
        console.log(this.editAuth);
      });
    } else if (this.data.voId) {
      const vo = {
        id: this.data.voId,
        beanName: "Vo"
      };
      this.editAuth = this.authResolver.isAuthorized('vo-addMail_ApplicationForm_ApplicationMail_policy', [vo]);
    }
```
This is suggestion for whole block starting from row 45. The method isAuthorized in this case need whole object group. It needs voId in the group object

xkureck (changes requested):

xkureck (approved):

PR #437, opened 2020-08-31
Settings-AdvancedSettings, attributes moved to overview

### LINKED ISSUE #411: Little structual changes
Every entity(vo, member, ..) or almost every entity have now in their menu item "Settings". We want to rename that to "Advanced settings". Also we want to move item "attributes" (in every entity) from under the settings up, so the item attributes will be right under the "overview"

xpavlic:
Resolves #411
- Settings renamed to Advanced Settings

- Attribute pages of entities moved from Advanced Settings to overview pages


xkureck (changes requested):
You probably want to continue with working on this PR when the authorization will be completely done(merged into the master). Then you will rebase and see what needs to be done.

xkureck (approved):

PR #443, opened 2020-09-02
Delete loa

### LINKED ISSUE #436: Delete loa from whole gui
We have this attribute (extSourceLoa) showed in whole GUI in some of ours tables. This attribute is old an we want to remove it from our tables and everywhere where is showed(now I know for sure it is showed in applications list, and in user - identities.
1. Find the usage of loa
2. delete if everywhere it is used

bodnara:
Resolves #436 

* removed every mention of loa (except openAPI generated classes/methods)


xkureck (changes requested):
For now in expiration-settings.component.html let the loa be still present, othewise its good

xkureck (approved):

PR #446, opened 2020-09-07
Resource list - height of rows

esterbando:
Resolves#366

I found out that in resources-list.component.scss was one class 'mat-row' wrongly named and that caused that rows had smaller height than others lists. So I changed class name 'mat-row' to 'mat-row-height'.

### DIFF
--- a/libs/perun/components/src/lib/resources-list/resources-list.component.scss
+++ b/libs/perun/components/src/lib/resources-list/resources-list.component.scss
@@ -2,10 +2,6 @@ table {
   width: 100% !important;
 }
 
-.mat-row {
-  height: auto;
-}
-
 .mat-cell {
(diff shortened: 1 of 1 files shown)


libs/perun/components/src/lib/resources-list/resources-list.component.scss
```diff
@@ -2,7 +2,7 @@ table {
   width: 100% !important;
 }
 
-.mat-row {
+.mat-row-height {
   height: auto;
 }
```
xkureck, line 7:
```suggestion
```
You can completely delete this attribute in the code, it doesnt bring something new

xkureck (changes requested):

xkureck (approved):

PR #455, opened 2020-09-09
Reload principal

### LINKED ISSUE #452: Reload principal
When creating new entity(vo) the user gets automatically in perun vo admin rule for this vo. But that information isnt getting to us in the GUI. We store the data about user roles in perunPrincipal. We need to reload perunPrincipal to get the new information about user rules. With updated perunPrincipal we evaluate correctly if user is or is not admin of some entity.
The problem is definitely when creating vo, facility, group. Please explore if the bug is anywhere else.

bodnara:
Resolves #452 

* principal is reloaded after creating Vo, Group or Facility

### DIFF
--- a/apps/admin-gui/src/app/facilities/pages/facility-select-page/facility-select-page.component.ts
+++ b/apps/admin-gui/src/app/facilities/pages/facility-select-page/facility-select-page.component.ts
@@ -1,6 +1,6 @@
 import {Component, OnInit} from '@angular/core';
 import {SideMenuService} from '../../../core/services/common/side-menu.service';
-import { FacilitiesManagerService, RichFacility } from '@perun-web-apps/perun/openapi';
+import { AuthzResolverService, FacilitiesManagerService, RichFacility } from '@perun-web-apps/perun/openapi';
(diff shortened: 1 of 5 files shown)


apps/admin-gui/src/app/facilities/pages/facility-select-page/facility-select-page.component.ts
```diff
@@ -60,9 +64,14 @@ export class FacilitySelectPageComponent implements OnInit {
...
-      if (result) {
-        this.refreshTable();
+    dialogRef.afterClosed().subscribe(facilityCreated => {
+      if (facilityCreated) {
+        this.loading = true;
+        this.authzResolver.getPerunPrincipal().subscribe(perunPrincipal => {
+          this.storeService.setPerunPrincipal(perunPrincipal);
+          this.guiAuthResolver.init(perunPrincipal);
+          this.refreshTable();
+        }, () => this.loading = false);
```
xkureck, line 74:
```suggestion
        this.loading = true;
        this.initAuthService.loadPrincipal().then(() => this.refreshTable());
```
you can remove console log from loadPrincipal method, please use this method in vos and groups too

xkureck (changes requested):

xkureck (approved):

PR #462, opened 2020-09-15
Dashboard changes, MyProfileAdded

### LINKED ISSUE #440: Sidemenu changes
Changes are around dashboard=home=overview. Dashboard will be still the main page. The dashboard will be in the side menu on the Home. Home will not have any "under items"
Create new item in the side menu named "My profile" (the same level as home/access management/facility management. Position of this new item should be between facility management and perun admin. In My profile there should be pages:
- page "organization" - where will be shown only the organizations in which I am member
- page "groups" - where will be shown only the groups where I am member
- my settings - as it is in the Home section now, with no changes
Also in the Home(dashboard) implement icon "information". Position: on every card, on the card title right after the text title and above. This icon should have tooltip which says what it means eg.: "Your organizations"("Organizations where are you Vo admin")

xpavlic:
Resolves #440 

- Home now contains only Dashboard

- Every dashboard card has information icon with dialog

- pages of logged user moved from Home to new main side menu item My profile

- organizations and groups pages of My profile now show only organizations and groups in which user is member, added filter into groups page

- User attributes moved from User advanced settings overview to User overview ( Admin section and My profile section)

- Bug showing wrong attributes of User in Admin section fixed


xkureck (changes requested):
image
Button profile(right top) when clicked leads nowhere.
image
The page subtitle should be above reload button and filter. Same problem in groups page.
Also another problem, same screenshot  - in the sidemenu item organization is selected and is active, but at the same time item overview is active.
image
Facility attributes section doesnt have subtitle at all(every pageshould have subtitle). Also filter description is wrong. I would also add refresh button.
image
Gui configuration section also doesnt have page subtitle.
image
Please make the order in the side menu and on the big blue block the same.

apps/admin-gui/src/app/shared/components/two-entity-attribute-page/two-entity-attribute-page.component.html
```diff
@@ -1,8 +1,8 @@
-<perun-web-apps-immediate-filter (filter)="applyFilter($event)" [placeholder]="'MEMBER_DETAIL.SETTINGS.RESOURCE_ATTRIBUTES.FILTER' | translate"></perun-web-apps-immediate-filter>
-
+<h1 class="page-subtitle">{{getTitle() | translate}}</h1>
```
xkureck, line 1: This is not working at all
image
The title could be for sure not in this component but in the "user-settings-facility-component"

apps/admin-gui/src/app/users/components/user-profile/user-profile.component.html
```diff
@@ -0,0 +1,27 @@
+<div *ngIf="user !== undefined" class="container-fluid pl-xl-5 pr-xl-5 user-theme">
```
xkureck, line 1: I think that loading to this component is missing (you were doing the PR for adding loading to overview and headlines i believe)

xkureck (changes requested):
The gap between the page title and page subtitle isnt consistent. Also the page subtitle size(size of font) is not consistent(password reset)

apps/admin-gui/src/app/users/components/user-profile/user-profile.component.html
xpavlic: There are no api requests in this component so loading isn't needed.

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -605,14 +609,26 @@
...
+      "RESOURCE_PAGE": {
+        "TITLE": "Resource Attributes",
         "DELETE": "Remove",
         "FILTER": "Filter resources by name",
         "SAVE": "Save",
         "CREATE": "Add",
-        "NO_FILTERED_RESOURCES": "No resources that match the filter found",
-        "NO_RESOURCES": "You have no assigned resources"
+        "NO_FILTERED_ENTITY_MESSAGE": "No resources that match the filter found",
+        "NO_ENTITY_MESSAGE": "You have no assigned resources"
```
xkureck, line 619: image
*No assigned resources

apps/admin-gui/src/app/users/pages/user-detail-page/user-organizations/user-organizations.component.html
```diff
@@ -0,0 +1,39 @@
+<h1 *ngIf="this.showPrincipal"
```
xkureck, line 1: tbh this variable "showPrincipal" is not very describing for what it is needed. Rename it so it is more descriptive.

xkureck: same in users-groups

xkureck (changes requested):

xkureck (approved):

PR #474, opened 2020-09-23
Facility task results

### LINKED ISSUE #459: Facilities - service status - task result page
This page exists in old gui. Implement it. Also this page leads into another page. Make research why this two pages cant be merged into one. Othewise keep the functionality as it is in old gui. Make sure that the task that ended up in error state are showed on top of the page.

bodnara:
Resolves #459 

* added page with task results of services


xkureck (changes requested):
image
The old gui when you look into the column success message, every message starts on the new line - that looks really nice, and we would like it too.
image
Page subtitle: "task result: {{nameOfTheService/nameOfTheTask}}. How it is right now I dont understand where I am in the system
Filtering in table is not working according to name of the destination.

xkureck (changes requested):
image
Tasks with id 0 shouldnt be "routable" or clickable.
image
page subtitle - we always show ids in muted and also with # 
Filtering in table still not working

xkureck (changes requested):
image
The page subtitle please without the ":"

xkureck (approved):

PR #475, opened 2020-09-23
Group synch indication

### LINKED ISSUE #458: group - synchronization enabled
if group has attribute "synchronization enabled" set on true, we want to show to the user this fact on the group detail. I would place it somewhere in the subtitle of the page, or at least in the overview of the groups. Also when you are on the group entity and are in the member subsection the button add should be disabled for this groups and there should be tooltip that says somethink like: "adding to this group is disabled because the group is filled with members from another service" - or somethink like that. For more details ask someone from core, eg. Vojta.

bodnara:
Resolves #458 

* added info about synchronization to group detail
* disabled add button in group members if group is synched to some ext. source

### DIFF
--- a/apps/admin-gui/src/assets/i18n/en.json
+++ b/apps/admin-gui/src/assets/i18n/en.json
@@ -475,6 +475,7 @@
     "ENTITY": "Group",
     "VO_NAME": "Organization",
     "DESCRIPTION": "Description",
+    "SYNCHRONIZATION": "Synchronization",
     "SUBGROUPS": {
       "TITLE": "Subgroups",
       "CREATE_GROUP": "Create subgroup",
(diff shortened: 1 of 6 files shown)


apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -490,7 +492,8 @@
       "LIST_ALL": "List all members",
       "SEARCH_ALERT": "You can search member by name, login and email. Also, you can list all valid members.",
       "NO_MEMBERS_ALERT": "No members found.",
-      "INVITE_MEMBER": "Invite"
+      "INVITE_MEMBER": "Invite",
+      "SYNCHRONIZED": "Adding to this group manually is disabled, group is filled with members from another service."
```
xkureck, line 496: "...from external source"
I would probably have this tooltip for that

xkureck (changes requested):
Also we came to another use case - the invite and also removing should be disabled for this type of groups.
Also I thing instead of info "ENABLED" I would probably rather have the icon that is also at vo-groups with also the functionality that icon provides.

xkureck (approved):

xkureck (approved):

PR #479, opened 2020-09-29
Service destinations

### LINKED ISSUE #457: Service section - destinations
implement page destinations in the service subsection. Functionality should be the same as in old gui.

bodnara:
Resolves #457 

* added page with destinations to service detail


xkureck (changes requested):
steps to reproduce the problem:
1. filter something that will filter 1 item in the list
2. remove that item
3. then the table refresh, but there is nothing in the table even tho there is lot of items in the table in fact

image

xkostka2 (commented):
image
I think it would be nice to show the user, which field is used for the value.

image

Destinations can be added one by one and multiple destinations can be removed at once. Just change singular/plural in these 2 dialogs.

Otherwise, it seems good to me.

xkostka2 (changes requested):
image

I managed to be able to add service destination without selecting the service. @xkureck this should not happen, right? First, I selected the value and then I selected "No service available".

Apart from this, the solutions seem to be working well.

xkureck (changes requested):
image
still cannot sort in table according to the facility column. The sorting according to destination id is not working as well.

xkureck (approved):

PR #481, opened 2020-09-29
Copy form item privilege bug fix

### LINKED ISSUE #473: Copy form item(Application form)
Copy form items dialog is now also using "getFormItems" method. This dialog is checking the form items of other vos if the application form isnt empty. This was for the use case to check if the application form isnt empty. - and if it is dont allow the copy. The problem is that vo manager that should have to do this doesnt have the priviledge to call this method on other vos/groups - which is right. Find another way how to tell the user that application form that he wants to copy is empty or remove using the check at all.

bodnara:
Resolves #473 

* fixed bug when copy form items dialog would load Vos that user cannot access

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/application-form-copy-items-dialog/application-form-copy-items-dialog.component.html
+++ b/apps/admin-gui/src/app/shared/components/dialogs/application-form-copy-items-dialog/application-form-copy-items-dialog.component.html
@@ -1,47 +1,24 @@
 <div class="{{theme}}">
   <h1 mat-dialog-title>{{'DIALOGS.APPLICATION_FORM_COPY_ITEMS.TITLE' | translate}}</h1>
   <mat-spinner *ngIf="loading" class="ml-auto mr-auto"></mat-spinner>
-  <div *ngIf="!loading" mat-dialog-content class="dialog-container">
+  <div [hidden]="loading" mat-dialog-content class="dialog-container">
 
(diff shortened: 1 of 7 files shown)


xkureck (changes requested):
I will ask you for some more things.
1. on application form comment button clear form - it is not working, ale we dont have use case for this button, its only for testing purpouses
2. I would like the same select field as it is on Vo for choosing groups
3. Also you didnt fix the bug that I wanted from you to fix. 
image
in best scenario we would like to inform the user if the application form(that he wants to copy) is empty. But the method that we call to be informed about that needs priviledges that user normally doesnt have. Find out if that information cant be get any other way and if not remove this check.

xkostka2 (changes requested):
image

There are still displayed VOs where the user doesn't have permission to copy Application form.

image

When I selected group and then changed VO, it still made a request for group-to-vo instead of vo-to-vo

image

This is just a minor cosmetic issue. I don't see the point of copying the Application form of the same entity into itself (whether it is Vo or group)

bodnara:
> There are still displayed VOs where the user doesn't have permission to copy Application form.

VOs are displayed correctly because you need to select VO in order to select Group from which you can copy. The problem is that you cannot copy from the VO itself. After some consultation with @xkureck, we decided that upon some incorrect combination of VO and Group notification will pop to explain to user what happened.

xkostka2 (approved):

xkureck (approved):

PR #491, opened 2020-09-30
AddMemberGroupDialog changes

### LINKED ISSUE #485: AddMemberGroupDialogComponent
Add to this dialog functionality to disable checkbox in the table when:
1. user is already member of the group  - with tooltip that user is already member of group
2. the groups is synchronized - group is filled with members from external source, with tooltip - group is filled with members from external source

bodnara:
Resolves #485 

* selection in this dialog is disabled for groups that are synchronized and groups that user is already member of


xkureck (changes requested):
image
This is weird, happened when I clicked on select all

xkureck (approved):

PR #492, opened 2020-09-30
Create group as subgroup

### LINKED ISSUE #487: Create group - option create as subgroup
Old gui can make group as subgroup from scratch. New gui still cant make it. Make it happen!
Also on edit group dialog we have restriction that description of group can have only 128 lenght. Do this restriction also for create dialog.

bodnara:
Resolves #487 

* added option of creating subgroup from scratch to create group dialog


xkureck (changes requested):
As select for the group I would also like that select that Vojta did in the create resource dialog.
image
No group can be subgroup of members group.

xkureck (approved):

PR #499, opened 2020-10-02
Allowed groups filter, Group list org column

### LINKED ISSUE #494: Allowed groups - get filter, organization column - name
facility management - Allowed groups - get filter for groups(according to the name/id/description)
organization column in the group list - now there is shown id of organization. That is really not good information for the user. We want to show in that column name of the org.

bodnara:
Resolves #494 

* added filter to allowed groups page
* group list organization column now shows name of organization instead of id

### DIFF
--- a/libs/perun/components/src/lib/groups-list/groups-list.component.ts
+++ b/libs/perun/components/src/lib/groups-list/groups-list.component.ts
@@ -4,7 +4,7 @@ import {
   EventEmitter,
   HostListener,
   Input,
-  OnChanges, Output,
+  OnChanges, OnInit, Output,
   SimpleChanges,
   ViewChild
 } from '@angular/core';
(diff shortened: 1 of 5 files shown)


xkureck (changes requested):
Filtering according to organization column not working, also sorting(at the header) is not working. 
image
And some rows even dont have the column filled with the org.
image
Same here .

libs/perun/components/src/lib/groups-list/groups-list.component.ts
```diff
@@ -132,9 +136,25 @@ export class GroupsListComponent implements AfterViewInit, OnChanges {
   setDataSource() {
     this.displayedColumns = this.displayedColumns.filter(x => !this.hideColumns.includes(x));
     if (!!this.dataSource) {
+      this.dataSource.sortingDataAccessor = (item, property) => {
+        switch (property) {
+          case 'vo' :{
+            if (item.id) {
+              return this.voNames.get(item.voId);
```
xkureck, line 143: make the filtering also case insensitive

libs/perun/components/src/lib/groups-list/groups-list.component.ts
```diff
@@ -132,9 +136,25 @@ export class GroupsListComponent implements AfterViewInit, OnChanges {
...
+            }
+            break;
+          }
+          default: return item[property];
+        }
+      };
       this.dataSource.sort = this.sort;
-      this.dataSource.paginator = this.paginator;
+      this.dataSource.filterPredicate = (data: Group, filter: string) => {
```
xkureck, line 151: the filtering is also not working as it should

xkureck (changes requested):

xkureck (approved):

PR #500, opened 2020-10-05
App alert reworked

### LINKED ISSUE #461: App alert rework
Check the AlertComponent. 
- This component we are using for showing alert (eg. No groups) when there is no item in the table or when filtering filters no items. Create for this new component. That component will just show the plain text in the middle of the screen in bold typing that eg.: "No groups founded". 
- This component we are also using for INFO (eg. searching doesnt started yet, type something to search). In this case we would like also new component that will show the same information but without the blue colour.
- We are using this component also for ERROR - this is what we will use this component for. Dont change it.

xpavlic:
Resolves #461 

- created Alert component showing bold text in the middle of page

- created Info component showing info message as used to, but without blue background

- old alert component renamed to Error component 

- usage of old alert component updated


xkureck (changes requested):
image
The style of  when table is empty/no items are filtered should be like this alert, that means some icon - dont have to be like this one, frame around the text and the text is on left side

xkureck (changes requested):
image
forget to use the component here
ok, originally I thought that app-info is for this
image
and for error it is component app-alert
but it is vice versa, that is really weird, I would expect for some error app-alert component and for some info app-info component
also I made a mistake, when I thought it must be another component, it can be just 1 component it can be named app-alert but, in the specification you shouldnt specify color, but the component should take some input where we would specify it it is used for warn or info, and in the component itself there could be *ngIf that would specify which look to use


and also 
image
This should be app-alert not info

xpavlic:
image
this is warn alert not info alert (info allert doesn't have image)

xkureck (approved):

PR #521, opened 2020-10-22
Expiration of member in groups

### LINKED ISSUE #466: Expiration of member in groups
In member detail next to the expiration in vo create new window for selecting groups of the member. When group is selected the expiration date appears. Also the will be pencil to pop the dialog that will change the expiration date to some different date(something similar to change expiration in VO).

xkostka2:
* Actual OpenAPI methods were generated
* Added expiration column into groups-list
* Resolves #466

### DIFF
--- a/libs/perun/pipes/src/lib/parse-date.pipe.ts
+++ b/libs/perun/pipes/src/lib/parse-date.pipe.ts
@@ -1,0 +1,16 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { formatDate } from '@angular/common';
+
+@Pipe({
+  name: 'parseDate'
+})
+export class ParseDatePipe implements PipeTransform {
+
+  transform(value: string): string {
+    if(!value || value.toLowerCase() === 'never'){
+      return value;
+    }
+    return formatDate(value, 'd.M.yyyy', 'en');
+  }
+
+}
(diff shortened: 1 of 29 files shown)


xkureck (changes requested):
bug:
image
i clicked on table pick a date and the check which option i have choosen has dissapeard 
When I think about this, lets not invent wheel again. Lets inspire somewhere. Eg. how google is picking date.
image
(because now you made the input disabled and it make these problems)

libs/perun/pipes/src/lib/parse-date.pipe.ts
```diff
@@ -0,0 +1,16 @@
...
+@Pipe({
+  name: 'parseDate'
+})
+export class ParseDatePipe implements PipeTransform {
+
+  transform(value: string): string {
+    if(value.toLowerCase() === 'never'){
+      return value;
+    }
+    return formatDate(value, 'd.MM.yyyy', 'en');
```
xkureck, line 13:
```suggestion
    return formatDate(value, 'd.M.yyyy', 'en');
```
without this the format of the date was like this 
image

xkureck (changes requested):
Ok, I actually liked the look of how the dialog looked before this.
image
I think this was good, I just wanted so that the input field for expiration would function like button that would show the callendar.
This is how they handle date picking in outlook.
image
And this is google
image
Its very similar.
In member-overview.component.html please remove <i> around expiration date.

xkureck (approved):

PR #523, opened 2020-10-23
Copy notification new selectors

### LINKED ISSUE #517: Copy mail notifications from Vo/group
Add new select components to copy notification dialog.

bodnara:
Resolves #517 

* added new selectors to copy notification dialog


xkureck (changes requested):
I dont like it when some error appears(form not exist when you choose some group that doesnt have form) the search fields are reseted, is that something we have problem also in the application form?(ok, i checked that and we doesnt have problem with that there, so please fix this)

xkureck (approved):

PR #524, opened 2020-10-23
User roles

### LINKED ISSUE #470: My profile - my roles
There is something similar  but it is only shown in the admin section. 
image
This page should be also accessible from the section "My profile".
The page needs page-subtitle.
The main part of this issue is to make this page more user-friendly. That means there shouldnt be just IDs of the entities. There should be eg. colour of the entity and for sure the name of the entity, the relationship which that entity has, etc..
Just so now the casual user know what Perun roles he have.

xkostka2:
* Almost all roles now display tables with entities instead of only IDs
* Other(or newly added) roles be displayed in old format
* Roles tab was also added to 'My profile' section
* Resolves #470


xkureck (changes requested):
First of all the page title is missing.
image
Bug in the admin section - routing of the page of roles. You have it in both overview and advanced settings. Make it only on one place.

The biggest problem for me is UX. 
image
In the top of the card there is some text "self" and above that "Membership" and under that is some table. I have no idea what is in the table. I have no idea what self means. And the text is the same size, in that card it is almost invisible. Same with the table identities. The text "identities" I dont know what that means and what the hell is in the table, why it should be there?
Also maybe we would like to be listed the groups where user is just member.
image
Same problems that were in the previous picture. Here it is even more problem. I see id and the name of the group. I need description and name of the vo, for sure. With that information the table with the Vos is useles here.

Add roles - sponsor, sponsorship, group observer, facility observer, trusted facility admin, resource observer, resource self service, 
image
I think that you should some serious rethinking about UX and how tell it to user so it would be understandable. 
Also next to the role I think could be some info(icon with tooltip on in) what the role means, what you can do with that

bodnara (changes requested):
There is an error in Group admin panel, caused by the 'expiration' column of group table.
image

Another error in Sponsorship panel, this panel probably shouldn't be displayed when User isn't sponsored as is the case here.
image


I guess that in the meantime some new Roles were added perhaps you could rework those as well.
image


One small thing I noticed is that the PERUNADMIN role can be opened even though there is no additional info dispalyed.
And another small thing is that when you have small amout of roles. The role name is offcenter with the icon.
image
I wouldn't mind this myself but I already had to fix something similar. So it would most likely be pointed out by someone else.

apps/admin-gui/src/app/users/pages/user-detail-page/user-settings/user-roles/user-roles.component.html
```diff
@@ -1,8 +1,341 @@
...
+
+      <div class="mt-5">
+        {{'ROLES.GROUPADMIN_VOS' | translate}}
+        <perun-web-apps-vo-select-table
+          [displayedColumns]="['id', 'name', 'shortName']"
+          [vos]="vos">
+        </perun-web-apps-vo-select-table>
+      </div>
+    </div>
+  </mat-expansion-panel>
```
xkureck, line 53: image
The table of the organizations is really unnecessary because we show the names of the VOs in the table with groups, please remove it.

apps/admin-gui/src/app/users/pages/user-detail-page/user-settings/user-roles/user-roles.component.html
```diff
@@ -1,8 +1,341 @@
...
+
+      <div class="mt-5">
+        {{'ROLES.GROUPOBSERVER_VOS' | translate}}
+        <perun-web-apps-vo-select-table
+          [displayedColumns]="['id', 'name', 'shortName']"
+          [vos]="vos">
+        </perun-web-apps-vo-select-table>
+      </div>
+    </div>
+  </mat-expansion-panel>
```
xkureck, line 78: In group observer is the same as in groupAdmin

apps/admin-gui/src/app/users/pages/user-detail-page/user-settings/user-roles/user-roles.component.html
```diff
@@ -1,8 +1,341 @@
...
+        {{'ROLES.RESOURCEADMIN_FACILITIES' | translate}}
+        <app-facility-select-table
+          [displayedColumns]="['id', 'name']"
+          [checkAuth]="true"
+          [facilities]="facilities">
+        </app-facility-select-table>
+      </div>
+
+    </div>
+  </mat-expansion-panel>
```
xkureck, line 148: image
In the resource admin section also show only the table with resources(the table itself is showing the name of the VO and the name of the facility). The tables with facility and vo are redundant.

apps/admin-gui/src/app/users/pages/user-detail-page/user-settings/user-roles/user-roles.component.html
```diff
@@ -1,8 +1,341 @@
...
+        {{'ROLES.RESOURCEOBSERVER_FACILITIES' | translate}}
+        <app-facility-select-table
+          [displayedColumns]="['id', 'name']"
+          [checkAuth]="true"
+          [facilities]="facilities">
+        </app-facility-select-table>
+      </div>
+
+    </div>
+  </mat-expansion-panel>
```
xkureck, line 184: same problem as in resourceAdmin

apps/admin-gui/src/app/users/pages/user-detail-page/user-settings/user-roles/user-roles.component.html
```diff
@@ -1,8 +1,341 @@
...
+      <div class="mt-5">
+        {{'ROLES.RESOURCESELFSERVICE_FACILITIES' | translate}}
+        <app-facility-select-table
+          [displayedColumns]="['id', 'name']"
+          [checkAuth]="true"
+          [facilities]="facilities">
+        </app-facility-select-table>
+      </div>
+    </div>
+  </mat-expansion-panel>
```
xkureck, line 219: same problem as in resourceAdmin

apps/admin-gui/src/app/users/pages/user-detail-page/user-settings/user-roles/user-roles.component.html
```diff
@@ -1,8 +1,341 @@
...
+    </mat-expansion-panel-header>
+    <mat-spinner *ngIf="loading" class="ml-auto mr-auto"></mat-spinner>
+    <div *ngIf="!loading">
+      {{'ROLES.SPONSORSHIP_MEMBERS' | translate}}
+      <perun-web-apps-members-list
+        [members]="members"
+        [hideColumns]="['checkbox', 'email', 'logins', 'status']"
+      ></perun-web-apps-members-list>
+    </div>
+  </mat-expansion-panel>
```
xkureck, line 306: when I want to show myself a Sponsorship card an error occurs
image

xkureck (changes requested):
on role "self it is giving me some sort of error 
image
Group Admin is fine but when I am in Perun admin section it tells me that I am group admin of something not that martin kuba(same in vo manager, vo observer, ... everywhere - my suggestion is to show this helping information only when I am not in the perun admin subsection - perun admin dont need this info
image
The icon for trusted facility admin should be rather Vo icon I think
image
Same in top group creator - its role that is connected to VO and you are showing in the table Vos not groups
image

xkureck (changes requested):
image
some error - I think thats maybe specific to pavel zlamal

image
to facility table it would  be maybe nice if there was also column 'description'

image
image
For the role 'self' I think we need some description of what the hell is in those two tables. I was really confused  what means these two tables

xkureck (changes requested):
image
the problem with pavel zlamal and his role sponsor is not fixed yet

also rebase
I really want to merge this PR as soon as possible

xkureck (approved):

PR #526, opened 2020-10-29
Attribute list/map too many items

### LINKED ISSUE #519: Atributes list - too long row, because there are too many items in map/array
image

Atributes list row can be really way too long if the map or array is too long. 
When there is 5 items its ok, but if it is bigger than that the table start to look very cluttered.

I would like when there are more than 5 items in array/map in one row of the table to hide that other items of map/array and show button "show more" or "...". When user click on that button it will show him all the items in the array/map. The button "show less" would be nice too(when the items are expanded).

Something like that already exist in the microsoft teams. You can inspire there. Dont forget about good UX.

bodnara:
Resolves #519 

* attribute list/map now shows only set amount of items with option to show more/less if the amount is exceeded
* changed the system of adding items to attribute map to bring more consistency


xkureck (changes requested):
It is looking great! GJ
bug: 
image
When we are adding new attributes we definitely dont want this functionality
Also when editing those attributes it looks really weird and it is bugging.
The position of three dots would be better in the start of the column, not in the middle
Also maybe the icon should change to something more appropriate to express "show less"

xkureck (changes requested):
My suggestion:
1. for "show more" the three dots were actually good, I would suggest the icon for "show more" the three dots, I think that people are more used to it. The show less button is good, keep that.
2. adding item to list - Your solution is working but adding item to the top of list is not intuitive I think. Better solution I believe is that when user make the decision that he wants to add something, just do action as if the user clicked on "show more". Lets user show whole list, so he knows to what he is actually adding. And make it also that user is adding to the bottom of the list as it was before that.

xkureck (approved):

PR #528, opened 2020-10-29
Overview button leads to more buttons sign

### LINKED ISSUE #495: beak on the cubes
We would like to have some sort of sign on some of our cubes(in the overview). The sign will be on the cubes that "leads into another cubes". The goal is to improve user experience so that user know before he clicks on the cube that the cube leads into other cubes. On the side menu we implemented it with the sign of the beak of the item in the menu. Find out what sign would be the best in your opinion and place that sign on the cube.

bodnara:
Resolves #495 

* added drop down menu for overview buttons that redirect to another overview
* this works like a sign for the user and can also be used for quicker access
* added authorisation to drop-down menu items

### DIFF
--- a/apps/admin-gui/src/app/admin/pages/admin-page/admin-overview/admin-overview.component.ts
+++ b/apps/admin-gui/src/app/admin/pages/admin-page/admin-overview/admin-overview.component.ts
@@ -35,7 +35,18 @@ export class AdminOverviewComponent implements OnInit {
       cssIcon: 'perun-preview',
       url: '/admin/visualizer',
       label: 'MENU_ITEMS.ADMIN.VISUALIZER',
-      style: 'admin-btn'
+      style: 'admin-btn',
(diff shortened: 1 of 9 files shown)


xkureck (commented):
1. This is not what was the issue about
2. I absolutely love the idea, it looks great. I just must do some discussion with UX master if this is good idea, and how to use it correctly

xkureck (changes requested):
I dunno, I think the icon should be right bottom. But with that there appears problem with mobile devices. But if you would fix that it would be great.
image
Also rebase please.

xkureck (changes requested):
It is looking good.
Please rebase. Dont forget that there are some changes that you have made in PR for attributes that are already merged(in members the advanced settings are completely deleted, facility - security teams moved to advanced settings, sponsored members from advanced settings.. etc)

xkureck (approved):

PR #538, opened 2020-10-30
Two entities attributes

### LINKED ISSUE #532: attributes between entities
image
exist user-facility
member-resource
member-group
group-resource atributes
we implemented some of them, but not all, and we want to change the behaviour a bit, take look on the screen
we want to have the tab, and when user click on the tab the select of entity appears and then the standart table with attributes appear
image
we want to have look from the both sides of the attributes - eg.: member-resource attributes there will be tab on the member entity and resource entity as well
another problem with select: sometimes there are too many members on the resource/group(90K) that means there must be some good mechanics that this problem solves - there cant be standart select

bodnara:
Resolves #532 

* added tabs with 'connected' attributes to attribute pages of entities that have attribute connection with another entity
* reworked two entity attribute page
* added new selectors for facility, member, user and resource


apps/admin-gui/src/app/facilities/pages/facility-detail-page/facility-attributes/facility-attributes.component.html
```diff
@@ -1,26 +1,34 @@
 <div>
   <h1 class="page-subtitle">{{'FACILITY_DETAIL.SETTINGS.ATTRIBUTES.TITLE' | translate}}</h1>
-  <perun-web-apps-refresh-button (refresh)="refreshTable()"></perun-web-apps-refresh-button>
-  <button mat-flat-button class="mr-2" color="accent" (click)="onCreate()">
-    {{'FACILITY_DETAIL.SETTINGS.ATTRIBUTES.CREATE' | translate}}
-  </button>
+  <mat-tab-group>
+    <mat-tab label="Facility attributes">
```
xkureck, line 4: the label should be translatable

xkureck (changes requested):
image
please remove this sections(resource/group attributes, ...) from whole menu, we would like it on only 1 place
image
image
Also I dont think it is neccessary to have the select component that big(in terms of width) i think that w-50 would be ok
There is big problem if there is a lot of members on that resource/group. Take the try on the muni-test instance. Try open some random group in masaryk university organization and go into the attributes. Even group attributes didnt load which is problem. It is ok, if the select opening is really slow, but we dont want to slow getting the attributes that are connected only to resource/group/entity - for this try Tab group where the tab content is loaded lazily (when activated) https://material.angular.io/components/tabs/examples
image
refresh button is missing
also try to use icons for tab headers
image

xkureck (changes requested):
Ok, I was wrong with the icons, it doesnt look good, please remove them.
image
You forget to remove this tab.
The text on the tabs headers are good.

apps/admin-gui/src/app/users/pages/user-detail-page/user-attributes/user-attributes.component.ts
```diff
@@ -48,6 +55,7 @@ export class UserAttributesComponent implements OnInit {
   selection = new SelectionModel<Attribute>(true, []);
   attributes: Attribute[] = [];
   userId: number;
+  facilities: Facility[] = [];
```
xkureck, line 58: dunno why you added this, it doesnt seems like you are using it somewhere
if thats the case please delete also unused things from constructor and import

apps/admin-gui/src/app/vos/pages/group-detail-page/group-attributes/group-attributes.component.ts
```diff
@@ -47,6 +55,8 @@ export class GroupAttributesComponent implements OnInit {
   selection = new SelectionModel<Attribute>(true, []);
   attributes: Attribute[] = [];
   groupId: number;
+  members: RichMember[] = [];
+  resources: Resource[] = [];
```
xkureck, line 59: dunno why you added this, it doesnt seems like you are using it somewhere
if thats the case please delete also unused things from constructor and import

apps/admin-gui/src/app/shared/components/two-entity-attribute-page/two-entity-attribute-page.component.ts
```diff
@@ -43,32 +47,98 @@ export class TwoEntityAttributePageComponent implements OnInit, OnChanges {
...
-    } else if (this.firstEntity === 'group') {
-      this.setMessages('RESOURCE');
-    } else (this.setMessages(this.firstEntity.toUpperCase()));
+    this.loadEntityValues();
+    this.setMessages(this.secondEntity.toLowerCase());
   }
 
-  ngOnChanges(): void {
-    this.filteredEntityValues = this.entityValues;
+  loadEntityValues() {
```
xkureck, line 65: in whole this method you are calling method like getAllowedResources but you are not checking if the user have priviledge to call this method, how it is right now this can very easily happen
image
If the user not to authorized to call the method which will show him the select where he can select member/resource/.. then I wouldnt show him the tab actually

apps/admin-gui/src/app/shared/components/two-entity-attribute-page/two-entity-attribute-page.component.ts
```diff
@@ -43,32 +47,98 @@ export class TwoEntityAttributePageComponent implements OnInit, OnChanges {
...
+              this.entityValues = groups;
+              this.preselectEntity();
+              this.loading = false;
+            })
+        }
+        break;
+      case 'group':
+        switch (this.secondEntity){
+          case 'resource':
+            this.resourcesManagerService.getAssignedRichResourcesWithGroup(this.firstEntityId).subscribe(resources => {
```
xkureck, line 88: you can use method "getAssignedResources" you dont need the rich object

apps/admin-gui/src/app/shared/components/two-entity-attribute-page/two-entity-attribute-page.component.ts
```diff
@@ -43,32 +47,98 @@ export class TwoEntityAttributePageComponent implements OnInit, OnChanges {
...
+              this.entityValues = resources;
+              this.preselectEntity();
+              this.loading = false;
+            });
+            break;
+          case 'member':
+            this.membersManagerService.getCompleteRichMembersForGroup(this.firstEntityId, []).subscribe(members => {
```
xkureck, line 95: same here 
```suggestion
            this.groupManagerService.getGroupRichMembers(this.firstEntityId).subscribe(members => {
```

xkureck (changes requested):
image
I dont like this info I woul like the proper <app-alert alertType="warn"
image
When the page gets tighter then there is no margin between the tabs and the buttons

xkureck (changes requested):

xkureck (approved):

PR #551, opened 2020-11-05
Filtering in vo groups fixed

### LINKED ISSUE #547: filtering in vo - groups not working
filtering in page vo - groups not working

bodnara:
Resolves #547 

* fixed filtering in vo groups by changing the filter element


xkureck (changes requested):
I think we would like to use debounce filter. Your "fix" doesnt fix the problem.
Problem is that some of our groups for some reason doesnt have description(it is null)
image
image
group-tree.component.ts
image
The problem is on our muni instances.
Please solve this by that our filtering function in group-filter components is ready for groups with null names and null descriptions.

xkureck (approved):

PR #552, opened 2020-11-06
Group tree view description

### LINKED ISSUE #540: group tree view - description
the description of group should be on 1 row like it is now, but some description are longer so we want to do ... like now
what we want to change is only and only when the row is too big for the row, we want to pop up the whole description in the tooltip
also when user have role group admin on some groups + resource admin in the vo the margin of the description is terrible, try to make something with that

bodnara:
Resolves #540 

* description text is now shown in tooltip only if description is too  long
* made group menu buttons always visible but disabled according to user rights to fix description offset caused by non constant 
   number of  buttons shown

### DIFF
--- a/apps/admin-gui/src/styles.scss
+++ b/apps/admin-gui/src/styles.scss
@@ -1502,4 +1502,5 @@ th, td.mat-cell {
 //font size of tooltip
 .mat-tooltip {
   font-size: 14px !important;
+  word-wrap: break-word !important;
 }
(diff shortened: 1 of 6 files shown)


xkureck (changes requested):
GJ!
One problem with the invisible synchronization button(tested as group admin + resource admin in the vo)
image
Maybe better solution would be if the user didnt see the icons that he cannot use(as it is now) and the width for the menu would be set to fixed width, so there is not that problem with description. Yeah, that would be better and you would have less work with that. 
Because its weird as vo observer 
image
also 
image

xkureck (approved):

PR #560, opened 2020-11-17
User session expiration

### LINKED ISSUE #404: Handle user session expiration
* The user should be informed if his session has expired. Ideally, the application should show a dialog where would be a button that would start the authentication. 
* The expiration can be detected by subscribing to a particular event of the UsersManager in the AuthService. 
* To force the login, it should be enough to refresh the page. The application will detect that the user is not logged in and will redirect him to the login page.

bodnara:
Resolves #404 

* when session expires, dialog will pop-up to notify user about this event and asks him to sign in


xkureck:
Circular dependency between "perun-dialogs" and "perun-services" detected: perun-dialogs -> perun-services -> perun-dialogs

libs/perun/services/src/lib/auth.service.ts
```diff
@@ -4,6 +4,9 @@ import {from, Observable} from 'rxjs';
 import {map} from 'rxjs/operators';
 import { Params, Router } from '@angular/router';
 import { StoreService } from './store.service';
+import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
+import { SessionExpirationDialogComponent } from '../../../dialogs/src/lib/session-expiration-dialog/session-expiration-dialog.component';
```
Vojtech-Sassmann, line 8: All imports should start with `@perun-web-apps`

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann:
Hi @bodnara, could you please show this dialog also in ApiInterceptor? You can replace the UserNotLogedIn error.

Vojtech-Sassmann (approved):

xkureck (approved):

PR #562, opened 2020-11-23
Detection of new version of GUI

### LINKED ISSUE #544: detection of new version
when we are deploying new version of gui, the changes didnt get to the users that already opened the gui in some older version and the application will load for them in the old version, because the browsers cached the old version of the gui
Make a research how to resolve this problem and implement it

xkostka2:
* THE FUNCTIONALITY IS YET TO BE TESTED
* Added periodic check (every 30 seconds), whether the version of GUI changed and if so, the app is reloaded
* Resolves #544

### DIFF
--- a/apps/admin-gui/src/app/app.component.ts
+++ b/apps/admin-gui/src/app/app.component.ts
@@ -2,6 +2,12 @@ import { Component, HostListener, OnInit } from '@angular/core';
 import { CacheHelperService } from './core/services/common/cache-helper.service';
 import { StoreService } from '@perun-web-apps/perun/services';
 import { PerunPrincipal } from '@perun-web-apps/perun/openapi';
+import { interval } from 'rxjs';
+import { HttpClient, HttpHeaders } from '@angular/common/http';
+import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
+import { NewVersionDialogComponent } from './shared/components/dialogs/new-version-dialog/new-version-dialog.component';
+import { MatDialog } from '@angular/material/dialog';
+import { version } from '../../../../package.json';
 
 @Component({
   selector: 'app-root',
@@ -12,7 +18,9 @@ export class AppComponent implements OnInit {
 
   constructor(
     private cache: CacheHelperService,
-    private store: StoreService
+    private store: StoreService,
+    private http: HttpClient,
+    private dialog:MatDialog
   ) {
 
     this.cache.init();
@@ -33,6 +41,8 @@ export class AppComponent implements OnInit {
   displayWarning: boolean = this.store.get('display_warning');
   warningMessage: string = this.store.get('warning_message');
 
+  version = version;
+
   @HostListener('window:resize', ['$event'])
   getScreenSize(event?) {
     this.sidebarMode = this.isMobile() ? 'over' : 'side';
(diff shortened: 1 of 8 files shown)


apps/admin-gui/src/app/app.component.ts
```diff
@@ -46,6 +51,29 @@ export class AppComponent implements OnInit {
 
   ngOnInit(): void {
     this.principal = this.store.getPerunPrincipal();
+    interval(30000).subscribe(() => {
```
xkureck, line 54: so I we would assume that your solution is working, I dont think that I would like to check every 30 sec for new version and  if yes, just reload app. What if the user have some form half filled and then suddenly the gui reloads. That would be super anoying. I think we would like it just on the start of the application. 
Another viable solution, I think, could be checking the version when user enters component "user-dashboard" in the onInit method in that component
what do you think @Vojtech-Sassmann

xkostka2: If we checked for a new version just on the start of the app, users who already had the app opened, would not be using the latest version, in theory, for the rest of the day.

Vojtech-Sassmann: Checking for the new version every 30s might be a viable solution; the only problem I can see is the reload. We should inform users that we have released a new version and that they can reload the application. This should also be done across the whole application, not only the home screen since some users might be idling on a different page.

create-version-file.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+PACKAGE_VERSION=$(node -p "require('./package.json').version")
+echo "{\"version\":\"$PACKAGE_VERSION\"}" > ./dist/apps/admin-gui/assets/config/version.json
+echo "{\"version\":\"$PACKAGE_VERSION\"}" > ./apps/admin-gui/src/assets/config/version.json
```
Vojtech-Sassmann, line 4: On the app release, you don't want to write a change to the versioned code. This file shouldn't be edited here; it should be done manually or somewhere else.

apps/admin-gui/src/assets/config/version.json
```diff
@@ -0,0 +1 @@
+{"version":"0.10.3"}
```
Vojtech-Sassmann, line 1: As I mentioned in a different comment, this file should be either handled manually or maybe a better solution would be to put in the version some information, that this is a test version. 
```suggestion
{"version":"SNAPSHOT"}
```
Maybe something like this should work. That way, we don't need to manually bump the version each time a new version was released.

Vojtech-Sassmann (changes requested):
Also, please rework the reload of the app. Some closeable dialog might be the simplest solution for this problem.

apps/admin-gui/src/app/app.component.ts
xkureck: yep, agree with vojta, there should be dialog if the loadAppVersion detected changes in version. if there is detected change in version dialog should appear - in the title should be something like: "New version was detected. You must reload browser to load new version. ... " and two buttons: 1. "Reload GUI" 2. "Got it! I will reload GUI later"
something like windows has when they want to update
Also on the start of the application if the change in the version is detected then no dialog should appear and the gui should automatically reload

apps/admin-gui/src/app/app.component.ts
```diff
@@ -33,6 +40,8 @@ export class AppComponent implements OnInit {
   displayWarning: boolean = this.store.get('display_warning');
   warningMessage: string = this.store.get('warning_message');
 
+  version: string;
```
Vojtech-Sassmann, line 43: The first version you use for a check should be the one, which was used during the build of the app. 
```suggestion
  version = version;
```
You can import this version like this:
```
import { version } from '../../../../package.json';
```

apps/admin-gui/src/assets/config/version.json
Vojtech-Sassmann: Please, change this version to the suggested value.

apps/admin-gui/src/app/app.component.ts
```diff
@@ -46,6 +55,31 @@ export class AppComponent implements OnInit {
...
+    const httpHeaders = new HttpHeaders({
+      'Cache-Control': 'no-cache, no-store, must-revalidate, post-check=0, pre-check=0',
+      'Pragma': 'no-cache',
+      'Expires': '0'
+    });
+    this.http.get('/assets/config/version.json', { headers: httpHeaders })
+      .subscribe(result => {
+        const version = result['version'];
+        if (version) {
+          if (this.version && this.version !== version) {
```
Vojtech-Sassmann, line 74:
```suggestion
          if (this.version && version !== 'SNAPSHOT' && this.version !== version) {
```
Also, please add this change that would stop the alert from being shown in the development mode.

Vojtech-Sassmann (changes requested):
Once these changes are merged, we should probably test them at the devel instance.

xkureck (approved):

PR #564, opened 2020-11-24
Home page not loding

bodnara:
* hopefully fixed problem with sponsor role not having voId


apps/admin-gui/src/app/users/pages/user-detail-page/user-dashboard/user-dashboard.component.ts
```diff
@@ -120,6 +120,9 @@ export class UserDashboardComponent implements OnInit {
     return new Promise((resolve, reject) => {
       if (this.guiAuthResolver.isOnlySponsor()) {
         const vosSponsoring = this.storeService.getPerunPrincipal().roles['SPONSOR']['Vo'];
```
Vojtech-Sassmann, line 122: This line is where the problem occurs, to be more precise, in the last part `roles['SPONSOR']['Vo']` the Vo field might NOT be present. You need to detect if it is defined, not only the length.

Vojtech-Sassmann (changes requested):

xkureck (approved):
tested and its working

PR #566, opened 2020-11-24
Password reset of existing sponsored user

bodnara:
* added button with dialog for password reset to sponsored-members-list

### DIFF
--- a/apps/admin-gui/src/app/shared/components/sponsored-members-list/sponsored-members-list.component.html
+++ b/apps/admin-gui/src/app/shared/components/sponsored-members-list/sponsored-members-list.component.html
@@ -60,6 +60,15 @@
               mat-icon-button>
               <mat-icon>more_horiz</mat-icon>
             </button>
+            <button
+              (mouseenter)="disableRouting = true"
+              (mouseleave)="disableRouting = routingStrategy"
+              class="dark-hover-list-item"
+              aria-label="sponsors of member"
+              (click)="resetPassword(sponsoredMember)"
+              mat-icon-button>
+              <mat-icon>more_horiz</mat-icon>
+            </button>
           </td>
         </ng-container>
 
--- a/apps/admin-gui/src/app/shared/components/sponsored-members-list/sponsored-members-list.component.html
+++ b/apps/admin-gui/src/app/shared/components/sponsored-members-list/sponsored-members-list.component.html
@@ -60,15 +60,22 @@
               mat-icon-button>
               <mat-icon>more_horiz</mat-icon>
             </button>
-            <button
-              (mouseenter)="disableRouting = true"
-              (mouseleave)="disableRouting = routingStrategy"
-              class="dark-hover-list-item"
-              aria-label="sponsors of member"
-              (click)="resetPassword(sponsoredMember)"
-              mat-icon-button>
-              <mat-icon>more_horiz</mat-icon>
+            <span
+              *ngIf="passwdResetAuth(sponsoredMember)"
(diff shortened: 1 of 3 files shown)


apps/admin-gui/src/app/shared/components/sponsored-members-list/sponsored-members-list.component.html
```diff
@@ -60,6 +60,21 @@
...
+              [matTooltipPosition]="'above'"
+            >
+              <button
+                (mouseenter)="disableRouting = true"
+                (mouseleave)="disableRouting = routingStrategy"
+                class="dark-hover-list-item"
+                aria-label="sponsors of member"
+                (click)="resetPassword(sponsoredMember)"
+                mat-icon-button>
+                <mat-icon>restore</mat-icon>
```
xkureck, line 75:
```suggestion
                <mat-icon>vpn_key</mat-icon>
```
for this we probably dont have the perfect icon but I think this would be a little bit better

xkureck (changes requested):
It looks nice.
Just add check for the filtered logins - in some future cases it could be [] - and in that case the whole dialog gets broken. Look how martin checked it in the member-overview.component where this check is.

xkureck:
> It looks nice.
> Just add check for the filtered logins - in some future cases it could be [] - and in that case the whole dialog gets broken. Look how martin checked it in the member-overview.component where this check is.

in this solution you call 477 times method attributesManager/getLogins, thats not a good solution
If in the future this problem would occur we will take a loot on it but now just delete the call of this method

apps/admin-gui/src/app/shared/components/sponsored-members-list/sponsored-members-list.component.html
```diff
@@ -60,6 +60,21 @@
               mat-icon-button>
               <mat-icon>more_horiz</mat-icon>
             </button>
+            <span
+              *ngIf="passwdResetAuth(sponsoredMember)"
+              [matTooltip]="'SHARED.COMPONENTS.SPONSORED_MEMBERS_LIST.PASSWORD_RESET' | translate"
+              [matTooltipPosition]="'above'"
+            >
+              <button
```
Vojtech-Sassmann, line 68: Since you are not showing the dialog immediately and you are getting the logins, you should disable this button until the request is finished. This way, the user could click it multiple times because he would not be shown anything asap he clicks the button. Once the request has finished (or when the dialog is closed), you should enable this button.

Vojtech-Sassmann (changes requested):
Also, even it was not part of this PR, I would suggest changing the label of the email from `Email attribute` to `Email, where the request will be send.` This value is located at `DIALOGS.PASSWORD_RESET_REQUEST.EMAIL`.

Also, I am not sure if this functionality can be used at production, since it might require some additional changes to be made. (maybe including a whole new js application) This is more a note to @xkureck just as a reminder.

apps/admin-gui/src/app/shared/components/sponsored-members-list/sponsored-members-list.component.ts
```diff
@@ -145,4 +149,37 @@ export class SponsoredMembersListComponent implements OnChanges, AfterViewInit {
...
+        memberId: sponsoredMember.member.id,
+        logins: filteredLogins
+      };
+
+      const dialogRef = this.dialog.open(PasswordResetRequestDialogComponent, config);
+
+      dialogRef.afterClosed().subscribe(() => {
+        this.loading = false;
+      })
+    });
```
Vojtech-Sassmann, line 174: You should also set the loading to false if the method `getLogins` fails.

Vojtech-Sassmann (changes requested):

xkureck (approved):

PR #568, opened 2020-11-25
Add new roles to Managers

HejdaJakub:
* Roles are now fetching through api method getAllRolesManagementRules and they are also assigned to Vo/Group/Facility/Resource via method attribute named primaryObject.


bodnara (approved):
Everything seems to be working correctly. However I would like to ask @xkureck for review just to make sure I didn't miss anything.

xkureck (changes requested):
image
I got one problem with the order of the items in the select role selector. In the facility/group/resource managers page its actually ok, and I like that order, but in the vo-managers page the order of the items its not the way I would like. I would like as the first default item vo manager for sure, definitely not Sponsor. The order of the items here should be something like - vo manager, vo observer, sponsor, top group creator, trusted facility admin. Its because of UX, when user initially come here to this page he would automatically see Sponsors, but this role is on some our instances not used at all.

HejdaJakub:
> image
> I got one problem with the order of the items in the select role selector. In the facility/group/resource managers page its actually ok, and I like that order, but in the vo-managers page the order of the items its not the way I would like. I would like as the first default item vo manager for sure, definitely not Sponsor. The order of the items here should be something like - vo manager, vo observer, sponsor, top group creator, trusted facility admin. Its because of UX, when user initially come here to this page he would automatically see Sponsors, but this role is on some our instances not used at all.

The order has been edited. If some new role will be added in the future, the Vo Admin will be always at first position and the Vo Observer will be always at second position, but the rest of roles will be ordered by alphabet. Is it ok?

xkureck (approved):

PR #570, opened 2020-11-30
create sponsored members from CSV

xkureck:
* added functionality where gui uses new method for creating sponsored members in bulk
* functionality can be properly tested when will be new version of perun released on idm-test instance, now we can test only on cesnet-dev


Vojtech-Sassmann (changes requested):

Vojtech-Sassmann:
@xkureck I have tested this new functionality on idm-test and it seems to be working just fine.

Vojtech-Sassmann (approved):

PR #573, opened 2020-12-03
Redirect to a VO detail after creating the VO

### LINKED ISSUE #350: Redirect to a VO detail after creating the VO
I'm always frustrated when I create a VO and then I have to find that VO on the list again.

It would be nice if there is a redirection to the VO detail after the VO creation.

Another solution could be to add the newly created VO to the beginning of the list (as recently created or something like that).

**Actual situation:**
image
image

**Desired solution:**
image
image

**Alternative solution:**
image
image

HejdaJakub:
* If user created new VO before this implementation, he or she had to find the created VO manually, if he or she wanted to do something with this VO.
* Now user is automatically redirected to the VO detail after creating a new VO.
* resolves #350

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/create-vo-dialog/create-vo-dialog.component.ts
+++ b/apps/admin-gui/src/app/shared/components/dialogs/create-vo-dialog/create-vo-dialog.component.ts
@@ -4,6 +4,7 @@ import { NotificatorService } from '@perun-web-apps/perun/services';
 import { TranslateService } from '@ngx-translate/core';
 import { FormControl, Validators } from '@angular/forms';
 import { VosManagerService } from '@perun-web-apps/perun/openapi';
+import { Router } from '@angular/router';
 
 export interface CreateVoDialogData {
   theme: string;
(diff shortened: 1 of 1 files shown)


apps/admin-gui/src/app/shared/components/dialogs/create-vo-dialog/create-vo-dialog.component.ts
```diff
@@ -45,10 +47,11 @@ export class CreateVoDialogComponent implements OnInit {
 
   onSubmit() {
     this.loading = true;
-    this.voService.createVoWithName(this.fullNameCtrl.value, this.shortNameCtrl.value).subscribe(() => {
+    this.voService.createVoWithName(this.fullNameCtrl.value, this.shortNameCtrl.value).subscribe(data => {
```
xkureck, line 50: I have just a small note to the naming conventions.
You named "data" the object that the method returned. For better code readability I think it would be nicer if we rename it in more "describing way". Because what the method is returning is vo, so why not to name it vo?
```suggestion
    this.voService.createVoWithName(this.fullNameCtrl.value, this.shortNameCtrl.value).subscribe(vo => {
```

xkureck (changes requested):

apps/admin-gui/src/app/shared/components/dialogs/create-vo-dialog/create-vo-dialog.component.ts
HejdaJakub: It's true. Renamed!

xkureck (approved):

PR #578, opened 2020-12-10
Make existing member sponsored

bodnara:
* added option to make existing member sponsored in Vo


xkureck (changes requested):
Please make the dialog react to errors - eg. when in dialog you click on submit and error appears the expiration is reseted and user cannot click on submit for some reason
Also the loading for this dialog is weird - in the start user can see setting of the expiration and them it disappears when it is loading and then it just reappers

xkostka2 (changes requested):
image

When I sponsored an existing member I got this error.

image

I suggest adding some alert when no member is found.

xkureck (changes requested):
image
This is even more weird :D please when the dialog is loading leave here the input field then the search button, then there should be the mat-spinner and then you could leave here the whole sponsorship expiration section(label and the date picker - all of that)

xkostka2 (changes requested):
image

Please, hide or disable already sponsored members.

Also, after removing a sponsored member, there is still an error displayed, the case is the same as the previous error, that you already fixed. Sorry for not mentioning those problems before.

xkostka2 (approved):

xkureck (approved):

PR #579, opened 2020-12-11
Prevent proxy overload

bodnara:
* added very simple check to determine path validity unauthenticated user is trying to access
* added dialog that will pop-up if path is invalid, after clicking a button user will be redirected to home page which will trigger 
   authentication


xkureck (changes requested):
Please add to methods "isPotentiallyValidPath" and "verifyAuthentication" proper documentation(coment) of what that methods are doing

xkureck (approved):

PR #590, opened 2021-01-14
Groups tree optimization

Vojtech-Sassmann:
* The tree view of groups was not optimized for a large number of
groups. With the number of groups being high (e.g. 1000) the whole
application would lag and freeze.
* The tree view component has been reworked so it uses a virtual scoll.
This improves the performance by a lot. I have tested this changes on a
vo with 3000 groups and it works much better.
* I have also removed the check that the group description is shown only
if it doesn't fit. This check caused some errors because it used native
html properties to detect this. Also, this way the behaviour is more
consistent.

### DIFF
--- a/apps/admin-gui/src/styles.scss
+++ b/apps/admin-gui/src/styles.scss
@@ -1514,3 +1514,8 @@ th, td.mat-cell {
 .mat-tab-icon {
   font-size: 16px !important;
 }
+
+// hack to make the content of the virtual scroll be limited by its width
+.cdk-virtual-scroll-content-wrapper {
+  contain: size !important;
+}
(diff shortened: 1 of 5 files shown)


xkureck (commented):
I really like the optimization. On Vo is looks nice, but on some group-subgroup page the viewport is broken. eg. http://gui-dev.org/organizations/363/groups/6754/subgroups (muni-test backend)
image

Vojtech-Sassmann:
> I really like the optimization. On Vo is looks nice, but on some group-subgroup page the viewport is broken. eg. http://gui-dev.org/organizations/363/groups/6754/subgroups (muni-test backend)
> image

It should be fixed now.

xkureck (approved):

PR #596, opened 2021-01-17
Search select rework

Vojtech-Sassmann:
* Search selects now use a virtual scroll so they can work well with a
large number of entities.
* The search select logic was duplicated so I have created a new general
component that is used by all specific searchSelect components.
* The attribute page with members was loading all attributes for members
which was unnecessary since these attributes were not used. I have
specified a single core attribute to be returned. If an empty list was
passed, all attributes were loaded.


xkureck (changes requested):
image
I wanted to select Vo guests, so I clicked on it
image
and this is what happened, also fix the error messages please

xkureck (approved):

PR #597, opened 2021-01-18
Fix header checkbox "select all" in all tables

HejdaJakub:
* Header checkbox in all tables used to check all rows which included rows that wasn't be rendered.
* Now that header checkbox checks only that rows that are rendered. This implementation solves also page size, filtering and sorting.


bodnara (changes requested):
The implementation seems to be working pretty well, however it has some flaws regarding the header checkbox and master toggle function.

First thing is it assumes that user is only working on the first page of the table.  So if you go to other then the first page and use the master toggle it will deselect all check boxes and select whole first page instead. 

Another thing that I noticed that sometimes when using master toggle it selects items using different order and not all items on page are selected instead items from different pages can be selected as well. I can't explain this better but this is how to reproduce this issue.
Go to Perun admin -> attributes -> using page size of 10 -> filter "namespace" -> sort by column "Type" -> select header checkbox -> deselect header checkbox -> sort by column "Def" -> select header checkbox.  You should get something like this.
image

I couldn't quite figure what is the root of this problem.

HejdaJakub:
> The implementation seems to be working pretty well, however it has some flaws regarding the header checkbox and master toggle function.
> 
> First thing is it assumes that user is only working on the first page of the table. So if you go to other then the first page and use the master toggle it will deselect all check boxes and select whole first page instead.
> 
> Another thing that I noticed that sometimes when using master toggle it selects items using different order and not all items on page are selected instead items from different pages can be selected as well. I can't explain this better but this is how to reproduce this issue.
> Go to Perun admin -> attributes -> using page size of 10 -> filter "namespace" -> sort by column "Type" -> select header checkbox -> deselect header checkbox -> sort by column "Def" -> select header checkbox. You should get something like this.
> image
> 
> I couldn't quite figure what is the root of this problem.
> 
> Also in group-list and resource-list you added some logging. Please comment or remove them.

Functionality of header checkbox on next pages has been fixed. Also this special use-case with filter value has been fixed and now it seems that masterToggle function behaves correctly. I also removed forgotten console logs.

bodnara (commented):
The changes seem to work properly.

One thing I struggle with, is identifying what should be the proper behaviour for multi page selection.  Right now using master toggle on any page deselects checkboxes on other pages.
I would say that this is not something we would generally want. However I am not sure how should the master toggle behave when switching between pages with some selected items.

Maybe @xkureck could give her opinion on what the desired behaviour should look like.

xkureck (approved):
@bodnara The user when doing such a big operation can always change the paginating. But I believe such an operations are quite rare.

PR #600, opened 2021-01-21
Dashboard

xkureck:
* created special page for users that have only role 'SELF', this page is also configurable when added url of user profile
* added new component - recently viewed on main page, that shows 5 last visited entities
* tables are now under each other
* right menu will be implemented in next phase
* added recently viewed column to resources-list and group-list and sorting for that


bodnara (changes requested):
One minor issue is when checking if user has only role SELF don't count role PERUNADMIN. Otherwise you get the 'Manageable entities' subtitle but nothing there. It also causes the home icon before welcoming message to go missing for some reason. As seen here:
image

Another thing is that while resources show under your "manageable entities" they don't show in the recently view section.
When user have roles TRUSTEDFACILITYADMIN or RESOURCESELFSERVICE it is not shown on the dashboard. Maybe this is intentional, in that case just ignore this last comment.

xkureck:
> Another thing is that while resources show under your "manageable entities" they don't show in the recently view section.
When user have roles TRUSTEDFACILITYADMIN or RESOURCESELFSERVICE it is not shown on the dashboard. Maybe this is intentional, in that case just ignore this last comment.
All this is intentional, but I think it is good to notice that this could be confusing for users
The previous minor issue I will fix tomorrow

HejdaJakub (commented):
This implementation looks really good. Only one think I noticed (excepting the issue described by @bodnara) is that the "Recently viewed" icon (eye icon) at Access management and Facility management behaves correctly on refresh or on redirect from side menu, but it doesn't behaves correctly on back button or on back/forward button in browser (the icon doesn't show next to recently viewed entity or entities in this use-case).

xkureck:
> One minor issue is when checking if user has only role SELF don't count role PERUNADMIN. Otherwise you get the 'Manageable entities' subtitle but nothing there. It also causes the home icon before welcoming message to go missing for some reason.
fixed
@HejdaJakub thats because these two pages are cached when you go back and it is something we want

bodnara (commented):
Everything is working fine. One thing I suspect is that, in same rare case you could get stuck, with an entity you can no longer access, in your recently viewed list. Not sure if it can truly happen or if we want to address this in any way.

PR #601, opened 2021-01-22
Identities table fixes

xkureck:
* added mat sort header to all columns in user-ext-sources-list.component
* fixed paginating in table user-ext-sources-list.component

image


libs/perun/components/src/lib/user-ext-sources-list/user-ext-sources-list.component.ts
```diff
@@ -69,8 +83,36 @@ export class UserExtSourcesListComponent implements AfterViewInit, OnChanges {
   }
 
   setDataSource() {
-    this.displayedColumns = this.displayedColumns.filter(x => !this.hideColumns.includes(x));
     if (!!this.dataSource) {
+      this.dataSource.sortingDataAccessor = (item, property) => {
+        switch (property) {
+          case 'id' : {
+            if (item.userExtSource) {
+              return +item.userExtSource.id;
```
bodnara, line 91: the '+' seems redundant

bodnara (approved):
Looks good.

PR #603, opened 2021-01-26
fix add external identity to VO

xkureck:
* fixed adding external identity into the VO/Group - to this day user couldnt add candidate because we were sending the method wrong data


bodnara (changes requested):
Methods for creating member from candidate are called from the old members service component. They should be already rewritten into members service manager. Please use the members service manager to call them.
Other than that it looks good.

PR #606, opened 2021-01-29
Added filters for applications in Vo/Group

HejdaJakub:
* In Vo and Group has been added date filter which is by default set to display one year old applications.
* User can edit range of the date and this setting is also saved into local storage.
* In Vo has been also added other new filter which allows user to display only applicaions to Vo (without Groups applications).

### DIFF
--- a/apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.html
+++ b/apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.html
@@ -7,7 +7,7 @@
   <button mat-stroked-button *ngIf="showAllDetails" class="mr-2" (click)="this.showAllDetails = false">
     {{'VO_DETAIL.APPLICATION.SHOW_LESS_DETAILS' | translate}}
   </button>
-  <mat-form-field class="w-25 mr-2">
+  <mat-form-field class="input-width-250 mr-2">
     <mat-label>{{'VO_DETAIL.APPLICATION.TABLE_STATE' | translate}}</mat-label>
     <mat-select (selectionChange)="select()" [(value)]="state" disableOptionCentering>
       <mat-option value="all">{{'VO_DETAIL.APPLICATION.SELECTION_ALL' | translate}}</mat-option>
@@ -22,6 +22,16 @@
     [placeholder]="'VO_DETAIL.APPLICATION.FILTER'"
     (filter)="applyFilter($event)">
   </app-debounce-filter>
+  <perun-web-apps-date-range
+    [startDate]="startDate"
+    [endDate]="endDate"
+    (datePicker)="select()">
+  </perun-web-apps-date-range>
+  <div>
+    <mat-checkbox color="primary" [checked]="checked" (change)="showGroupApplications($event)">
+      {{'VO_DETAIL.APPLICATION.SHOW_GROUP_APPLICATIONS' | translate}}
+    </mat-checkbox>
+  </div>
 </div>
 <mat-spinner *ngIf="loading" class="ml-auto mr-auto"></mat-spinner>
 <div *ngIf="!loading">
--- a/apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.ts
+++ b/apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.ts
@@ -6,6 +6,8 @@ import {
   TABLE_VO_APPLICATIONS_DETAILED, TABLE_VO_APPLICATIONS_NORMAL,
   TableConfigService
 } from '@perun-web-apps/config/table-config';
+import { FormControl } from '@angular/forms';
+import { formatDate } from '@angular/common';
 
 @Component({
   selector: 'app-vo-applications',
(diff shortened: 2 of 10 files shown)


apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.html
```diff
@@ -18,10 +18,32 @@ <h1 class="page-subtitle">{{'VO_DETAIL.APPLICATION.TITLE' | translate}}</h1>
...
+      <input matInput [matDatepicker]="datePickerStart" [min]="minDate" [max]="maxDate" [formControl]="startDate" (dateChange)="dateChange()">
+      <mat-datepicker-toggle [for]="datePickerStart" matSuffix></mat-datepicker-toggle>
+      <mat-datepicker #datePickerStart></mat-datepicker>
+    </mat-form-field>
+    <mat-form-field>
+      <mat-label>{{'VO_DETAIL.APPLICATION.DATE.END' | translate}}</mat-label>
+      <input matInput [matDatepicker]="datePickerEnd" [min]="minDate" [max]="maxDate" [formControl]="endDate" (dateChange)="dateChange()">
+      <mat-datepicker-toggle [for]="datePickerEnd" matSuffix></mat-datepicker-toggle>
+      <mat-datepicker #datePickerEnd></mat-datepicker>
+    </mat-form-field>
```
xkureck, line 45: please create reusable component for datepicker that would copy the behaviour that is in the sponsored members page - dialog for creating 
image
create new component that you will use here and in the sponsored members and copy their behaviour

xkureck (changes requested):
image
The start date should be lower than end date

apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.html
HejdaJakub: I created separate component for date range and I added new condition which doesn't allow to set older end date than start date and newer start date than end date. I also replaced Group/VO applications selection with checkbox.

apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.ts
```diff
@@ -94,6 +108,25 @@ export class VoApplicationsComponent implements OnInit {
     }
   }
 
+  dateChange() {
+    this.dateObject = {startDate: this.startDate.value, endDate: this.endDate.value};
+    localStorage.setItem("dateObjectVo" + this.vo.id, JSON.stringify(this.dateObject));
```
xkureck, line 113: so the problem with this is that,  when I open this page on 2.2.2020 and change the start date to something(doesnt matter) then the start and end date will be saved to local host
problem will happen when I open this page on 3.2.2020(or any day after the 2.2.2020) and the end day will still be 2.2.2020, so if somebody made new application on 3.2 then the user will not see it - and you can bet that he/she will not notice the end date
I think the end date shouldn't be cached into the localStorage and I think that if the start date should be cached is questionmark, maybe it would be best if you ask Pavel Vyskocil/our support if the start date should be cached

xkureck (changes requested):
I think the ux has improved and for make changes in this we would need user testing, so for now its good

apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.ts
HejdaJakub: I discussed it with Pavel Vyskočil and in his opinion this dates shouldn't be cached, so I removed date storing in localStorage. Due to these changes I was able to simplify the code and remove dateChange() function in vo and group applications components. This function was replaced by already existing function select() in html.

apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.ts
```diff
@@ -82,7 +94,7 @@ export class VoApplicationsComponent implements OnInit {
         break;
       }
       case 'all': {
-        this.registrarManager.getApplicationsForVo(this.vo.id).subscribe(applications => {
+        this.registrarManager.getApplicationsForVo(this.vo.id, null, formatDate(this.startDate.value, 'yyyy-MM-dd','en-GB'), formatDate(this.endDate.value, 'yyyy-MM-dd','en-GB')).subscribe(applications => {
```
xkureck, line 97: on this place you miss the check to filter the applications for groups
you can rewrite part of code to 
`this.setData(null);
        break;`

xkureck: if you choose the option of rewriting, dont forget to do that in group-applications too

xkureck (changes requested):

apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.ts
HejdaJakub: Done.

xkureck (approved):

PR #607, opened 2021-02-04
Sort and filter changes

xkostka2:
* Listed changes applies to all list components
* Adjusted sort algorithm to sort czech characters properly
* Filter now checks only visible columns 
* Filter works case insensitive
* Latest OpenAPI changes were applied


libs/perun/components/src/lib/members-list/members-list.component.ts
```diff
@@ -83,12 +84,45 @@ export class MembersListComponent implements OnChanges, AfterViewInit {
...
     if (!!this.dataSource) {
-      this.dataSource.filterPredicate =
-        (data: RichMember, filter: string) => parseFullName(data.user).toLowerCase().includes(filter.toLowerCase()) ||
-          data.id.toString(10).includes(filter);
+      this.dataSource.filterPredicate = (data: RichMember, filter: string) => {
+        return customDataSourceFilterPredicate(data, filter, this.displayedColumns, this.getFilterDataForColumn, this);
+      };
+      this.dataSource.sortData = (data: RichMember[], sort: MatSort) => {
+        return customDataSourceSort(data, sort, this.getSortDataForColumn, this);
+      };
```
xkureck, line 125: image
this is the desired sorting in member-list when it comes to sorting according to names
primarily we want to sort according to what is in the "lastName" if last name is null then sort according to "firstName" if first name is null then sort according to something else so the method at least doesnt throw error

xkureck: image
this is your sorting according to full name, you are even counting the titles..

xkureck: same problem is everywhere where is fullName of user sorted - application-list, candidate-list

xkureck (changes requested):
please sort TS lint problems 
`/home/xkureck/perun-web-apps/libs/perun/utils/src/lib/perun-utils.ts:649:28
ERROR: 649:28  triple-equals  == should be ===
ERROR: 653:9   prefer-const   Identifier 'dataStrA' is never reassigned; use 'const' instead of 'let'.
ERROR: 654:9   prefer-const   Identifier 'dataStrB' is never reassigned; use 'const' instead of 'let'.
ERROR: 656:62  triple-equals  == should be ===

Lint errors found in the listed files.
`

libs/perun/openapi/src/lib/api/servicesManager.service.ts
```diff
@@ -1330,13 +1330,14 @@ export class ServicesManagerService {
...
      * @param reportProgress flag to report request and response progress.
      */
-    public deleteService(service: number, observe?: 'body', reportProgress?: boolean): Observable<any>;
-    public deleteService(service: number, observe?: 'response', reportProgress?: boolean): Observable<HttpResponse<any>>;
-    public deleteService(service: number, observe?: 'events', reportProgress?: boolean): Observable<HttpEvent<any>>;
-    public deleteService(service: number, observe: any = 'body', reportProgress: boolean = false ): Observable<any> {
+    public deleteService(service: number, force?: boolean, observe?: 'body', reportProgress?: boolean): Observable<any>;
+    public deleteService(service: number, force?: boolean, observe?: 'response', reportProgress?: boolean): Observable<HttpResponse<any>>;
+    public deleteService(service: number, force?: boolean, observe?: 'events', reportProgress?: boolean): Observable<HttpEvent<any>>;
+    public deleteService(service: number, force?: boolean, observe: any = 'body', reportProgress: boolean = false ): Observable<any> {
```
mvocu, line 1340: How is this related to the pull request objective?

libs/perun/openapi/src/lib/api/usersManager.service.ts
```diff
@@ -128,7 +128,7 @@ export class UsersManagerService {
 
     /**
      * Adds user\&#39;s external sources.
-     * @param addUserExtSourceInput
+     * @param addUserExtSourceInput
```
mvocu, line 131: Please do not commit empty space changes.

mvocu (commented):
Looks good to me (as far as I can tell), but contains unnecessary white-space changes and changes probably not related to the main purpose.

apps/admin-gui/src/app/vos/components/applications-list/applications-list.component.ts
```diff
@@ -75,6 +79,28 @@ export class ApplicationsListComponent implements OnChanges, AfterViewInit {
...
+        return data.id.toString();
+      case 'createdAt':
+        return data.createdAt;
+      case 'type':
+        return data.type;
+      case 'state':
+        return data.state;
+      case 'user':
+        return data.user ?
+          parseFullName(data.user) : data.createdBy.slice(data.createdBy.lastIndexOf('=')+1, data.createdBy.length);
```
xkureck, line 94: here we would also like to sort by the last name, if we got object User

libs/perun/components/src/lib/services-status-list/services-status-list.component.ts
```diff
@@ -124,6 +150,12 @@ export class ServicesStatusListComponent implements OnChanges, AfterViewInit {
             return item[property];
         }
       };
+      this.dataSource.filterPredicate = (data: ServiceState, filter: string) => {
+        return customDataSourceFilterPredicate(data, filter, this.displayedColumns, this.getDataForColumn, this)
+      };
+      this.dataSource.sortData = (data: ServiceState[], sort: MatSort) => {
+        return customDataSourceSort(data, sort, this.getDataForColumn, this);
+      };
```
xkureck, line 158: when sorting by ID - its not working - no error appears but it doesnt do anything
when sorting by end and start date the error in console aapears
image

xkureck (changes requested):
image
with this changes you break the functionality of "sorting according to recently viewed" 
the default sorting is set to column 'recent' in some tables(group, resource, facility, vo) so that the recent items are on top of the table - user have easy access to them

I guess that you can completely remove the
`this.dataSource.sortingDataAccessor = (item, property) => {`
method, because you are overriding the behaviour in the method sortData
I suggest to use the sortData function and then use the sortingDataAccessor to take into consideration the unusualness of each table

Also I think you should change the filter method - when you are adding one string to another like id + name + description the filter could find something that we dont want
image

libs/perun/components/src/lib/resources-list/resources-list.component.ts
```diff
@@ -81,26 +85,47 @@ export class ResourcesListComponent implements AfterViewInit, OnChanges {
...
-              }
-            }
-            return item.name.toLocaleLowerCase();
-          }
-          case 'name' : {
-            return item.name.toLocaleLowerCase();
-          }
-          default: return item[property];
         }
+        return data[column];
```
xkureck, line 106:
```suggestion
        return data['name'];
```

libs/perun/components/src/lib/groups-list/groups-list.component.ts
```diff
@@ -152,34 +158,41 @@ export class GroupsListComponent implements OnInit, AfterViewInit, OnChanges {
...
+      case 'expiration':
+        const expirationStr = getGroupExpiration(data);
+        return  parseDate(expirationStr);
+      case 'recent':
+        if (otherThis.recentIds) {
+          if (otherThis.recentIds.indexOf(data.id) > -1) {
+            return '#'.repeat(otherThis.recentIds.indexOf(data.id));
+          }
+        }
+        return data[column];
```
xkureck, line 180:
```suggestion
        return data['name'];
```

libs/perun/components/src/lib/facility-select-table/facility-select-table.component.ts
```diff
@@ -74,45 +78,38 @@ export class FacilitySelectTableComponent implements AfterViewInit, OnChanges {
...
+        return parseTechnicalOwnersNames(data.facilityOwners);
+      case 'recent':
+        if (otherThis.recentIds) {
+          if (otherThis.recentIds.indexOf(data.id) > -1) {
+            return '#'.repeat(otherThis.recentIds.indexOf(data.id));
           }
-          default: return item[property];
         }
-      };
+        return data[column];
```
xkureck, line 97:
```suggestion
        return data['name'];
```

xkureck (changes requested):

PR #612, opened 2021-02-09
Back button visibility

bodnara:
* back button is not visible when it would cause user to leave the app

### DIFF
--- a/apps/admin-gui/src/app/app.component.ts
+++ b/apps/admin-gui/src/app/app.component.ts
@@ -8,6 +8,8 @@ import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
 import { NewVersionDialogComponent } from './shared/components/dialogs/new-version-dialog/new-version-dialog.component';
 import { MatDialog } from '@angular/material/dialog';
 import { version } from '../../../../package.json';
+import { NavigationStart, Router } from '@angular/router';
+import { filter } from 'rxjs/operators';
 
 @Component({
   selector: 'app-root',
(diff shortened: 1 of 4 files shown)


HejdaJakub (changes requested):
I am not sure, that we want behavior like this. The back button disappear also on page reload. If I reload some page and go through the appliction via browser back button, our back button is disabled on every single page. In my opinion back button behaved correctly in most use cases, so you maybe should fix only the use case, when the back button tries to redirect user back to authentication.

bodnara (commented):
This solution is not optional, there is a at least one known scenario in which it will get broken. This scenario in described as comment in the code.

HejdaJakub (commented):
This scenario is acceptable for me. Worse use-case is, in my opinion, when user do all steps you described in your scenario and then will continue using browser back button. If he/she then will accesses the page A, this page won't be evaluated as initial, so our back button will appears on this page and after click on this button it will cause the primal error. But I think, that this use-case shouldn't be so frequent and if we want to cover all use-cases, then this task will be quite complicated, so I only want to leave this comment (for potential further discussion) and I agree with these changes.

xkureck (changes requested):
Fix the button visibility. I really dont like when my backbutton somehow disappears. Is the back button in Africa doing madagascar party or what?
Best would be if the backbutton appeared as "disabled" second best solution would be if the back button looked normally but did nothing if user clicked on it on onitPage.

bodnara (commented):
I don't really see why the back button couldn't enjoy free time it has earned by hard work and go partying to Africa.
But ok, it will be displayed and not functioning because the disabled option looks quite odd.

xkureck (approved):

PR #615, opened 2021-02-17
Search facility by destination or host

xkostka2:
* Added columns 'Destinations' and 'Hosts' displaying destinations and hosts of a facility
* Facility select table now uses EnrichedFacility objects instead of RichFacility
* Generic component for displaying list values of objects was created
* Generic pipe for filtering unique values from list of objects was created


Vojtech-Sassmann (changes requested):
There is a performance problem during loading the page. Even though there is pagination for the table when the given data are set for the first time, the pagination is not initialized. Thus, angular renders rows for all of the elements. Then immediately, when the pagination is loaded, the table is rerendered only with elements from the single page. 

This problem has probably with almost all of the tables within our application. We have discovered this already when dealing with a large number of applications in the applications list component. 

I have created a commit, which should fix this issue. Please, study it well to fully understand, what is happening, so you can deal with it next time on your own. I am only providing the solution for you, because it took me a lot of time, when I was dealing with it for the first time: https://github.com/CESNET/perun-web-apps/commit/0e47539a3930c30411da509ff4afcc172fb7b888

Vojtech-Sassmann (approved):

PR #620, opened 2021-02-23
Admin - core attributes

Vojtech-Sassmann:
* The core attributes were filtered on the admin page, however, we want
to be able to see them.
* Also, we want to be able to create core attributes.


apps/admin-gui/src/app/admin/pages/admin-page/admin-attributes/admin-attributes.component.ts
```diff
@@ -75,7 +75,7 @@ export class AdminAttributesComponent implements OnInit {
   refreshTable() {
     this.loading = true;
     this.attributesManager.getAllAttributeDefinitions().subscribe(attrDefs => {
-      this.attrDefinitions = filterCoreAttributesDefinitions(attrDefs);
```
xkureck, line 78: then the function "filterCoreAttributesDefinitions" can be deleted from imports and since it is not used anywhere it can be completely deleted from perun-utils right?

Vojtech-Sassmann: Yes, good point. I have removed it.

xkureck (approved):

PR #623, opened 2021-02-26
AdminGUI - new sponsorhip API

Vojtech-Sassmann:
* The API for the method used to create sponsored members has changed.
* Fixed alert colors on the dialog that was shown after a creation of a
sponsored user.
* Regenerated OpenAPI.


bodnara (changes requested):
I am getting an error when trying to create sponsored member. It suggests the password is missing even when it is provided or the generate password option is selected. 
This may be problem on the core part of this issue: https://github.com/CESNET/perun/pull/3106
image

libs/perun/openapi/src/lib/api.module.ts
```diff
@@ -49,7 +49,7 @@ import { VosManagerService } from './api/vosManager.service';
     VosManagerService ]
 })
 export class ApiModule {
-    public static forRoot(configurationFactory: () => Configuration): ModuleWithProviders<any> {
+    public static forRoot(configurationFactory: () => Configuration): ModuleWithProviders {
         return {
```
bodnara, line 53: Also this change is causing error: ` ERROR in ../libs/perun/openapi/src/lib/api.module.ts:52:71 - error TS2314: Generic type 'ModuleWithProviders<T>' requires 1 type argument(s).`

Vojtech-Sassmann:
> I am getting an error when trying to create sponsored member. It suggests the password is missing even when it is provided or the generate password option is selected.
> This may be problem on the core part of this issue: [CESNET/perun#3106](https://github.com/CESNET/perun/pull/3106)
> image

I forgot to mention this in the PR description. This PR should be merged together with https://github.com/CESNET/perun/pull/3106 These changes have not been merged yet and therefore it is failing.

libs/perun/openapi/src/lib/api.module.ts
Vojtech-Sassmann: This is weird, I don't know why this line has changed. Anyway, I have removed this change from the PR.

bodnara (approved):

Vojtech-Sassmann:
@xkureck This PR should not have been merged yet, as the changes to the backend are still not finished. I have mentioned this in one of the comments above. This will now cause the functionality to fail on devel and it definitely should not be released for production. I don't know if we shouldn't revert this, but we have to be really careful here, so we don't put it into production without the changes to the backend.

PR #630, opened 2021-03-04
Support different namespaces for sponsorship

bodnara:
* supported namespaces are now loaded from configuration
* 'empty' namespace must be allowed in the instance config


apps/admin-gui/src/app/shared/components/dialogs/create-sponsored-member-dialog/create-sponsored-member-dialog.component.ts
```diff
@@ -54,27 +59,56 @@ export class CreateSponsoredMemberDialogComponent implements OnInit {
...
 
   ngOnInit(): void {
     this.loading = true;
     this.theme = this.data.theme;
-    this.parseNamespace();
-    if (this.namespaceOptions.length === 0) {
-      this.functionalityNotSupported = true;
+
+    this.membersService.getAllNamespacesRules().subscribe(rules => {
+      if (this.store.get('allow_empty_sponsor_namespace')) {
```
Vojtech-Sassmann, line 71: I would add this property to the default config with a value `false` so it is clear, that such value exists.

Vojtech-Sassmann (changes requested):
Just a minor problem, otherwise it looks ok.

Vojtech-Sassmann (approved):

PR #631, opened 2021-03-05
Add sponsored member to groups when created

bodnara:
* when creating sponsored members in bulk, there is now option to add them to existing groups or even new group
* added stepper to dialog for creating sponsored members in bulk for better orientation


Vojtech-Sassmann (changes requested):
I think this works pretty well. But I have one thing, that I would say we can do better now. Could you validate the email format immediately on the first step? This is kinda weird when I have filled out all of the information and on the last page, it just sends me back to the first one, because I did something wrong there.

Vojtech-Sassmann (approved):
I think that this should work well. Although, the filter is still lagging the whole app. But I have discovered, it is not because of these changes. The problem is in the `isAllSelectedWithDisabledCheckbox` method in the `TableCheckbox` service. This method is taking too much time to process. We should definitely fix it as soon as possible.

PR #635, opened 2021-03-10
ExtSource management added to group section

HejdaJakub:
* ExtSource page was added to group section.
* I also fixed the add ExtSource dialog in VO section, where was not shown correct data.


bodnara (changes requested):
Overall it is well done. However, there are some changes needed before merging. 
Please add ext. sources page to the side-menu in group section.
I assume you followed the structure of ext. sources page in organizations section which is great. Unfortunately this page along side with the add ext. source dialog had some issues on their own.  
First of all it is missing the remove dialog completely. Take a look at some other page on how the remove dialog looks like and create one that will be used for both organizations and groups. It should be fairly straight forward.
Secondly it makes asynchronous calls in a for loop which I believe should be replaced with a recursive function (take a look how it is done on service destination page with allowing/blocking). But honestly none of those two are perfect and it can maybe stay the way it is now.
Last thing I would recommend is instead of creating a new add dialog for groups, just make some changes in the existing one to support both options. These dialogs are mostly identical and it leads to code duplication which we obviously try to avoid. (you can take inspiration from add member dialog for this)

HejdaJakub:
> Overall it is well done. However, there are some changes needed before merging.

- External sources was added to the side menu.
- Remove dialog was created and added to both sections.
- Async function call in for loop was replaced by recursoin.
- I modified the old add dialog and now both sections are using the same dialog.

apps/admin-gui/src/app/shared/side-menu/side-menu-item.service.ts
```diff
@@ -897,6 +897,15 @@ export class SideMenuItemService {
       });
     }
 
+    //SettingsExtSources
+    if (this.authResolver.isAuthorized('getGroupUnions_Group_boolean_policy', [group])) {
+      settingsChildrenLinks.push({
```
bodnara, line 902: Use the same policy as in the overview.

apps/admin-gui/src/app/vos/pages/group-detail-page/group-settings/group-settings-extsources/group-settings-extsources.component.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  onAdd() {
+    const config = getDefaultDialogConfig();
+    config.width = '1000px';
+    config.data= {
+      voId: this.voId,
+      groupId: this.groupId,
+      extSources: this.extSources,
+      theme: 'vo-theme'
```
bodnara, line 88: 'group-theme'

apps/admin-gui/src/app/vos/pages/group-detail-page/group-settings/group-settings-extsources/group-settings-extsources.component.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  onRemove() {
+    const config = getDefaultDialogConfig();
+    config.width = '600px';
+    config.data= {
+      voId: this.voId,
+      groupId: this.groupId,
+      extSources: this.selection.selected,
+      theme: 'vo-theme'
```
bodnara, line 106: 'group-theme'

bodnara (changes requested):
It is looking good. There are only a couple of really small issues.

apps/admin-gui/src/app/shared/side-menu/side-menu-item.service.ts
HejdaJakub: Policy fixed.

apps/admin-gui/src/app/vos/pages/group-detail-page/group-settings/group-settings-extsources/group-settings-extsources.component.ts
HejdaJakub: Done.

apps/admin-gui/src/app/vos/pages/group-detail-page/group-settings/group-settings-extsources/group-settings-extsources.component.ts
HejdaJakub: Done.

bodnara (approved):

PR #636, opened 2021-03-10
Set different sponsor when creating sponsored member

bodnara:
* added option to choose different sponsor when user is authorized to do so
* implemented stepper to "create sponsored member" dialog

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/create-sponsored-member-dialog/create-sponsored-member-dialog.component.ts
+++ b/apps/admin-gui/src/app/shared/components/dialogs/create-sponsored-member-dialog/create-sponsored-member-dialog.component.ts
@@ -1,13 +1,16 @@
 import { Component, Inject, OnInit } from '@angular/core';
 import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
 import {
+  AuthzResolverService,
   InputCreateSponsoredMember,
   MembersManagerService,
   NamespaceRules,
   RichMember,
+  RichUser,
+  User,
   UsersManagerService
 } from '@perun-web-apps/perun/openapi';
-import { ApiRequestConfigurationService, StoreService } from '@perun-web-apps/perun/services';
+import { ApiRequestConfigurationService,GuiAuthResolver, StoreService } from '@perun-web-apps/perun/services';
 import {
   AsyncValidatorFn,
   FormControl,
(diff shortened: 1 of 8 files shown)


xkostka2 (changes requested):
image

Disable button when the user is not selected, this happens for example when Vo has no sponsors.

image

Even tho I am Perun admin, it does not show me on the list of potential sponsors.

xkostka2 (changes requested):
image

After displaying the dialog this error pops out. Although, this error won't pop up in production, according to best practices it shouldn't be there. Try to get rid of it if possible.

image

I suggest moving text in the alert into the tooltip of Create button which will be disabled if there are no sponsors.

apps/admin-gui/src/app/shared/components/dialogs/create-sponsored-member-dialog/create-sponsored-member-dialog.component.ts
```diff
@@ -81,7 +90,28 @@ export class CreateSponsoredMemberDialogComponent implements OnInit {
...
+      const complementaryObject = {
+        id: this.data.voId,
+        beanName: 'Vo'
+      }
+      const availableRolesPrivileges = new Map<string, any>();
+
+      this.guiAuthResolver.getRolesAuthorization(availableRoles, complementaryObject, availableRolesPrivileges);
+      this.findSponsorsAuth = availableRolesPrivileges.get(availableRoles[0]).readAuth;
+
+      if (this.findSponsorsAuth) {
```
xkureck, line 104: if this.findSponsorsAuth is false then dont show at all the select field for making somebody else sponsor

xkureck (changes requested):
image
Actually between all the section insert some gaps please

image
In that try I was sponsor + vo observer so I could find for other sponsor in the organization so I have choosen somebody else and it throw at me priviledge exception
Please make visible the select box to only the users that have the priviledge to make somebody else sponsor

Another problem for me with the ux is that If I want to make me the sponsor then I dont really know what to do on this page
image
Why is jana Prepechalova set as a sponsor as default(why is first on list set as default sponsor) then if I want to make myself as sponsor I need to find me on the list. I think that the default option should be me(the user). Another solution could be that we could solve this by radio buttons where default chosen option would be "make myself sponsor" and the other would be the select field with selecting somebody else the sponsor.

Vojtech-Sassmann (approved):

xkureck (changes requested):
image
Hi there is still problem with authorization. On this account I have role Sponsor + vo admin and I have choosen as sponsor somebody else. Which I cant do.
According to the policy
`createSponsoredMember_Vo_String_Map<String_String>_String_User_LocalDate_policy:
    policy_roles:
      - SPONSOR: Vo
        SELF: User
    include_policies:
      - default_policy`
I must have the role sponsor and the sposor must be me or I must be perun admin. Please show only the selection of sponsor only to those who have right to make somebody else sponsor. Or even better make the option "select other sponsor from this organization" disabled with tooltip that says the user dont have the priviledge to make somebody else sponsor.

xkostka2 (changes requested):
image

Check one of the radio buttons by default so this does not happen.

image

I selected a sponsor, then selected myself and then checked to select another sponsor. Clear the value or display it. 

image

Just a minor suggestion. Mark fields Login and Password as required.

bodnara:
> Just a minor suggestion. Mark fields Login and Password as required.

These fields can be disabled, optional or reqiured depending on the configuration and selected namespace.

xkostka2 (approved):

xkureck (approved):

PR #640, opened 2021-03-15
Table checkbox optimization

xkostka2:
* Table checkbox service was optimized

### DIFF
--- a/libs/perun/services/src/lib/table-checkbox.service.ts
+++ b/libs/perun/services/src/lib/table-checkbox.service.ts
@@ -28,11 +28,12 @@ export class TableCheckbox {
     this.pageIterator = 0;
     this.dataLength = filter === '' ? dataSource.data.length :
       dataSource.filteredData.length;
+    if (!nextPage) {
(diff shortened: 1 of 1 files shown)


libs/perun/services/src/lib/table-checkbox.service.ts
```diff
@@ -112,7 +107,7 @@ export class TableCheckbox {
           });
         } else {
           dataSource.sortData(dataSource.data, sort).forEach(row => {
-            if (dataSource.sortData(dataSource.filteredData, sort).includes(row)) {
+            if (dataSource.filteredData.includes(row)) {
```
HejdaJakub, line 110: The condition `dataSource.filteredData.includes(row)` doesn't work correctly. I already tried it in issue #597 and concretely for this line you can try the use case described in the issue (use case for attributes with filter in Perun admin).

HejdaJakub (changes requested):

HejdaJakub (approved):

PR #643, opened 2021-03-17
AdminGui - create sponsored user pwd check

Vojtech-Sassmann:
* When creating a single sponsored user, the password is now validated
using the backend.
* User is shown a detailed error if the backend returns an error.
* Also, the password is now hidden unless the user decides to show it.


Vojtech-Sassmann:
For reviewers: you can test these changes against the mu-test instance and the mu namespace.

xkureck (commented):
As a basic user I would probably not expected that more information about why the password doesn't meet the selected namespace's criteria is in the red alert so the direct message why its not good password would miss me. Why the message cant be under the input field as usually is?
Otherwise its good. You can merge it if the above irrelevant.

Vojtech-Sassmann:
> As a basic user I would probably not expected that more information about why the password doesn't meet the selected namespace's criteria is in the red alert so the direct message why its not good password would miss me. Why the message cant be under the input field as usually is?
> Otherwise its good. You can merge it if the above irrelevant.

Hi, I did it this way, because it was part of the specification for this issue. However, you make a good point. I will ask the requestor, if we can rework this so the whole message would be visible.

PR #653, opened 2021-03-24
Find users without organization

bodnara:
* added option to list users that are not member of any organization to  perun admin - users section

### DIFF
--- a/apps/admin-gui/src/app/admin/pages/admin-page/admin-users/admin-users.component.html
+++ b/apps/admin-gui/src/app/admin/pages/admin-page/admin-users/admin-users.component.html
@@ -18,6 +18,13 @@
           (click)="onSearchByString()">
     <mat-icon>search</mat-icon>
   </button>
+  <button
+    mat-stroked-button
+    class="mr-2"
+    [disabled]="loading"
+    (click)="findUsersWithoutVO()">
+    {{'ADMIN.USERS.USERS_WITHOUT_VO' | translate}}
+  </button>
   <mat-spinner *ngIf="loading" class="ml-auto mr-auto">
   </mat-spinner>
   <div *ngIf="!loading">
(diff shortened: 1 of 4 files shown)


libs/perun/openapi/src/lib/api.module.ts
```diff
@@ -49,7 +49,7 @@ import { VosManagerService } from './api/vosManager.service';
     VosManagerService ]
 })
 export class ApiModule {
-    public static forRoot(configurationFactory: () => Configuration): ModuleWithProviders<any> {
+    public static forRoot(configurationFactory: () => Configuration): ModuleWithProviders {
```
HejdaJakub, line 52: ModuleWithProviders requires one type argument. You can use `<any>` or according to official angular guide `<ApiModule>` in this case.

HejdaJakub (changes requested):
This implementation looks good, I have only one comment.

libs/perun/openapi/src/lib/api.module.ts
bodnara: Fixed, for some reason this argument gets removed every time openAPI is generated.

HejdaJakub (approved):

apps/admin-gui/src/app/admin/pages/admin-page/admin-users/admin-users.component.html
```diff
@@ -18,6 +18,14 @@ <h1 class="page-subtitle">{{'ADMIN.USERS.TITLE' | translate}}</h1>
           (click)="onSearchByString()">
     <mat-icon>search</mat-icon>
   </button>
+  <button
+    mat-flat-button
+    color="primary"
+    class="mr-2"
```
xkureck, line 24:
```suggestion
    mat-stroked-button
```
the red almost burned my eyes, otherwise its good

xkureck (changes requested):

apps/admin-gui/src/app/admin/pages/admin-page/admin-users/admin-users.component.html
bodnara: red colour was removed to prevent further eye damage

xkureck (approved):

PR #654, opened 2021-03-26
Implemented sections My publications and All publications

HejdaJakub:
* All publications section is allowed only for CABINETADMIN and PERUNADMIN.
* My publications section displays only user's publications.
* Both sections can be filtered and users can also delete publications.


apps/publications/src/app/pages/my-publications-page/my-publications-page.component.html
```diff
@@ -1 +1,29 @@
...
+      <button
+        mat-flat-button
+        color="warn"
+        class="mr-2"
+        [disabled]="selected.selected.length === 0"
+        (click)="removePublication()">
+        {{'AUTHOR_DETAIL.REMOVE' | translate}}
+      </button>
+      <perun-web-apps-publication-filter (filteredPublication)="filterPublication($event)"></perun-web-apps-publication-filter>
+    </div>
```
bodnara, line 18: Please create a separate placeholder "MY_PUBLICATIONS.REMOVE" for the remove button. Same in All publications.

bodnara (changes requested):
Other than one small issue, the implementation looks good. However I would like to ask you if you could fix the problem with redirecting in publication-list when the "unlocked" icon is clicked. It is done for "locked" icon, so it will take just couple of seconds.

apps/publications/src/app/pages/my-publications-page/my-publications-page.component.html
HejdaJakub: Placeholders edited and redirecting fixed.

bodnara (approved):

PR #655, opened 2021-03-29
Sponsor already sponsored members

HejdaJakub:
* Now users can sponsor some member who is already sponsored by someone else.
* In sponsor existing member dialog is new column that represent state if member is already sponsored or not.


bodnara (changes requested):
When trying to sponsor already sponsored member I get "AlreadySponsoredMemberException".  I assume the reason is that method 'setSponsorshipForMember' transforms non-sponsored member to sponsored one and cannot be used on already sponsored members. Only other method I found in API documentation that seems like it could be used for this purpose is "sponsorMember" but I am not sure.

bodnara (approved):

xkureck (approved):

PR #657, opened 2021-03-30
Publication detail

bodnara:
* implemented page for publication detail
* implemented publication editing
* adding/removing of authors and acknowlegments is not part of this PR


xkostka2 (changes requested):
image

Make the input fields (where it makes sense) wider, so the whole value is shown if possible. 

image

I suggest providing an example of how the correct value looks like.

image

I suggest disabling years above the current one.

image

I suppose this is yet to be implemented. If not, don't show the disabled edit button.

I would also add some tooltip for disabled edit buttons, saying why they are all disabled.

Also, why do u reload both the publications list component and the whole publication detail page when any field is saved? I mean why both at once?

xkureck (changes requested):
please make the publication detail page accesible from pages: "All publications" and "My publications"

image
I wouldnt mind if the buttons "edit" and buttons "add" and "remove" in authors and acknowledgement sections were completely invisible if the publication is locked

image
also you forget to add to publication detail table "created date" and "full cite". also 
image
this row is missing

xkostka2 (changes requested):
image

Fix the width of this input, please.

I also suggest that spinner will appear only on affected tables, not on all tables.

xkureck (changes requested):
The icon next to All publication is much more smaller than on page All Publication
image
When I click on the lock/unlock its weird that the loaders also are shown in the tables of authors and Acknowledgements
I like the big edit/save button
Also you didnt fix the urls and headers of the publication detail page
when I go from all publications - url is fine, even tho there dont have to be 'publication' in it I think its redundat, but w/e
image
when I go from my publications - the header of the page is wrong and wrong is also the url
image
when I go from authors page, the page detail is fine

xkostka2 (approved):

xkureck (approved):

PR #658, opened 2021-03-30
Displaying attribute values in profile

xkostka2:
* Improved displaying of longer values of attributes in the Profile page
* List and map attribute values now show all their elements by default on the Profile page


apps/user-profile/src/app/pages/profile-page/profile-page.component.html
```diff
@@ -11,7 +11,7 @@
         <div [matTooltip]="translateService.currentLang === 'en' ? attribute.tooltip_en : attribute.tooltip_cz"
              matTooltipPosition="above">{{translateService.currentLang === 'en' ? attribute.displayName_en : attribute.displayName_cz}}</div>
       </td>
-      <td [ngSwitch]="attribute.attribute.friendlyName">
+      <td [ngSwitch]="attribute.attribute.friendlyName" class="w-75">
```
Vojtech-Sassmann, line 14: This is not sufficient to fix the issue:
image

The problem is, that the width of the whole table is not set to the whole available space.

Vojtech-Sassmann (approved):

PR #659, opened 2021-03-31
Password reset implementation

bodnara:
* it is now possible to change password in Password reset app
* added password strength check
* added token validity check


xkureck (changes requested):
for password activation it gave me this error 
image

image
Another thing - on password activation you dont check if the token is valid?

Also I can see that on onSubmit method you dont actually call the method for changing the password, even in the description of the task it was mentioned that its part of the task. 

image
Also probably different text for activation would be good.

Ok I see that your part of the task was to do the password reset part,  I just realized that. So I guess only the part with that the submit button is not working because the code is commented is valid.

xkureck (changes requested):
so one small things - the size of the tooltips text is super small, please make it bigger
image
also some error maybe happened when I clicked on submit, it was tested on devel, so maybe this only happens on devel?

xkureck (approved):

PR #660, opened 2021-03-31
dynamic paginating in vo-members page implemented

xkureck:
* initially members of vo in vo-members page are dynamically acquired from database
* now backend is supporting options for sorting according to id and names
* buttons "list all" and "search" do not support this because the backend does not support this yet
* implemented according to https://blog.angular-university.io/angular-material-data-table/


Vojtech-Sassmann (approved):
That looks good. We should consider some extension of this functionality (selecting members with a certain status, etc.)

PR #663, opened 2021-04-01
Group overview statistics

cuadradek:
- Added two statistic cards to group overview - members counts by group
and VO status.

### DIFF
--- a/apps/admin-gui/src/app/vos/pages/group-detail-page/group-overview/group-overview.component.ts
+++ b/apps/admin-gui/src/app/vos/pages/group-detail-page/group-overview/group-overview.component.ts
@@ -38,6 +38,12 @@ export class GroupOverviewComponent implements OnInit {
   parentGroup: Group = null;
   loading = false;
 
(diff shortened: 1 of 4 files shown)


apps/admin-gui/src/app/vos/pages/group-detail-page/group-overview/group-overview.component.ts
```diff
@@ -52,6 +58,23 @@ export class GroupOverviewComponent implements OnInit {
...
+          Object.entries(stats).forEach(([status, count]) =>
+            this.membersCountsByVoStatus.set(status.toLowerCase(), count)
+          );
+        }, () => this.loading = false);
+
+        this.groupService.getGroupMembersCountsByGroupStatus(this.groupId).subscribe(stats => {
+          Object.entries(stats).forEach(([status, count]) =>
+            this.membersCountsByGroupStatus.set(status.toLowerCase(), count)
+          );
+        }, () => this.loading = false);
```
xkureck, line 77: we usually want to make this asynchronous operations chained - so that on the end of the chain is loading = false and we can say that every information is loaded
but this section will be redesigned in this sprint

xkureck (approved):

PR #665, opened 2021-04-07
Create publication layout

bodnara:
* added layout to create publication page
* implementation of importing/creating publication is not part of this PR

### DIFF
--- a/apps/publications/src/app/pages/create-publication-page/create-publication-page.component.html
+++ b/apps/publications/src/app/pages/create-publication-page/create-publication-page.component.html
@@ -1,1 +1,40 @@
-<p>create-publication-page works!</p>
+<div class="user-theme">
+  <h1 class="page-subtitle d-flex">
+    <span>
+    <mat-icon class="add-icon mr-1">add_box</mat-icon>
+    {{'CREATE_PUBLICATION.TITLE' | translate}}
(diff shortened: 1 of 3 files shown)


apps/publications/src/app/pages/create-publication-page/create-publication-page.component.html
```diff
@@ -1 +1,38 @@
-<p>create-publication-page works!</p>
+<div class="user-theme">
+  <h1 class="page-subtitle d-flex">
+    <mat-icon class="add-icon">add_box</mat-icon>
+    <span class="ml-3">{{'CREATE_PUBLICATION.TITLE' | translate}}</span>
```
xkureck, line 4: image
The page title is a little bit too high than we would like it, It should be on the same "line" as the icon, or at least centered

xkureck (changes requested):

apps/publications/src/app/pages/create-publication-page/create-publication-page.component.html
bodnara: Title and icon were aligned.

xkureck (changes requested):
check slack

xkureck (approved):

PR #667, opened 2021-04-08
Group/Vo statistics page

bodnara:
* moved statistics from overview to new page for both Vo and Groups
* Vo statistics are loaded via **old methods** and should be updated as soon as the new method is implemented


xkureck (commented):
So in my opinion everything looks nice and cool.
One thing you should do - ask Adrian and other people from our team(on our daily scrum) where the page with statistics should be situated - eg. in my opinion it should be hided in the advanced settings, because its not that important and also there is already a lot of 'cubes' on the main overview
If other people would tell you that this is fine, the PR can be merged

PR #668, opened 2021-04-08
creating sponsored members in bulk is now respecting Namespace rules

xkureck:


Vojtech-Sassmann (approved):
I think this looks good. I would have only one suggestion. I think we could change the text `Create in bulk` to something like `Generate sponsored users`. I think it might be more clear, what it actually does.

PR #677, opened 2021-04-19
Displaying membership type implemented

xkostka2:
* Added column to members-list containing membership type
* Disabled removing of indirect members
* This is a preparation for further changes


HejdaJakub (commented):
Everything looks good, I have only one minor comment. As we discussed, the rows with indirect members can be displayed in italic font type for better UX (this font type for indirect members is displayed in old gui). But this is only a tip which is not necessary.

xkureck (changes requested):
I also have suggestion that the status would be presented by some icon, icon for indirect membership and no icon for direct membership, no label for headercelldef, the icon would have tooltip with "indirect membership" also I am not sure If I would like that column as last column
Also please consider if user experience would be improved if the disabled checkbox would have some tooltip with description why is checkbox disabled

xkureck (approved):

HejdaJakub (approved):

PR #687, opened 2021-04-30
Dialog for displaying indirect member's groups implemented

xkostka2:
* Added dialog for displaying paths to member's direct groups
* Newest OpenAPI was generated

### DIFF
--- a/libs/perun/components/src/lib/members-list/members-list.component.ts
+++ b/libs/perun/components/src/lib/members-list/members-list.component.ts
@@ -224,7 +224,7 @@ export class MembersListComponent implements OnChanges, AfterViewInit {
 
   viewMemberGroupTree(member: RichMember) {
     const config = getDefaultDialogConfig();
-    config.width = '500px';
+    config.width = '800px';
     config.maxHeight = '500px';
     config.data = {member: member, groupId: this.groupId};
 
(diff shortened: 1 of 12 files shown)


libs/perun/components/src/lib/members-list/members-list.component.ts
```diff
@@ -170,6 +175,12 @@ export class MembersListComponent implements OnChanges, AfterViewInit {
     this.displayedColumns = this.displayedColumns.filter(x => !this.hideColumns.includes(x));
     this.dataSource = new MatTableDataSource<RichMember>(this.members);
     this.setDataSource();
+    this.disabledRouting = this.disableRouting;
```
balcirakpeter, line 178: I am just curious. Is this line needed?

Vojtech-Sassmann (changes requested):
* The dialog should definitely be wider, the group names can take a lot of space, there is no point in making it this small.
* I don't understand the meaning of the House icon, or the men with arrows icon? I think you wanted to distinguish between a subgroup and group inclusion, but this is just confusing. There is no explanation, what they mean.

libs/perun/components/src/lib/members-list/members-list.component.ts
xkostka2: disabledRouting is used to disable routing when clicking on a button located on a row

balcirakpeter: I know the purpose of the variable. What Ido not understand is why do you assign this variable to itself.

PR #689, opened 2021-05-04
Embedded application settings conditional

balcirakpeter:
- Changing embedded group application approval is now possible only when
  autoregistration is enabled.
- Also visibility of approval is now relying on the same condition.
- Fixed icons allignment.


Vojtech-Sassmann (approved):

xkureck (approved):
I dont understand why it should work that way, but technically it works

PR #691, opened 2021-05-06
Dialog for displaying indirect member's origin 

xkostka2:
* Added dialog for displaying paths to member's direct groups
* Dialog support filtering the paths to direct/included groups
* Newest OpenAPI was generated

### DIFF
--- a/apps/admin-gui/src/assets/i18n/en.json
+++ b/apps/admin-gui/src/assets/i18n/en.json
@@ -2457,8 +2457,13 @@
           "SUBMIT_BUTTON": "Submit"
         },
         "MEMBER_TREEVIEW_DIALOG": {
-          "TITLE": "Member tree view",
-          "CLOSE": "Close"
+          "TITLE": "Indirect membership’s origin of {{name}}",
+          "INFO": "Member is coming to this group through the following group hierarchy.",
+          "DIRECT": "(Direct membership)",
+          "INCLUDE": "(Included group)",
+          "FILTER": "Search by group name",
+          "CLOSE": "Close",
+          "NO_PATHS": "No paths include the entered group name."
         }
       },
       "ORGANIZATIONS": {
--- a/apps/admin-gui/src/assets/i18n/en.json
+++ b/apps/admin-gui/src/assets/i18n/en.json
@@ -2455,6 +2455,10 @@
           "ASK": "Do you want to proceed?",
           "CANCEL_BUTTON": "Cancel",
           "SUBMIT_BUTTON": "Submit"
+        },
+        "MEMBER_TREEVIEW_DIALOG": {
+          "TITLE": "Member tree view",
+          "CLOSE": "Close"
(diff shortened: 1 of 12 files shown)


xkureck (changes requested):
image
1. The search field is completely different than what we have in rest of the application. Please make this search field consistent with other search fields. Look how it works in tree view, I would like to have the same behaviour
2. align the icon with the name of the group
3. only one scrollbar please :)

libs/perun/dialogs/src/lib/member-tree-view-dialog/member-tree-view-dialog.component.html
```diff
@@ -0,0 +1,74 @@
+<div class="group-theme">
+  <h1 mat-dialog-title>{{'SHARED_LIB.PERUN.COMPONENTS.MEMBER_TREEVIEW_DIALOG.TITLE' | translate}}</h1>
+  <mat-spinner *ngIf="loading" class="ml-auto mr-auto"></mat-spinner>
+  <div *ngIf="!loading" mat-dialog-content class="mr-0">
+    <p>{{'SHARED_LIB.PERUN.COMPONENTS.MEMBER_TREEVIEW_DIALOG.INFO' | translate}}</p>
+
+    <mat-form-field class="filter-field ml-2">
+      <input matInput [formControl]="formControl"
+             placeholder="{{'SHARED_LIB.PERUN.COMPONENTS.MEMBER_TREEVIEW_DIALOG.FILTER' | translate}}">
+    </mat-form-field>
```
xkureck, line 10: The filter field itself is small, I would make it bigger - or reuse the filter field we normally use

xkureck (changes requested):
image
The filtering is kind of meh. When I type the name of some of the group(which is there) it just shows me the same as if I did nothing. The filtering just feels useless now. Inspire how the filtering works in the treeview of groups and ideally reuse it if its possible. 
Also why is always the root of the tree the group name of the group where we are? Wouldnt it make more sense if the root group wouldnt be here, since we know in which group we are?
image
Like this

xkureck (commented):
Also do we even need filtering? Even in VO like ceitec there is not that much groups

xkureck (changes requested):
Put a name of the user on the dialog title and its fine

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -2455,6 +2455,15 @@
           "ASK": "Do you want to proceed?",
           "CANCEL_BUTTON": "Cancel",
           "SUBMIT_BUTTON": "Submit"
+        },
+        "MEMBER_TREEVIEW_DIALOG": {
+          "TITLE": "Indirect membership’s origin of {{name}}",
```
xkureck, line 2460: didnt know that this is possible

xkureck (approved):

PR #693, opened 2021-05-08
Embedded form item type just once

balcirakpeter:
- Embedded form item types can be added just once to the form.


balcirakpeter:
@xkureck I've made suggested changes but I had to also re-generate the openApi specification.

xkureck (approved):

PR #695, opened 2021-05-11
add branding

xkureck:
* instance config now can have property <domainName>_branding and inside this can be any property that is able to be configured
* method "get" now looks for properties in branding(if the branding exist) then it looks on the instanceConfig and then it looks on the defaultConfig
TESTING: the testing domain is gui-dev.org and gui-dev2.org, you can test the branding on these two domain, you need to change settings in apache2, allowed hosts, and also default config(oidc) 
eg. how the instance config can look: `"gui-dev2.org_branding": {
    "footer_support_mail": "gui2 mail test"
  },
  "gui-dev.org_branding": {
    "footer_support_mail": "idm@mail.muni.cz",
  },`

### DIFF
--- a/libs/perun/services/src/lib/store.service.ts
+++ b/libs/perun/services/src/lib/store.service.ts
@@ -102,7 +102,6 @@ export class StoreService {
           currentValue = this.defaultConfig[keys[i]];
         } else {
           if (currentValue === undefined) {
-            console.error('Missing value in default config: ' + keys);
             break;
           }
           currentValue = currentValue[keys[i]];
--- a/libs/perun/services/src/lib/store.service.ts
+++ b/libs/perun/services/src/lib/store.service.ts
@@ -102,6 +102,7 @@ export class StoreService {
           currentValue = this.defaultConfig[keys[i]];
         } else {
           if (currentValue === undefined) {
+            console.error('Missing value in default config: ' + keys);
             break;
           }
           currentValue = currentValue[keys[i]];
--- a/libs/perun/services/src/lib/store.service.ts
+++ b/libs/perun/services/src/lib/store.service.ts
@@ -67,7 +67,7 @@ export class StoreService {
     let currentValue: string;
 
     if (this.branding !== '') {
-      const brandingConfig = this.instanceConfig[this.branding];
+      const brandingConfig = this.instanceConfig['brandings'][this.branding];
       for (let i = 0; i < keys.length; ++i) {
         if (i === 0) {
           currentValue = brandingConfig[keys[i]];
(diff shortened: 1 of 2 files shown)


Vojtech-Sassmann (commented):
Hmm, I'm thinking about the format of the config and I think, that if we would keep all brandings in the config under the same property, it might be a bit easier to navigate. Something like:

```JSON
{
  "brandings": {
    "gui-dev2.org": {
      ...
    }, 
    "einfra.org": {
      ...
    }
  }
}
```
What do you think? :thinking:

xkureck:
> Hmm, I'm thinking about the format of the config and I think, that if we would keep all brandings in the config under the same property, it might be a bit easier to navigate. Something like:
> 
> ```json
> {
>   "brandings": {
>     "gui-dev2.org": {
>       ...
>     }, 
>     "einfra.org": {
>       ...
>     }
>   }
> }
> ```
> 
> What do you think? thinking

Since Pavel put the thumbs up emoji I changed the format. I guess it is more easily readable this way

libs/perun/services/src/lib/store.service.ts
```diff
@@ -61,7 +66,21 @@ export class StoreService {
...
+          }
+          currentValue = currentValue[keys[i]];
+        }
+      }
+    }
+
+    if (this.instanceConfig !== undefined && currentValue === undefined) {
       for (let i = 0; i < keys.length; ++i) {
         if (i === 0) {
           currentValue = this.instanceConfig[keys[i]];
```
Vojtech-Sassmann, line 86: A couple of lines below is an error printed to the console if the configuration for the branding is not found. Could you please change it, so this error is not printed whilst search the `brandings` property? 
```
    console.error('Missing value in default config: ' + keys);
```

Vojtech-Sassmann (changes requested):

libs/perun/services/src/lib/store.service.ts
```diff
@@ -83,7 +102,6 @@ export class StoreService {
           currentValue = this.defaultConfig[keys[i]];
         } else {
           if (currentValue === undefined) {
-            console.error('Missing value in default config: ' + keys);
```
Vojtech-Sassmann, line 86: Well, this log was there for a reason. I think it should print some error if some default value is not specified, since it almost certainly shows that we have some bug in the code. I just didn't want it to print error, when the branding is not found, since that is OK and we expect it to happen.

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann (approved):

PR #698, opened 2021-05-12
adding/removing acknowledgements from detail of publication

xkureck:
* added add dialog for adding acknowledgements
* reusing universalRemove dialog for removing acknowledgements


xkostka2 (changes requested):
image

From this, I have no clue who created it.

Also, it took me a while to realize the list of what is displayed in add dialog and what the column "Type" means (whether it has some impact on acknowledgement). If target group of users should already know this then you can ignore this comment.

xkureck:
> image
> 
> From this, I have no clue who created it.
> 
> Also, it took me a while to realize the list of what is displayed in add dialog and what the column "Type" means (whether it has some impact on acknowledgement). If target group of users should already know this then you can ignore this comment.

I spoke with @zlamalp about both things, the first thing that you mentioned should be that way, the second thing I fixed

xkostka2 (approved):

PR #700, opened 2021-05-13
support of login screen added to publications

xkureck:


Vojtech-Sassmann (changes requested):
The implementation looks good I am just wondering if we need to duplicate the `publications-login-screen-base` component? I think it would be better to use the already existing `perun-web-apps-login-screen-base` component and if you want the book icon to appear, you can add some input parameter which can be set.

libs/perun/services/src/lib/auth.service.ts
```diff
@@ -72,13 +72,6 @@ export class AuthService {
     this.manager.events.addUserLoaded(user => {
       this.user = user;
     });
-
-    this.manager.events.addAccessTokenExpired(() => {
-      const config = getDefaultDialogConfig();
-      config.width = '450px';
-
-      this.dialog.open(SessionExpirationDialogComponent, config);
-    });
```
xkureck, line 81: check this. Was it there for some reason? Is it safe to delete it?

Vojtech-Sassmann: I think so. Probably, it is used to display a dialog when the access token has expired. It shouldn't be probably removed.

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann (changes requested):
Please, keep the SessionExpirationDialogComponent. Otherwise, I think these changes are ok.

Vojtech-Sassmann (approved):

PR #703, opened 2021-05-13
feat(profile): added login screen

Vojtech-Sassmann:
* Added support for login screen to the user profile.

### DIFF
--- a/apps/user-profile/src/app/components/profile-login-screen-base/profile-login-screen-base.component.html
+++ b/apps/user-profile/src/app/components/profile-login-screen-base/profile-login-screen-base.component.html
@@ -1,0 +1,8 @@
+<div class="app-min-width">
+  <perun-web-apps-header [hideToggle]="true" [disableLogo]="true"></perun-web-apps-header>
+  <div [ngStyle]="{'min-height': getContentHeight()}" id="content-wrap">
+    <router-outlet></router-outlet>
+  </div>
+
+  <perun-web-apps-footer-user-profile></perun-web-apps-footer-user-profile>
+</div>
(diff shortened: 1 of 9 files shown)


Vojtech-Sassmann:
Unfortunately, I didn't manage to use the shared base layout component, because the profile is using its custom footer and the base layout is a bit different than in the other applications. In the future, it would be good to unify the footer and use one footer for all of our apps.

apps/user-profile/src/app/components/profile-login-screen-base/profile-login-screen-base.component.html
```diff
@@ -0,0 +1,9 @@
+<div class="app-min-width">
+  <perun-web-apps-header [hideToggle]="true"></perun-web-apps-header>
+  <div [ngStyle]="{'min-height': getContentHeight()}" id="content-wrap">
+    <perun-web-apps-breadcrumbs></perun-web-apps-breadcrumbs>
```
xkureck, line 4: I dont think it needs to be here

apps/user-profile/src/app/components/profile-login-screen-base/profile-login-screen-base.component.html
```diff
@@ -0,0 +1,9 @@
+<div class="app-min-width">
+  <perun-web-apps-header [hideToggle]="true"></perun-web-apps-header>
```
xkureck, line 2: The logo in this component is clickable

xkureck (changes requested):

apps/user-profile/src/app/components/profile-login-screen-base/profile-login-screen-base.component.html
Vojtech-Sassmann: You are right, I have removed it.

apps/user-profile/src/app/components/profile-login-screen-base/profile-login-screen-base.component.html
Vojtech-Sassmann: Fixed.

xkureck (approved):

PR #705, opened 2021-05-18
Passing query parameters through middle click

xkostka2:


Vojtech-Sassmann (changes requested):
This, in fact, fixes the problem when using the `PerunWebAppsMiddleClickRouterLink` component. However, on some places we use the combination of the `<a>` element with `[routerLink]` set. For these instances, the problem still exists. This can be fixed by also specifying the `[queryParamsHandling]="'merge'"` attribute next to the `[routerLink]`.

You can search all of such places using the IDE's search everywhere feature, with regex like `<a[^<>]*\[routerLink][^<>]*>`

I'm also wondering, why do we even have the component `MiddleClickRouterLinkDirective`? :thinking: I think it would make much more sense to simply use the `<a>` elements with `[routerLink]` set. This way, we would also be able to open links in new windows using `ctrl + mouse click` and we would not have to hack the middle click feature by ourselves. We already use this approach in multiple places (like the buttons on entities' overviews) and it works just fine. I think we might consider removing this special component if there was no specific reason why we did it this way.

Vojtech-Sassmann (approved):

PR #706, opened 2021-05-20
Fix - matTooltip didn't vanish after routing

HejdaJakub:
* In cache-route-reuse-strategy is new check, which is cleaning old matTooltips during the routing.


Vojtech-Sassmann (approved):
I really like how simple this solution is. @HejdaJakub so the problem was caused by our caching, or did you just put the code here for another reason?

I'm just a little worried if this won't cause some additional weird bugs, since Angular doesn't like manipulating the dom directly. However, for now, it seems to work well.

HejdaJakub:
> I really like how simple this solution is. @HejdaJakub so the problem was caused by our caching, or did you just put the code here for another reason?

What I found, the problem is probably in material library and this is the most recommended workaround solution on GitHub (https://github.com/angular/components/issues/11478). I am little confused why some tooltips disappear and other don't, but I think, that this solution should cover all cases.

PR #707, opened 2021-05-20
Delete entities from detail

bodnara:
* groups, facilities and resource can now be deleted from detail


HejdaJakub (approved):

xkureck:
On group detail it works fine. Just one thing - we would probably want this to have the same dialog also for deleting group on the vo-groups page when we want to delete group there, just so we are consistent
image

xkureck:
The same problem is with the resources - please make sure that - if it makes sence - these dialog are consistent = looks the same

xkureck (changes requested):
image
image

On this places I would expect the same dialog as it is on resource detail

image
On this place I would expect the completely same dialog as it is in the group detail

xkureck (changes requested):
The button cancel move to "right" and the remove group dialog change the label - with force remove group we are only deleting the subgroups and the members and resources we are not deleting we are just removing the connection to that group of these object and the labels on that dialog could be confusing with this

xkureck (approved):

PR #715, opened 2021-05-25
Displayed item properties hidden/disabled on the form configuration and form preview

HejdaJakub:
* To the form configuration were added two columns which display hidden and disabled item properties. In columns are also displayed hidden and disabled dependencies.
* To the form preview were added icons which refer to hidden and disabled item properties. In the icon tooltip is dependency info.
* If some item has property disabled='ALWAYS', then is shown as disabled on the form preview.

### DIFF
--- a/apps/admin-gui/src/app/vos/components/application-form-list/application-form-list.component.ts
+++ b/apps/admin-gui/src/app/vos/components/application-form-list/application-form-list.component.ts
@@ -1,4 +1,4 @@
-import {Component, Input, OnChanges, SimpleChanges, ViewChild, Output, EventEmitter} from '@angular/core';
+import { Component, Input, OnChanges, SimpleChanges, ViewChild, Output, EventEmitter, OnInit } from '@angular/core';
 import { MatDialog } from '@angular/material/dialog';
 import { MatTable } from '@angular/material/table';
 import {CdkDragDrop, moveItemInArray} from '@angular/cdk/drag-drop';
@@ -19,7 +19,7 @@ import { Router } from '@angular/router';
   templateUrl: './application-form-list.component.html',
   styleUrls: ['./application-form-list.component.scss']
 })
-export class ApplicationFormListComponent implements OnChanges {
+export class ApplicationFormListComponent implements OnInit, OnChanges {
 
   constructor(private dialog: MatDialog,
               private notificator: NotificatorService,
@@ -39,7 +39,7 @@ export class ApplicationFormListComponent implements OnChanges {
   theme: string;
 
   @Input()
-  displayedColumns: string[] = ['drag', 'shortname', 'type', 'preview', 'managegroups', 'edit', 'delete'];
+  displayedColumns: string[] = ['drag', 'shortname', 'type', 'disabled', 'hidden', 'preview', 'managegroups', 'edit', 'delete'];
 
   @Output()
   applicationFormItemsChange = new EventEmitter<ApplicationFormItem[]>();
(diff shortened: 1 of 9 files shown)


apps/admin-gui/src/app/vos/components/application-form-list/application-form-list.component.ts
```diff
@@ -57,6 +57,44 @@ export class ApplicationFormListComponent implements OnChanges {
...
 
+  disabledDependency(item: ApplicationFormItem): String {
+    let mess: String;
+    let dep: String;
+    switch (item.disabled) {
+      case 'ALWAYS':
+        return "";
+      case 'IF_PREFILLED':
+        mess = this.translate.instant('VO_DETAIL.SETTINGS.APPLICATION_FORM.DISABLED_HIDDEN_ICON.IF_PREFILLED');
+        dep = this.applicationFormItems.find(i => i.id === item.disabledDependencyItemId).shortname;
```
Vojtech-Sassmann, line 68: This will throw an error if the dependency is not specified. The find will return `undefined` it cannot access `.shortname` on it. The same applies for the other use-cases.

Vojtech-Sassmann (changes requested):
Some other thoughts:
* If the item is ALWAYS hidden, does it make sense to show it on preview? In the old gui, I think such items were not shown.
* I think it might be worth adding a tooltip (mat-hint) to the eye icon on the items list component (not in the preview). At least I expected something to be shown, but that's just my opinion. There could be a longer description like: `This item is hidden if the item 'blabla' was prefilled with empty value`

Vojtech-Sassmann (changes requested):
Well, the functionality seems to be working pretty well, I have no comments about that. However, it is a bit hard to navigate in the core. It is not clear what are the possible code flows. I can't even see how this prevents some code duplications. In general, you should avoid methods that have such complex nested structures. It is better to create multiple methods where it is on the first look clear, what it does in which scenarios.

 I suggest rewriting this so we can save some work of debugging this code in the future. :slightly_smiling_face: 

One tip: not sure if you will be able to do this but you can avoid some of the switch statements for getting the translations. E.g.: you can get the translation like this:

instead of this:
```typescript
switch(item.disabled)
  case 'ALWAYS':
    msg = this.translate.instant('VO_DETAIL...ALWAYS');
    ...
  case 'NEVER':
    msg = this.translate.instant('VO_DETAIL...NEVER');
    ...
```

You can use: 
```typescript
    msg = this.translate.instant('VO_DETAIL...' + item.disabled);
```

HejdaJakub:
> I suggest rewriting this so we can save some work of debugging this code in the future. 

I agree with you with you, that the previous implementation wasn't well readable. I rewrote it and I hope, that now it's more clear. :)

apps/admin-gui/src/app/vos/components/application-form-list/application-form-list.component.ts
```diff
@@ -52,11 +52,60 @@ export class ApplicationFormListComponent implements OnChanges {
...
+  ifEmpty = this.translate.instant('VO_DETAIL.SETTINGS.APPLICATION_FORM.DISABLED_HIDDEN_ICON.IF_EMPTY');
+  ifPrefilled = this.translate.instant('VO_DETAIL.SETTINGS.APPLICATION_FORM.DISABLED_HIDDEN_ICON.IF_PREFILLED');
+
+  // tooltips for hidden and disabled icons
+  alwaysDisabled = this.translate.instant('VO_DETAIL.SETTINGS.APPLICATION_FORM.DISABLED_HIDDEN_ICON.ALWAYS_DISABLED_HINT');
+  alwaysHidden = this.translate.instant('VO_DETAIL.SETTINGS.APPLICATION_FORM.DISABLED_HIDDEN_ICON.ALWAYS_HIDDEN_HINT')
+  isDisabledIf = this.translate.instant('VO_DETAIL.SETTINGS.APPLICATION_FORM.DISABLED_HIDDEN_ICON.DISABLED_IF_HINT');
+  isHiddenIf = this.translate.instant('VO_DETAIL.SETTINGS.APPLICATION_FORM.DISABLED_HIDDEN_ICON.HIDDEN_IF_HINT');
+  isEmpty = this.translate.instant('VO_DETAIL.SETTINGS.APPLICATION_FORM.DISABLED_HIDDEN_ICON.IS_EMPTY_HINT');
+  isPrefilled = this.translate.instant('VO_DETAIL.SETTINGS.APPLICATION_FORM.DISABLED_HIDDEN_ICON.IS_PREFILLED_HINT');
```
Vojtech-Sassmann, line 65: Such initialization should be performed in the `onInit` method. It might be also a cause, why some of the translations are not properly loaded. 

image

Vojtech-Sassmann (changes requested):
I think it is much easier now to navigate through the code :slightly_smiling_face:. However, I have found a problem with the translations.

apps/admin-gui/src/app/vos/components/application-form-list/application-form-list.component.ts
HejdaJakub: Initialization moved to `ngOnInit()` method.

PR #716, opened 2021-05-26
Create publication page

bodnara:
* added page for single publication creation
* created separate components for adding authors and acknowledgements


apps/publications/src/app/pages/create-publication-page/create-single-publication-page/create-single-publication-page.component.ts
```diff
@@ -0,0 +1,177 @@
...
+      publication: {
+        title: this.publicationControl.get("title").value,
+        categoryId: this.publicationControl.get("category").value.id,
+        year: this.publicationControl.get("year").value.year,
+        isbn: this.publicationControl.get("ISBN").value,
+        doi: this.publicationControl.get("DOI").value,
+        main: this.publicationControl.get("cite").value,
+      }
+    };
+    this.cabinetService.createPublication(publicationInput).subscribe(publication => {
```
xkureck, line 115: please add to input property "createdBy" with the email of the user who is creating the publication

xkureck (changes requested):
I would prefer horizontal stepper, but I guess thats just a personal opinion.
image
With this info I would like if you inspired from the old publications 
image
I liked here the caps lock yes and no, it feels like it more structured

In old gui we have checkbox where user can define if he wants to be the author of the publication right when creating the publication
image
The reason behind this that if you are not perun admin or cabinet admin or author of the publication then you cannot edit the publication(i am not sure - but you cannot also add authors/acknowledgements - which is problem right in the next steps).

bodnara:
> In old gui we have checkbox where user can define if he wants to be the author of the publication right when creating the publication
> image
> The reason behind this that if you are not perun admin or cabinet admin or author of the publication then you cannot edit the publication(i am not sure - but you cannot also add authors/acknowledgements - which is problem right in the next steps).

I have added checkbox for this purpose. However I tested it and adding authors and acknowledgements is possible with just role SELF.

xkureck (approved):

PR #718, opened 2021-05-27
fix(admingui): fixed url destination regex

Vojtech-Sassmann:
* The url destination regex was invalid. Changed to the same regex which is now used on backend.


xkureck (commented):
Tried this url from the test of backend and it passes our check but it shouldnt pass the test
image

Vojtech-Sassmann:
> Tried this url from the test of backend and it passes our check but it shouldnt pass the test
> image

Nice catch! There was a missing `$` symbol at the end of the pattern.

xkureck (approved):

PR #722, opened 2021-05-31
Invalid preferred mail change request handling

xkostka2:
* In case user is redirected to user-profile or admin-gui from link with invalid query parameters for changing preferred mail, dialog with info is displayed and query parameters are cleared

### DIFF
--- a/apps/admin-gui/src/app/users/pages/user-detail-page/user-dashboard/user-dashboard.component.ts
+++ b/apps/admin-gui/src/app/users/pages/user-detail-page/user-dashboard/user-dashboard.component.ts
@@ -5,10 +5,18 @@ import {
   User,
   UsersManagerService
 } from '@perun-web-apps/perun/openapi';
-import { GuiAuthResolver, NotificatorService, StoreService } from '@perun-web-apps/perun/services';
+import {
+  ApiRequestConfigurationService,
+  GuiAuthResolver,
+  NotificatorService,
+  StoreService
+} from '@perun-web-apps/perun/services';
 import { SideMenuService } from '../../../../core/services/common/side-menu.service';
 import { ActivatedRoute, Router } from '@angular/router';
 import { TranslateService } from '@ngx-translate/core';
+import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
+import { MailChangeFailedDialogComponent } from '@perun-web-apps/perun/dialogs';
+import { MatDialog } from '@angular/material/dialog';
 
 @Component({
   selector: 'app-perun-web-apps-user-dashboard',
@@ -28,7 +36,9 @@ export class UserDashboardComponent implements OnInit {
               private route: ActivatedRoute,
               private router: Router,
               private notificator: NotificatorService,
-              public translateService: TranslateService
+              public translateService: TranslateService,
+              private dialog: MatDialog,
+              private apiRequestConfiguration: ApiRequestConfigurationService
   ) {
     translateService.get('USER_DETAIL.DASHBOARD.MAIL_CHANGE_SUCCESS').subscribe(res => this.mailSuccessMessage = res);
   }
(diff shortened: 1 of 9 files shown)


apps/admin-gui/src/app/users/pages/user-detail-page/user-dashboard/user-dashboard.component.ts
```diff
@@ -5,10 +5,18 @@ import {
...
+  ApiRequestConfigurationService,
+  GuiAuthResolver,
+  NotificatorService,
+  StoreService
+} from '@perun-web-apps/perun/services';
 import { SideMenuService } from '../../../../core/services/common/side-menu.service';
 import { ActivatedRoute, Router } from '@angular/router';
 import { TranslateService } from '@ngx-translate/core';
+import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
+import { MailChangeFailedDialogComponent } from '../../../../../../../../libs/perun/dialogs/src/lib/mail-change-failed-dialog/mail-change-failed-dialog.component';
```
Vojtech-Sassmann, line 18: I think that all imports from the library should start with `@perun-web-apps/...`

apps/user-profile/src/app/pages/profile-page/profile-page.component.ts
```diff
@@ -12,7 +12,8 @@ import {
 import { UserFullNamePipe } from '@perun-web-apps/perun/pipes';
 import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
 import { ActivatedRoute, Router } from '@angular/router';
-import { NotificatorService, StoreService } from '@perun-web-apps/perun/services';
+import { ApiRequestConfigurationService, NotificatorService, StoreService } from '@perun-web-apps/perun/services';
+import { MailChangeFailedDialogComponent } from '../../../../../../libs/perun/dialogs/src/lib/mail-change-failed-dialog/mail-change-failed-dialog.component';
```
Vojtech-Sassmann, line 16: Same here.

libs/perun/dialogs/src/lib/mail-change-failed-dialog/mail-change-failed-dialog.component.ts
```diff
@@ -0,0 +1,22 @@
...
+export class MailChangeFailedDialogComponent implements OnInit {
+
+  constructor(private dialogRef: MatDialogRef<MailChangeFailedDialogComponent>,
+              private router: Router) { }
+
+  ngOnInit(): void {
+  }
+
+  onClose() {
+    this.router.navigate([], {queryParamsHandling: ''})
```
Vojtech-Sassmann, line 19: I'm not sure if you want to always redirect to `/`. I think you should navigate to the same path, where the user already is, only changes the parameters. Also, the `queryParamsHandling` parameter has only two valid options: 'merge' and 'preserve'. When you use `''` I think it's the same as not specifying it at all?

Vojtech-Sassmann (changes requested):

libs/perun/dialogs/src/lib/mail-change-failed-dialog/mail-change-failed-dialog.component.ts
xkostka2: Since the link that you get in the confirmation email points to / and logic for handling pref mail change is implemented on home pages for both user-profile and admin-gui, there is no reason to redirect elsewhere. As for the queryParamsHandling, you are right, it can be omitted.

Vojtech-Sassmann: Ok, we can keep it this way.

libs/perun/dialogs/src/lib/mail-change-failed-dialog/mail-change-failed-dialog.component.html
```diff
@@ -0,0 +1,12 @@
+<h1 mat-dialog-title>{{'SHARED_LIB.PERUN.COMPONENTS.MAIL_CHANGE_FAILED_DIALOG.TITLE'| translate}}</h1>
+<div mat-dialog-content>
+  {{'SHARED_LIB.PERUN.COMPONENTS.MAIL_CHANGE_FAILED_DIALOG.INFO'|translate}}
```
Vojtech-Sassmann, line 3: I'm just thinking if we could use here the `app-alert` component, which we have implemented for such notifications. This simple text doesn't look good and I think that if it would be with red color, the user would notice that it is a message for some error. What do you think?

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann (approved):

PR #725, opened 2021-06-01
edit of sponsorship added into the member overview

xkureck:


Vojtech-Sassmann (changes requested):
The implementation looks good. However, I would suggest some minor changes to be done as well:
*  `DIALOGS.EDIT_MEMBER_SPONSORS.CANCEL` label from `Cancel` to `Close` since you are not actually canceling any action.
* Not updating the list of sponsors when users don't change anything. I have actually come across this when viewing all sponsored members in MU vo. There is a lot of them and when you accidentally click on some member sponsors and immediately close the dialog, the whole table is refreshed.

Vojtech-Sassmann (changes requested):
When the last sponsor of a member is removed, the application throws an error, because it tries to load all member sponsors even though the member is not sponsored anymore. After the removal of a sponsor, the application should check again, if the member is still sponsored.

Vojtech-Sassmann (approved):

PR #726, opened 2021-06-08
change of initial value of select status in page vo-members to no sta…

xkureck:
…tus selected initially


xkostka2 (commented):
image

Just a suggestion, wouldn't it make more sense if we replaced "Select status" with "Filter by status"?

xkostka2 (approved):

PR #728, opened 2021-06-10
Added import for publications from external systems

HejdaJakub:
* Now user can import publications from external pub systems.
* If user imports some publications, then he or she should completes information about each publication and mark publication as complete.
* If all publications are marked as complete, then user can finish import.


bodnara (changes requested):
This looks good. The only thing I would probably change is the UI of completing/incompleting publication. With multiple publications all the green and red areas feel a little disturbing. Although this is just my opinion.

bodnara (commented):
Small suggestions, I would probalby prefer the single publication check button to be on the left. It can be overlooked on the right. Similary I would consider putiing the mark all button on top instead on the bottom. However this is just my opinion and you can decide against it.

**Don't forget to change the testing code for the correct one.**

bodnara (approved):

HejdaJakub:
> **Don't forget to change the testing code for the correct one.**

Testing code was changed for the correct code before merge.

PR #729, opened 2021-06-10
Pagination changes

xkostka2:
* Added pagination to the top of each table
* Offered page size options reduced to  5, 10, 25, 50, 100


xkureck:
image
The margin of the top paginator and the export button must be resolved. How it looks now its not good. Also With your changes the number of items showed in the paginator is 0 in the dynamic tables

Vojtech-Sassmann:
> image
> The margin of the top paginator and the export button must be resolved. How it looks now its not good. Also With your changes the number of items showed in the paginator is 0 in the dynamic tables

I would suggest moving the export button next to the page size selection, so it would be on a single like. We could probably create some kind of a shared component that could be used in all tables. What do you think? @xkureck

libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.ts
```diff
@@ -26,6 +26,7 @@ import {
 } from '@perun-web-apps/perun/services';
 import { merge } from 'rxjs';
 import { tap } from 'rxjs/operators';
+import { TableWrapperComponent } from '../table-wrapper/table-wrapper.component';
```
xkureck, line 29:
```suggestion
import { TableWrapperComponent } from '@perun-web-apps/perun/components';
```
Idea told me that you can shorten the import
do it on every table

apps/admin-gui/src/app/vos/components/applications-list/applications-list.component.ts
```diff
@@ -143,12 +145,12 @@ export class ApplicationsListComponent implements OnChanges, AfterViewInit {
...
   setDataSource() {
     // don't set the data if the paginator was not loaded yet
-    if (!this.paginator) {
+    if (!this.child.paginator) {
       return;
     }
     if (!this.dataSource) {
       this.dataSource = new MatTableDataSource<Application>();
-      this.dataSource.paginator = this.paginator;
+      this.dataSource.paginator = this.child.paginator;
```
xkureck, line 153: sorting in this table seems to stop working in your PR

apps/admin-gui/src/app/vos/components/application-list-details/application-list-details.component.ts
```diff
@@ -54,7 +55,7 @@ export class ApplicationListDetailsComponent implements OnChanges {
 
   addedColumns = new Set<string>();
 
-  @ViewChild(MatPaginator) paginator: MatPaginator;
+  @ViewChild(TableWrapperComponent, {static: true}) child: TableWrapperComponent;
```
xkureck, line 58: with your PR in this table some weird console errors are appearing
image

apps/admin-gui/src/app/shared/components/entityless-attribute-keys-list/entityless-attribute-keys-list.component.ts
```diff
@@ -46,7 +46,7 @@ export class EntitylessAttributeKeysListComponent implements OnChanges, OnInit {
     this.setDataSource();
   }
 
-  @ViewChild(MatPaginator) paginator: MatPaginator;
+  @ViewChild(TableWrapperComponent, {static: true}) child: TableWrapperComponent;
```
xkureck, line 49: image
To get to this do: perun admin -> attributes -> click on some entityless attribute definition -> in dialog click on show keys(left down)

libs/perun/components/src/lib/table-wrapper/table-wrapper.component.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  public paginator: MatPaginator;
+  @Input()
+  hideExport = false;
+  @Output()
+  page = new EventEmitter<PageEvent>();
+  @Input()
+  pageSizeOptions = TABLE_ITEMS_COUNT_OPTIONS;
+  @Input()
+  pageSize = 5
```
xkureck, line 20: This component could actually handle all the logic when pageSize is changed.
Instead of getting pagesize as input, you can get tableId as input and if the user change the pageSize the component can right here change it because it has all it needs to write it to localhost
Try to think about it. If you this this is too big change we can do it in the next task/in next spring

apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
```diff
@@ -47,7 +47,7 @@ export class SettingsAuthenticationComponent implements OnInit {
     translate.get('AUTHENTICATION.DELETE_IMG_DIALOG_DESC').subscribe(res => this.removeDialogDescription = res);
   }
 
-  @ViewChild(MatPaginator) paginator: MatPaginator;
+  @ViewChild(TableWrapperComponent, {static: true}) child: TableWrapperComponent;
```
xkureck, line 50: image
On user profile happens this. No idea if it is caused by your PR

xkureck (changes requested):
image
There is still some circular dependency?

libs/perun/components/src/lib/table-wrapper/table-wrapper.component.ts
xkostka2: Yes, you are right, I suggest making a new task out of it since it affects a lot of files (lists + their parent components) and I won't be able to do it this sprint.

xkureck (approved):

PR #731, opened 2021-06-14
add user and facility-user attributes as tab to member attributes page

xkureck:
* new component one-entity-attribute-page.component added for page where we are viewing attributes of one entity

### DIFF
--- a/apps/admin-gui/src/app/vos/pages/member-detail-page/member-attributes/member-attributes.component.html
+++ b/apps/admin-gui/src/app/vos/pages/member-detail-page/member-attributes/member-attributes.component.html
@@ -5,44 +5,8 @@
       <ng-template matTabLabel>
         {{'MEMBER_DETAIL.ENTITY' | translate}}
       </ng-template>
-      <perun-web-apps-refresh-button class="mt-2" (refresh)="refreshTable()"></perun-web-apps-refresh-button>
-      <button
-        *ngIf="!loading"
-        mat-flat-button
-        class="mr-2 mt-2"
-        color="accent"
-        (click)="onCreate()">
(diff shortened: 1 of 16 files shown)


apps/admin-gui/src/app/vos/pages/member-detail-page/member-attributes/member-attributes.component.html
```diff
@@ -63,7 +27,25 @@ <h1 class="page-subtitle">{{'MEMBER_DETAIL.SETTINGS.ATTRIBUTES.TITLE' | translat
...
-                                       [secondEntity]="'resource'" *ngIf="!loading"></app-two-entity-attribute-page>
+                                       [secondEntity]="'resource'"></app-two-entity-attribute-page>
+      </ng-template>
+    </mat-tab>
+    <mat-tab class="mb-2">
+      <ng-template matTabLabel>
+        {{'ADMIN_USER.ENTITY' | translate}}
+      </ng-template>
+      <app-one-entity-attribute-page [entityId]="this.member.userId" [entity]="'user'">
+      </app-one-entity-attribute-page>
```
bodnara, line 38: Wrap the body with \<ng-tamplate matTabContent\> tag so it is lazily loaded. Otherwise the member can be undefined when accessing userId.

bodnara (changes requested):
Other than the small fix it looks good. I like that you created separate component for the single entity attributes and reduced code duplicity. However, I have one suggestion and that is to keep in this component only such logic that is truly shared among all of the attribute pages. This would remove the stacking of the 'switch' statment. Possible solution would be to make the loading of attributes in the 'entity-attributes' components and pass them as input to the 'one-entitiy-attributes' component and make it trigger the refresh when needed.

bodnara (approved):

bodnara:
Please rebase this PR.

PR #734, opened 2021-06-17
Fixed texts for instances withotu cs localisation

xkostka2:
* Application forms and notifications now does not show option to add cs texts to them
* Localisation config added to all GUI applications (even if it is not used there yet)


bodnara (changes requested):
Remove to option to switch language in application form preview if only one language is supported.
image

bodnara (approved):

PR #741, opened 2021-06-22
Added option for change member status

HejdaJakub:
* User can change member status in VO and code for changing group status is prepared.
* In change status dialog user gets different labels based on current status and new status.
* In some use cases user should edit expiration, so after status change is open second dialog with expiration settings.
* In expiration dialog is new option which allows user with VALID status to set expiration according to Vo/Group expiration rules (if it is possible).
* Min and max possible expiration date differs according to current member status.


apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1355,12 +1355,24 @@
...
       "TITLE": "Change status",
-      "CANCEL": "Cancel",
-      "CURRENT_STATUS": "Current status",
-      "NEW_STATUS": "New status",
-      "CHANGE_VALID": "Changing status to VALID will trigger automatic configuration for provided resources. \nIf successful, member will have access on provided resources. \nIf not, see displayed error message and do manual configuration on 'settings' tab on members detail.",
+      "CURRENT_STATUS": "Current status: ",
+      "VALID_DESCRIPTION": "Member is properly configured and have access on provided resources.",
       "INVALID_DESCRIPTION": "Member have configuration error and DON'T have access on provided resources. You can check what is wrong by changing member's status to VALID. If possible, procedure will configure all necessary settings by itself.",
-      "SUCCESS": "Member status was changed"
+      "EXPIRED_DESCRIPTION": "Member didn't extend membership and DON'T have access on provided resources.",
```
Vojtech-Sassmann, line 1361: This might be a bit confusing but some services allow also expired members. This solely depends on service implementation. Some description like this might be more accurrate.
```suggestion
      "EXPIRED_DESCRIPTION": "Member didn't extend membership and might NOT have access to all services.",
```

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1355,12 +1355,24 @@
     },
     "CHANGE_STATUS": {
       "TITLE": "Change status",
-      "CANCEL": "Cancel",
-      "CURRENT_STATUS": "Current status",
-      "NEW_STATUS": "New status",
-      "CHANGE_VALID": "Changing status to VALID will trigger automatic configuration for provided resources. \nIf successful, member will have access on provided resources. \nIf not, see displayed error message and do manual configuration on 'settings' tab on members detail.",
+      "CURRENT_STATUS": "Current status: ",
+      "VALID_DESCRIPTION": "Member is properly configured and have access on provided resources.",
```
Vojtech-Sassmann, line 1359:
```suggestion
      "VALID_DESCRIPTION": "Member is properly configured and has access to all provided resources.",
```

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1355,12 +1355,24 @@
...
       "INVALID_DESCRIPTION": "Member have configuration error and DON'T have access on provided resources. You can check what is wrong by changing member's status to VALID. If possible, procedure will configure all necessary settings by itself.",
-      "SUCCESS": "Member status was changed"
+      "EXPIRED_DESCRIPTION": "Member didn't extend membership and DON'T have access on provided resources.",
+      "DISABLED_DESCRIPTION": "Member didn't extend membership long time ago or was manually disabled and DON'T have access on provided resources.",
```
Vojtech-Sassmann, line 1362:
```suggestion
      "DISABLED_DESCRIPTION": "Member was manually disabled and DON'T have access on provided resources.",
```

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1355,12 +1355,24 @@
...
+      "NEW_STATUS": "New status: ",
+      "SELECT_STATUS": "Select status",
+      "CHANGE_VALID_TO_INVALID": "Changing status to INVALID will <b>prevent member from access to provided resources (based on provided service's rules).</b>\n\nINVALID status means there is configuration error, which prevents him from access on provided resources.",
```
Vojtech-Sassmann, line 1365:
```suggestion
      "CHANGE_VALID_TO_INVALID": "Changing status to INVALID will <b>prevent the member from accessing all services.</b>\n\nINVALID members might not have correct configuration (required by services) and has to be validated before accessing them.",
```

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1355,12 +1355,24 @@
...
+      "CHANGE_VALID_TO_EXPIRED": "Changing status to EXPIRED will <b>prevent member from access to provided resources (based on provided service's rules).</b>\n\nEXPIRED status means, that member didn't extend his membership in VO, but it's still possible for him to do so.",
```
Vojtech-Sassmann, line 1366:
```suggestion
      "CHANGE_VALID_TO_EXPIRED": "Changing status to EXPIRED will <b>might prevent the member from accessing some of the services (based on provided service's rules).</b>\n\nEXPIRED status means, that member didn't extend his membership in VO, but it's still possible for him to do so.",
```

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1355,12 +1355,24 @@
...
+      "CHANGE_VALID_TO_DISABLED": "Changing status to DISABLED will <b>prevent member from access to provided resources (based on provided service's rules).</b>\n\nDISABLED status means, that member didn't extend his membership long ago or was manually disabled by administrator. Member can't enable/extend membership by himself.",
```
Vojtech-Sassmann, line 1367:
```suggestion
      "CHANGE_VALID_TO_DISABLED": "Changing status to DISABLED will <b>prevent the member from accessing all of the services. </b>\n\nDISABLED status means that the member is archived and is not allowed to any services.",
```

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1355,12 +1355,24 @@
...
+      "CHANGE_NO_VALID_TO_VALID": "Changing status to VALID <b>will trigger automatic configuration</b> for provided resources. \n\n<b>If successful</b>, member will have access on provided resources. \n\n<b>If not</b>, see displayed error message and do manual configuration on 'settings' tab on members detail.",
```
Vojtech-Sassmann, line 1368:
```suggestion
      "CHANGE_NO_VALID_TO_VALID": "Changing status to VALID <b>will trigger attribute validation</b>. \n\n<b>If successful</b>, member will have access on all services. \n\n<b>If not</b>, see displayed error message and do manual configuration of invalid attributes.,
```

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1355,12 +1355,24 @@
...
+      "CHANGE_NO_VALID_TO_INVALID": "INVALID status means there is configuration error, which prevents him from access on provided resources.",
```
Vojtech-Sassmann, line 1369:
```suggestion
      "CHANGE_NO_VALID_TO_INVALID": "INVALID status means there is a configuration error, which prevents the member from accessing provided services.",
```

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1355,12 +1355,24 @@
...
+      "CHANGE_NO_VALID_TO_EXPIRED": "EXPIRED status means, that member didn't extend his membership in VO, but it's still possible for him to do so.",
+      "CHANGE_NO_VALID_TO_DISABLED": "DISABLED status means, that member didn't extend his membership long ago or was manually disabled by administrator. Member can't enable/extend membership by himself.",
```
Vojtech-Sassmann, line 1371:
```suggestion
      "CHANGE_NO_VALID_TO_DISABLED": "DISABLED status means that the member is archived and is not allowed to any services. Member can't enable/extend membership by himself.",
```

apps/admin-gui/src/app/vos/pages/member-detail-page/member-overview/member-overview-membership/member-overview-membership.component.ts
```diff
@@ -39,23 +39,51 @@ export class MemberOverviewMembershipComponent implements OnChanges {
...
+    const oldStatus = this.member.status;
+
+    const dialogRef = this.dialog.open(ChangeMemberStatusDialogComponent, config);
+    dialogRef.afterClosed().subscribe( member => {
+      if (member) {
+        this.member = member;
+        if ((oldStatus === 'VALID' && member.status === 'EXPIRED') || (oldStatus === 'VALID' && member.status === 'DISABLED') || member.status === 'VALID') {
+          this.memberStatusChanged = true;
+          this.statusNotChangedToValid = member.status !== 'VALID';
+          this.changeVoExpiration();
```
Vojtech-Sassmann, line 63: Do we always want to change the expiration? I think that when you set the status to 'INVALID' or 'DISABLED' this should not appear.

Vojtech-Sassmann (changes requested):
I'm not sure if it makes sense to create a single component for handling the status changes for vo and group. The code just seems more complicated because of all of the if/else branches.  Wouldn't it be better to create two separate components (and maybe one with shared logic)?

apps/admin-gui/src/assets/i18n/en.json
HejdaJakub: All wrong texts were fixed.

PR #744, opened 2021-06-28
feat(admin-gui): Show urn on attr display names

Vojtech-Sassmann:
* In attribute lists we need to show the urn so the users can identify attributes.

### DIFF
--- a/libs/perun/components/src/lib/attributes-list/attributes-list.component.html
+++ b/libs/perun/components/src/lib/attributes-list/attributes-list.component.html
@@ -43,7 +43,14 @@
       <ng-container matColumnDef="displayName">
         <th *matHeaderCellDef mat-header-cell
             mat-sort-header>{{'SHARED_LIB.PERUN.COMPONENTS.ATTRIBUTES_LIST.NAME' | translate}}</th>
-        <td mat-cell *matCellDef="let attribute">{{attribute.displayName}}</td>
+        <td mat-cell
+            *matCellDef="let attribute">
+          <div
+            matTooltipClass='urn-tooltip'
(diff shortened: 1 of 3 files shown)


libs/perun/components/src/lib/attributes-list/attributes-list.component.html
```diff
@@ -43,7 +43,14 @@
...
         <th *matHeaderCellDef mat-header-cell
             mat-sort-header>{{'SHARED_LIB.PERUN.COMPONENTS.ATTRIBUTES_LIST.NAME' | translate}}</th>
-        <td mat-cell *matCellDef="let attribute">{{attribute.displayName}}</td>
+        <td mat-cell
+            *matCellDef="let attribute">
+          <span
+            matTooltipClass='urn-tooltip'
+            matTooltip='{{getAttributeFullName(attribute)}}'>
+            {{attribute.displayName}}
+          </span>
```
HejdaJakub, line 52: If you have some multi line `displayName`, then the tooltip hides between every single line on hovering. Maybe a `div` instead of `span` could solve it? But it's only a small suggestion, which is not necesary.

HejdaJakub (approved):

libs/perun/components/src/lib/attributes-list/attributes-list.component.html
Vojtech-Sassmann: I'm not sure, what do you mean. Could you please provide me a screenshot?

Vojtech-Sassmann: I have replaced the `span` element with `div` and it fixed the problem.

HejdaJakub (approved):

PR #758, opened 2021-07-02
Refresh member list

bodnara:
* added refresh button to vo and group members page
* table content is refresh with each change to filter value


xkureck (changes requested):
I dont like the search field - how you implemented with every change of letter new request to backend is made - that is really scary and can be really not optiomal
I would reuse debounce-filter component that we already have in our GUI, it has debounce time where it waits some time to actually search in something

xkureck (approved):

PR #767, opened 2021-07-08
Sorting items in dropdown-selects at member attributes page

xkostka2:


apps/admin-gui/src/app/shared/components/two-entity-attribute-page/two-entity-attribute-page.component.ts
```diff
@@ -62,21 +61,32 @@ export class TwoEntityAttributePageComponent implements OnInit {
...
+    return a.name.toLowerCase() > b.name.toLowerCase() ? 1 : (a.name.toLowerCase() === b.name.toLowerCase() ? 0 : -1)
+  }
+
+  compareFnUser(a,b) {
+    if(a.user){
+      return parseName(a.user) > parseName(b.user) ? 1 : (parseName(a.user) === parseName(b.user) ? 0 : -1);
+    }
+    return parseName(a) > parseName(b) ? 1 : (parseName(a) === parseName(b) ? 0 : -1);
+  }
+
```
xkureck, line 74: Maybe I am wrong but wouldn't it be better if the sorting would happen in the group-search-select.component/vo-search-select.component/... so it would be universal and we would have it sorted everywhere where we are using the selectors?
I think that we don't have use-case where sorted items in the selector wouldn't make sense. So it would probably make sense.

apps/admin-gui/src/app/shared/components/two-entity-attribute-page/two-entity-attribute-page.component.ts
```diff
@@ -86,15 +96,15 @@ export class TwoEntityAttributePageComponent implements OnInit {
...
+              this.entityValues = resources.sort(this.compareFnName);
               this.preselectEntity();
               this.loading = false;
             });
             break;
           case 'member':
             // return one attribute because if an empty list is passed, all attributes are returned
             this.membersManager.getCompleteRichMembersForGroup(this.firstEntityId, false, this.allowedStatuses, null, [Urns.MEMBER_CORE_ID]).subscribe(members => {
-              this.entityValues = members;
+              this.entityValues = members.sort(this.compareFnUser);
```
xkureck, line 107: when we are sorting the names(first name + last name) of members and users I think we really want to sort according to last name(then according to firstname) + count with czech alphabet

apps/admin-gui/src/app/shared/components/two-entity-attribute-page/two-entity-attribute-page.component.ts
```diff
@@ -112,14 +122,14 @@ export class TwoEntityAttributePageComponent implements OnInit {
         switch (this.secondEntity){
           case 'member':
             this.resourcesManagerService.getAssignedRichMembers(this.firstEntityId).subscribe(members => {
-              this.entityValues = members;
+              this.entityValues = members.sort(this.compareFnUser);
```
xkureck, line 125: when we are sorting the names(first name + last name) of members and users I think we really want to sort according to last name(then according to firstname) + count with czech alphabet

apps/admin-gui/src/app/shared/components/two-entity-attribute-page/two-entity-attribute-page.component.ts
```diff
@@ -128,7 +138,7 @@ export class TwoEntityAttributePageComponent implements OnInit {
         break;
       case 'facility':
         this.facilitiesManagerService.getAssignedUsers(this.firstEntityId).subscribe(users => {
-          this.entityValues = users;
+          this.entityValues = users.sort(this.compareFnUser);
```
xkureck, line 141: when we are sorting the names(first name + last name) of members and users I think we really want to sort according to last name(then according to firstname) + count with czech alphabet

xkureck (changes requested):

xkureck (approved):

PR #770, opened 2021-07-08
refactor: Remove unused apps

Vojtech-Sassmann:
* The consolidator and statistics are not actually used and can be removed.


xkureck (changes requested):
in .gitignore file remove the `apps/statistics/src/assets/config/instanceConfig.json` line and I think its fine

xkureck (approved):

PR #771, opened 2021-07-09
fix(admin): Fix user search input width

Vojtech-Sassmann:
* The input for user search was in a label element, which was wrong.


xkureck (approved):
cool

PR #772, opened 2021-07-13
fix(admin): show resource tags for group's resources

Johaney-s:
* Resource tags were missing in this table


libs/perun/openapi/src/lib/api.module.ts
```diff
@@ -49,7 +49,7 @@ import { VosManagerService } from './api/vosManager.service';
     VosManagerService ]
 })
 export class ApiModule {
-    public static forRoot(configurationFactory: () => Configuration): ModuleWithProviders<ApiModule> {
+    public static forRoot(configurationFactory: () => Configuration): ModuleWithProviders {
```
Vojtech-Sassmann, line 52: Please, remove this change. I dunno why but this always happens after openapi generation and it is needed to specify the ApiModule generic type.

Vojtech-Sassmann (changes requested):

libs/perun/openapi/src/lib/api.module.ts
Johaney-s: I see. Removed.

Vojtech-Sassmann (approved):

PR #774, opened 2021-07-15
fix(admin): Fix order of columns in groups list

cuadradek:
- The group-resource status in table in group-list component was placed
after description column. Now, it is placed after name column like in
other tables, e.g. in members-list.
- Each status icon has now different color.

### DIFF
--- a/libs/perun/components/src/lib/group-resource-status/group-resource-status.component.html
+++ b/libs/perun/components/src/lib/group-resource-status/group-resource-status.component.html
@@ -2,23 +2,23 @@
   *ngIf="status === 'ACTIVE'"
   matTooltip="{{'SHARED_LIB.PERUN.COMPONENTS.GROUP_RESOURCE_STATUS.ACTIVE' | translate}}"
   matTooltipPosition="above">
-  <mat-icon>check_circle_outline</mat-icon>
+  <mat-icon class ="green">check_circle_outline</mat-icon>
 </span>
 <span
   *ngIf="status === 'INACTIVE'"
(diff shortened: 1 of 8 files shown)


HejdaJakub (approved):

libs/perun/components/src/lib/group-resource-status/group-resource-status.component.html
```diff
@@ -2,23 +2,23 @@
   *ngIf="status === 'ACTIVE'"
   matTooltip="{{'SHARED_LIB.PERUN.COMPONENTS.GROUP_RESOURCE_STATUS.ACTIVE' | translate}}"
   matTooltipPosition="above">
-  <mat-icon>check_circle_outline</mat-icon>
+  <mat-icon [ngStyle]="{'color':'green'}">check_circle_outline</mat-icon>
```
xkureck, line 5:
```suggestion
  <mat-icon class="green">check_circle_outline</mat-icon>
```
if you could use that sort of style it would be nice, it also moves the icon to the middle, the styles are defined in the `styles.scss` and you could also extend it to add colours that you need(black, grey)
image

xkureck (changes requested):
I think we would like to also move the column with status also here 
image

libs/perun/components/src/lib/group-resource-status/group-resource-status.component.html
cuadradek: Done.

xkureck (approved):

PR #775, opened 2021-07-17
feat(admin): add perunobserver

metodej:
* added role perunobserver, this role should be able to read everything perunadmin can


HejdaJakub (approved):

libs/perun/services/src/lib/gui-auth-resolver.service.ts
```diff
@@ -199,19 +204,23 @@ export class GuiAuthResolver {
   }
 
   public canManageFacilities(): boolean {
-    return this.hasAtLeasOne(Role.PERUNADMIN, Role.FACILITYADMIN, Role.FACILITIYOBSERVER);
+    return this.hasAtLeasOne(Role.PERUNADMIN, Role.PERUNOBSERVER, Role.FACILITYADMIN, Role.FACILITIYOBSERVER);
```
xkureck, line 207: with this change there could be probably some change in the name of the method

xkureck (changes requested):
changes in this PR should be OK, but imcomplete
in the PERUN ADMIN section there is no restriction for perun observer for changing/deleting action - this can be done by some gui member in other PR if you dont feel like doing it

apps/admin-gui/src/app/admin/pages/admin-page/admin-services/admin-services.component.html
```diff
@@ -2,6 +2,7 @@
   <h1 class="page-subtitle">{{'ADMIN.SERVICES.TITLE' | translate}}</h1>
   <perun-web-apps-refresh-button (refresh)="refreshTable()"></perun-web-apps-refresh-button>
   <button
+    *ngIf="authResolver.isPerunAdmin()"
```
xkureck, line 5: on this page also put this if on the delete button please 
image

xkureck (changes requested):
on component user-identities please add also check in rights to add and remove button
on component user-settings-service-identities please also add check in right to add/remove button

apps/admin-gui/src/app/admin/pages/admin-page/admin-services/admin-services.component.html
metodej: Fixed, thanks.

xkureck (approved):

PR #780, opened 2021-07-20
feat(admin): add resource tags page

HejdaJakub:
* To the resource section has been added page with assigned resource tags.
* Resource tags can be added or removed from the resource tag list.


apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-tags/resource-tags.component.ts
```diff
@@ -0,0 +1,140 @@
...
+    // TODO: edit auth objects
+    this.addAuth = this.authResolver.isAuthorized('assignResourceTagToResource_ResourceTag_Resource_policy', [this.resource]);
+    this.removeAuth = this.authResolver.isAuthorized('removeResourceTagFromResource_ResourceTag_Resource_policy', [this.resource]);
+    this.editAuth = this.authResolver.isAuthorized('updateResourceTag_ResourceTag_policy', [this.resource]);
+
+    this.displayedColumns = this.removeAuth ? ['select', 'id', 'name'] : ['id', 'name'];
+
+    if(this.editAuth){
+      this.displayedColumns.push('edit');
+    }
```
xkureck, line 129: About edit column in assigned tags on resource: I dont think that it harms anything if its there but I found it personally counfusing honestly. Functionally this page is for adding/removing tags from resource and editing the actual tag name here is.. not what I would expect and I think the casual user can be confused. 
I would personally dont put this column here

xkureck (changes requested):

apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-tags/resource-tags.component.ts
HejdaJakub: Edit column has been removed.

xkureck (approved):

PR #784, opened 2021-07-23
feat(admin): add link to user profile app

HejdaJakub:
* In the nav menu was replaced link to My profile by link to user profile app.
* Label 'My profile' was replaced by label 'My account'.


apps/admin-gui/src/app/shared/perun-nav/perun-nav.component.ts
```diff
@@ -49,6 +52,13 @@ export class PerunNavComponent implements OnInit, AfterViewInit {
     this.isDevel = this.storeService.get('isDevel');
     this.logo = this.sanitizer.bypassSecurityTrustHtml(this.store.get('logo'));
     this.logoutEnabled = this.storeService.get('log_out_enabled');
+    this.utilsService.getAppsConfig().subscribe(config => {
+      this.profileUrl = !!config.brands[0].newApps.profile ? config.brands[0].newApps.profile : null;
+    });
+  }
+
+  redirectToProfile() {
+    window.location.href = this.profileUrl;
```
Vojtech-Sassmann, line 61: I think we should open other applications in a new tab, what do you think?
```suggestion
    window.open(this.profileUrl, '_blank');
```

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -30,7 +30,7 @@
     "FACILITIES_DESC": "Manage facilities and resources.",
     "ADMIN_DESC": "Manage attributes, users and do diagnostics.",
     "PROFILE_DESC": "View information about you and your organizations and groups.",
-    "MY_PROFILE": "My profile"
+    "MY_ACCOUNT": "My account"
```
Vojtech-Sassmann, line 33: I'm not sure if we want to rename this page to "My account". I get that you wanted to avoid duplicated profiles but we might want to remove the old profile completely. I would suggest keeping the labels the way they were.

apps/admin-gui/src/app/shared/perun-nav/perun-nav.component.html
```diff
@@ -52,7 +52,10 @@
     </span>
   </button>
   <mat-menu #menu="matMenu">
-    <button [routerLink]="['/myProfile']" mat-menu-item>{{'NAV.PROFILE' | translate}}</button>
+    <button *ngIf="profileUrl !== null" (click)="redirectToProfile()" mat-menu-item>
```
Vojtech-Sassmann, line 55: If the profileUrl is not specified, there is an empty string. This way, the button will not show even if the string is empty.
```suggestion
    <button *ngIf="profileUrl" (click)="redirectToProfile()" mat-menu-item>
```

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann (approved):
I think this is ok, however, we should discuss it tomorrow with the rest of the team.

xkureck (commented):
image
and what about like that? (both in user profile and admin gui

apps/admin-gui/src/app/shared/perun-nav/perun-nav.component.scss
```diff
@@ -35,3 +35,8 @@
     display: block;
   }
 }
+
+.icon-without-margin {
+  margin: 0 0 0 0 !important;
+  color: black;
+}
```
Vojtech-Sassmann, line 42: I think this is not necessary anymore.

Vojtech-Sassmann (approved):
Just a minor problem. Otherwise, it is fine by me.

apps/admin-gui/src/app/shared/perun-nav/perun-nav.component.scss
HejdaJakub: It's true, I removed this css class.

HejdaJakub:
> image
> and what about like that? (both in user profile and admin gui

I switched the items in menu in both applicatoins.

xkureck (changes requested):
so on phone its kind of weird 
image
the changes were probably made by the other PR and the problem is also on the user profile app

HejdaJakub:
> so on phone its kind of weird
> image
> the changes were probably made by the other PR and the problem is also on the user profile app

Mobile view fixed in admin gui.

xkureck (approved):

PR #785, opened 2021-07-23
feat(profile): add link to admin gui app

HejdaJakub:
* To the header was added link to admin gui, but this link is visible only for users with defined
roles in config file.


apps/user-profile/src/app/components/header/header.component.html
```diff
@@ -7,6 +7,13 @@
   <a class="ml-3" [innerHTML]="logo" [routerLink]="disableLogo ? [] : ['/']" queryParamsHandling="merge"></a>
 
   <p [ngStyle]="{'color': textColor}" class="ml-3 mt-auto mb-auto">{{label}}</p>
+  <div *ngIf="activeLink && adminGuiUrl !== null" id="nav-menu-user-info" class="justify-content-end">
```
Vojtech-Sassmann, line 10:
```suggestion
  <div *ngIf="activeLink && adminGuiUrl" id="nav-menu-user-info" class="justify-content-end">
```

apps/user-profile/src/app/components/header/header.component.html
```diff
@@ -7,6 +7,13 @@
...
+    <button
+      mat-button
+      class="ml-3 mt-auto mb-auto"
+      (click)="redirectToAdminGUI()">
+      <mat-icon [ngStyle]="{'color': iconColor}">forward</mat-icon>
+      <span [ngStyle]="{'color': textColor}">Admin GUI</span>
+    </button>
```
Vojtech-Sassmann, line 17: In the future, we will have multiple applications, where users might want to navigate (consolidator, registrar). I would implement the nav the same way you did for the admin gui.

apps/user-profile/src/app/components/header/header.component.ts
```diff
@@ -27,14 +28,38 @@ export class HeaderComponent implements OnInit {
...
+  linkRoles: string[];
+  activeLink = false;
 
   ngOnInit() {
     this.translate.onLangChange.subscribe(lang => {
       this.label = this.storeService.get(`header_label_${lang.lang}`)
     })
     this.logo = this.sanitizer.bypassSecurityTrustHtml(this.storeService.get('logo'));
+
+    this.authzResolverService.getPerunPrincipal().subscribe(principal => {
```
Vojtech-Sassmann, line 45: You can get synchronously the current principal from the `storeService` component.

apps/user-profile/src/app/components/header/header.component.ts
```diff
@@ -27,14 +28,38 @@ export class HeaderComponent implements OnInit {
...
+      if (this.activeLink) {
+        this.utilsService.getAppsConfig().subscribe(config => {
+          this.adminGuiUrl = !!config.brands[0].newApps.admin ? config.brands[0].newApps.admin : null;
+        });
+      }
+    });
+  }
+
+  redirectToAdminGUI() {
+    window.location.href = this.adminGuiUrl;
```
Vojtech-Sassmann, line 62:
```suggestion
    window.open(this.adminGuiUrl, '_blank');
```

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann (changes requested):
There seems to be a problem with the nav, the menu appears behind it. The fix should be easy, just set the `z-index` to 999 instead of 1020.

Vojtech-Sassmann (approved):
Code seems ok, however, we should discuss it with the rest of the team. Also, it might be worth unifying the header design of both applications. E.g. we could consider to show only the applications icon, next to the name, which would show overview of our applications.

apps/user-profile/src/app/components/header/header.component.html
```diff
@@ -7,6 +7,39 @@
...
+        apps
+      </mat-icon>
+    </button>
+    <mat-menu #appsMenu="matMenu">
+      <button
+        mat-menu-item
+        (click)="redirectToAdminGUI()">
+        <span class="black">{{adminLabel}}</span>
+      </button>
+    </mat-menu>
```
Vojtech-Sassmann, line 43: Would it be possible to move this component (or some part of it) to the shared libraries? Since we use it in two applications.

apps/user-profile/src/styles.scss
```diff
@@ -118,3 +118,8 @@ th, td {
 .app-min-width {
   min-width: 400px;
 }
+
+.mat-tooltip {
+  font-size: 14px !important;
+  word-wrap: break-word !important;
+}
```
Vojtech-Sassmann, line 125: Be careful with this. This way, you will affect all of the tooltips in the whole application. I don't know if the font-size change is necessary. About the word-wrap, I would maybe try to limit this change only for a certain tooltip.

apps/user-profile/src/assets/config/defaultConfig.json
```diff
@@ -113,6 +113,22 @@
   ],
   "header_label_en": "Generic user profile",
   "header_label_cs": "Generický profil uživatele",
+  "log_out_enabled": true,
+  "admin_gui_label_en": "Admin GUI",
+  "admin_gui_label_cs": "Admin GUI",
```
Vojtech-Sassmann, line 118: Not sure if something like `IDM Administration` or `Administration` wouldn't be more clear to users? You can try to think of something on your own. I just don't think the `Admin GUI` is very user-friendly.

Vojtech-Sassmann (changes requested):

apps/user-profile/src/app/components/header/header.component.html
HejdaJakub: This block of code has been moved to the separated shared component.

apps/user-profile/src/styles.scss
HejdaJakub: We have the same "global" class in admin-gui, so I used it also in user-profile.

apps/user-profile/src/assets/config/defaultConfig.json
HejdaJakub: I changed labels to `IAM Administration`/`Správa IAM`, but I'm not still sure, if it is the best solution..

apps/user-profile/src/app/components/header/header.component.ts
```diff
@@ -27,14 +31,42 @@ export class HeaderComponent implements OnInit {
...
+      });
+    }
+  }
+
+  onLogOut() {
+    this.authService.logout();
+  }
+
+  redirectToAdminGUI() {
+    window.open(this.adminGuiUrl, '_blank');
```
xkureck, line 69: this can be removed

apps/user-profile/src/app/components/header/header.component.html
```diff
@@ -7,6 +7,16 @@
...
+    <perun-web-apps-header-menu
+      [activeLink]="activeLink"
+      [url]="adminGuiUrl"
+      [label]="adminLabel"
+      [user]="principal.user"
+      [logoutEnabled]="logoutEnabled"
+      [iconColor]="iconColor"
+      [textColor]="textColor">
+    </perun-web-apps-header-menu>
+  </div>
```
xkureck, line 21:
```suggestion
  <perun-web-apps-header-menu
    id="nav-menu-user-info"
    class="justify-content-end"
    [activeLink]="activeLink"
    [url]="adminGuiUrl"
    [label]="adminLabel"
    [user]="principal.user"
    [logoutEnabled]="logoutEnabled"
    [iconColor]="iconColor"
    [textColor]="textColor">
  </perun-web-apps-header-menu>
```

apps/user-profile/src/assets/config/defaultConfig.json
```diff
@@ -113,6 +113,22 @@
...
+    "VOADMIN",
+    "VOOBSERVER",
+    "GROUPADMIN",
+    "GROUPOBSERVER",
+    "RESOURCEADMIN",
+    "RESOURCEOBSERVER",
+    "FACILITYADMIN",
+    "FACILITYOBSERVER",
+    "TRUSTEDFACILITYADMIN"
+  ],
```
xkureck, line 131: missing role SPONSOR and TOGROUPCREATOR

apps/user-profile/src/app/components/header/header.component.ts
```diff
@@ -27,14 +31,42 @@ export class HeaderComponent implements OnInit {
...
+      if (this.linkRoles.includes(roleKey)) {
+        this.activeLink = true;
+      }
+    }
+
+    if (this.activeLink) {
+      this.utilsService.getAppsConfig().subscribe(config => {
+        this.adminGuiUrl = !!config.brands[0].newApps.admin ? config.brands[0].newApps.admin : null;
+      });
+    }
```
xkureck, line 61: move this code to separate function

apps/user-profile/src/app/components/header/header.component.html
```diff
@@ -7,6 +7,16 @@
   <a class="ml-3" [innerHTML]="logo" [routerLink]="disableLogo ? [] : ['/']" queryParamsHandling="merge"></a>
 
   <p [ngStyle]="{'color': textColor}" class="ml-3 mt-auto mb-auto">{{label}}</p>
-</mat-toolbar>
-
 
+  <div id="nav-menu-user-info" class="justify-content-end">
+    <perun-web-apps-header-menu
+      [activeLink]="activeLink"
```
xkureck, line 13: do we need this property to be sent into the header menu? isnt it redundant?

xkureck (changes requested):

apps/user-profile/src/app/components/header/header.component.ts
HejdaJakub: Removed.

apps/user-profile/src/app/components/header/header.component.html
HejdaJakub: Div element has been removed and `id` and `class` has been moved to the element `perun-web-apps-header-menu`.

apps/user-profile/src/assets/config/defaultConfig.json
HejdaJakub: This roles have been added.

apps/user-profile/src/app/components/header/header.component.ts
HejdaJakub: This block of code has been moved to the separate function.

apps/user-profile/src/app/components/header/header.component.html
HejdaJakub: It's true. I removed this property from this html element and also from the header-menu comomponent.

xkureck (approved):

PR #788, opened 2021-07-26
feat(admin): change status of group-resource assigment

bodnara:
* group-resource assignment can be activated or deactivated from both group and resource lists

### DIFF
--- a/libs/perun/dialogs/src/lib/change-group-resource-assigment-dialog/change-group-resource-assigment-dialog.component.html
+++ b/libs/perun/dialogs/src/lib/change-group-resource-assigment-dialog/change-group-resource-assigment-dialog.component.html
@@ -1,0 +1,60 @@
+<div class="{{theme}}">
+  <h1 mat-dialog-title>{{'SHARED_LIB.PERUN.COMPONENTS.CHANGE_GROUP_RESOURCE_ASSIGMENT_DIALOG.TITLE' | translate}}</h1>
+  <mat-spinner *ngIf="loading" class="ml-auto mr-auto"></mat-spinner>
+  <div mat-dialog-content *ngIf="!loading">
+    <p class="mb-4">
+      {{group.name}} <span class="text-muted id-font">#{{group.id}}</span> - {{resource.name}} <span class="text-muted id-font">#{{resource.id}}</span>
+    </p>
+    <p>
+      <strong class="mr-2">{{'SHARED_LIB.PERUN.COMPONENTS.CHANGE_GROUP_RESOURCE_ASSIGMENT_DIALOG.CURRENT' | translate}}:</strong>
+      {{status}}
+    </p>
+
+    <p>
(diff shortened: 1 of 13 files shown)


HejdaJakub (changes requested):
The code seems to be ok, I have only one thing. If I want to change status from `INACTIVE` to `ACTIVE`, there is no default sync/async option selected and the submit button is not disabled, so I can submit the dialog without any selected option. I would prefer to choose one of them as the default selected option.
image

xkureck (changes requested):
image
I want some description of what would happen if I change the status

image
Nothing is selected but I can still submit the dialog, if something would be preselected that would be great. 
image
The description of what is asynchronous/synchronous assigment is great but I still dont have any idea what that means.

HejdaJakub (commented):
Loading spinner would look better in the middle of the dialog.
image

libs/perun/dialogs/src/lib/change-group-resource-assigment-dialog/change-group-resource-assigment-dialog.component.html
```diff
@@ -0,0 +1,56 @@
+<h1 mat-dialog-title>{{'SHARED_LIB.PERUN.COMPONENTS.CHANGE_GROUP_RESOURCE_ASSIGMENT_DIALOG.TITLE' | translate}}</h1>
+<mat-spinner *ngIf="loading"></mat-spinner>
+<div mat-dialog-content *ngIf="!loading">
+  <h5 class="mb-2">
```
xkureck, line 4:
```suggestion
  <p class="mb-4">
```

xkureck (changes requested):

xkureck:
> Loading spinner would look better in the middle of the dialog.
> image

agree with jakub, also some appropriate colour would be nice

libs/perun/dialogs/src/lib/change-group-resource-assigment-dialog/change-group-resource-assigment-dialog.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    }, () => this.loading = false);
+  }
+
+  onCancel() {
+    this.dialogRef.close(false);
+  }
+
+  onSubmit() {
+    this.loading = true;
+    if(this.status==="ACTIVE"){
```
xkureck, line 54:
```suggestion
    if(this.status === "ACTIVE") {
```

libs/perun/dialogs/src/lib/change-group-resource-assigment-dialog/change-group-resource-assigment-dialog.component.html
```diff
@@ -0,0 +1,63 @@
+<div class="{{theme}}">
```
xkureck, line 1: please check the format of this html, add spaces where it should be, remove them where they shouldnt be

xkureck (approved):
just repair this little code readibility upgrades and then merge it

HejdaJakub (approved):

PR #792, opened 2021-07-28
build: enabling dev mode

xkostka2:
* angular.json has been edited to support building the apps in dev mode by default

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -69,6 +69,26 @@
                   "maximumWarning": "6kb"
                 }
               ]
+            },
+            "development": {
+              "optimization": false,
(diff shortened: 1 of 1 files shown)


Vojtech-Sassmann (changes requested):
This is just a comment about the commit message:
* If you don't want to specify a scope, you should user this syntax instead:
```
fix: enabling dev mode
```
* This is not a fix. `Fix` is a change to the application logic, this is only related to how we test our applications. This should be either a `build` or a `chore`.

xkureck (approved):
GJ

PR #793, opened 2021-07-29
feat: add support for creating service members

xkostka2:
* Added page into VO detail, where we can view all the service members in current VO and also create new ones
* Added dialog for creating service members
* Actual OpenAPI has been generated


Vojtech-Sassmann:
@xkostka2 Again, don't use the `perun-web-apps` scope, see: https://github.com/CESNET/perun-web-apps/pull/792#pullrequestreview-717769187

xkureck (changes requested):
i am missing remove button(for removing the service member from VO) - you can use for that universal remove dialog I think
I think that also michal wanted it on the basic overview not on the advanced settings
image
I think that we have in our library some image called service_identity_<color> - that would be probably better than icon for group

image
When I change the namespace some warning appears

image
When I change the namespace back to not selected some error appears, which is weird.

image
The position of the buttons "next" "create" and "cancel"  is super weird. please inspire in our "create sponsored member" dialogs. Btw when I click on the next button nothing happens(when no namespace choosed).

image
I want some error message why the GUI dont like that email.

image
Search filed could be bigger and could react to enter key

image
Instead of the trash can I would like here more the "cross"
You could add the id of the user here also
Also when there is only one name I would just disable the icon for remove and add tooltip why is it disabled.

image
Add gap between these two elements 

image
I feel like the gap between these two elements is too big

image
When no namespace is selected its really weird that the third "step" in stepper is missing and when choosing some other namespace it suddenly out of nowhere appears. The number of steps should be constant no matter what. On the last step there could be just some text that tells the user he dont have to set the password

image
I would like here if the no namespace is selected the login field would be just disabled(with tooltip why it is disabled) not completely hided. Again I found it weird when somehow suddenly out of nowhere appears some input field

xkureck (changes requested):

xkureck (changes requested):
image
The password which is optional I would still have something else. I would inspire in our create sponsored member dialog 
image
where there is checkbox and if its checked the password input is disabled. The text next to the label could be something like: "Generate random password"

image
Also use green button for "next" as it is in create sponsored member dialog

xkureck (changes requested):
image
just when i open the dialog some error appears

image
I would add the second option this way

image
When login is checked you could add some debounce time to the input field so the gui dont have to do so much calls to backend

apps/admin-gui/src/app/shared/components/create-service-member-dialog/create-service-member-dialog.component.html
```diff
@@ -0,0 +1,203 @@
...
+      </mat-step>
+      <mat-step>
+        <ng-template matStepLabel>{{'DIALOGS.CREATE_SERVICE_MEMBER.ASSOCIATE_USERS' | translate}}</ng-template>
+        <h6>{{'DIALOGS.CREATE_SERVICE_MEMBER.ASSOCIATED_USERS' | translate}}</h6>
+        <mat-list>
+          <mat-list-item *ngFor="let member of assignedMembers">
+            <p class="truncate center-content">{{member.user | userFullName}} <span *ngIf="member.id" class="text-muted">#{{member.id}}</span>
+              <span [matTooltipDisabled]="assignedMembers.length > 1" [matTooltip]="'DIALOGS.CREATE_SERVICE_MEMBER.MINIMAL_USERS' | translate">
+                <button (click)="removeUser(member)" [disabled]="assignedMembers.length === 1" class="ml-1" color="warn" mat-icon-button>
+                  <mat-icon style="margin-bottom: 6px;">close</mat-icon>
```
xkureck, line 138: no css style in the html please

apps/admin-gui/src/app/shared/components/create-service-member-dialog/create-service-member-dialog.component.html
```diff
@@ -0,0 +1,203 @@
...
+        <mat-spinner *ngIf="loading" class="ml-auto mr-auto">
+        </mat-spinner>
+        <div *ngIf="!loading">
+          <perun-web-apps-members-list
+            *ngIf="firstSearchDone"
+            [disableRouting]="true"
+          [displayedColumns]="['checkbox', 'id', 'fullName']"
+          [members]="members"
+          [pageSize]="pageSize"
+          [selection]="selection"></perun-web-apps-members-list>
```
xkureck, line 174:
```suggestion
          <perun-web-apps-members-list
            *ngIf="firstSearchDone"
            [disableRouting]="true"
            [displayedColumns]="['checkbox', 'id', 'fullName']"
            [members]="members"
            [pageSize]="pageSize"
            [selection]="selection"></perun-web-apps-members-list>
```

xkureck (changes requested):
image
just remove the "optional" from the name of the stepper and merge it

image
fix eslint problems, also rebase, some eslint problems are already fixed I believe

image
I selected some namespace then I typed something to passwords inputs and then I selected no namespace and the passwords inputs stay as they are

xkureck (approved):

PR #794, opened 2021-08-02
refactor(admin-gui): admin uses correct user profile url when user ha…

xkureck:
…ve no admin role

* when user have no role special page in gui exist for this that will tell him to go to page "my profile" or go to user profile app
* the url of user profile is now got from backend and the name(label) of the user profile application is got from configuration

* the config property will exist when PR https://github.com/CESNET/perun-web-apps/pull/784 will be merged


HejdaJakub (changes requested):
Functionality seems good, I have only one UX thing. The navigate/redirect buttons maybe would look better, if they will have some static position (in the middle). On a big screen is a big space between these buttons. On the other hand, mobile view is little bit broken too.
image
image

HejdaJakub (approved):

PR #795, opened 2021-08-03
perf(admin-gui): tables optimilizations

xkureck:
* optimalizations of some tables rendering the items


bodnara (changes requested):
Just one small thing. Faclities allowed users throws ```ExpresionChangedAfterItHasBeenCheckedError```.
image

Also please rebase.

bodnara (changes requested):
Application list has the same problem as mentioned before. I must have missed it the first time, sorry about that. Everything else is working fine.

bodnara (approved):

PR #797, opened 2021-08-03
feat(admin): add support for Basic Auth

HejdaJakub:
* Now is possible to access admin-gui via basic authentication. This type of authentication is
available on address "/service-access".

### DIFF
--- a/apps/admin-gui/src/app/app-routing.module.ts
+++ b/apps/admin-gui/src/app/app-routing.module.ts
@@ -5,6 +5,7 @@ import { UserDashboardComponent } from './users/pages/user-detail-page/user-dash
 import { NotFoundPageComponent } from './shared/components/not-found-page/not-found-page.component';
 import { RedirectPageComponent } from '@perun-web-apps/perun/components';
 import { LoginScreenComponent } from '@perun-web-apps/perun/login';
+import { LoginScreenServiceAccessComponent } from '@perun-web-apps/perun/login';
 
 const routes: Routes = [
   {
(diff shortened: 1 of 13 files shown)


HejdaJakub:
We need to solve the problem probably with CORS policy before merging.

xkureck (commented):
compeletely missing: feedback to user when the username/password is wrong
the input field for a username and password I would like just a little bit wider
so when I sing in in the page /service-access and succesfully log in then whenever I want to reload the page/use middle click/open the gui and want to log throught oidc the "server error" dialog appears - and thats quite a big deal, for emergency use of GUI its fine, but anything else thats pretty disaster -  so the question is: is the service-acces only for very emergecy uses or not?

HejdaJakub:
> compeletely missing: feedback to user when the username/password is wrong
> the input field for a username and password I would like just a little bit wider
> so when I sing in in the page /service-access and succesfully log in then whenever I want to reload the page/use middle click/open the gui and want to log throught oidc the "server error" dialog appears

- added feedback to user when the username or password is wrong
- input fields are wider now
- fixed reload page and middle click
- also fixed log out from admin gui

bodnara (changes requested):
The middle click still starts oidc sign in , when used on the overview cards / recently viewed entities.
When credentials are wrong I would suggest informing users with the app alert component istead of the dialog poping up.
I still don't fell completly safe about storing name and password but don't have other solution as of now.

HejdaJakub:
> The middle click still starts oidc sign in , when used on the overview cards / recently viewed entities.
> When credentials are wrong I would suggest informing users with the app alert component istead of the dialog poping up.
> I still don't fell completly safe about storing name and password but don't have other solution as of now.

Pop up dialog was replaced by app alert. Middle click on overview/recently viewed cards was fixed in PR #835. In regard to security risks, I discussed it with @zlamalp and we can merge this PR with disabled page '/service-access' for now and we will discuss it  more deeply on Perun seminar.

xkureck (approved):

PR #799, opened 2021-08-04
fix(profile): fix header menu for mobile view

HejdaJakub:
* The header menu used to be placed out of the screen on mobile view.
* In this PR has been also fixed error 'ExpressionChangedAfterItHasBeenCheckedError', which was visible only in development mode.


xkureck (changes requested):
image
on user profile its still kind of iffy, the gui is wider than the viewport, while in the admin gui it looks really nice
image

PR #802, opened 2021-08-04
feat(admin-gui): add failure cause to group - resource relationship

xkureck:
* add failure cause to groups and resource table


bodnara (approved):
Just rebase please.

PR #805, opened 2021-08-09
feat(admin): delete service from service detail page

HejdaJakub:
* Now is possible to delete service from service detail page.
* Service can be also force delete now and for this use case is used universal delete dialog (there is one new div for longer description).


apps/admin-gui/src/app/shared/components/dialogs/delete-entity-dialog/delete-entity-dialog.component.html
```diff
@@ -53,6 +53,9 @@ <h1 mat-dialog-title>{{'DIALOGS.DELETE_ENTITY.DELETE' | translate}}
           </span>
         </mat-list-item>
       </mat-list>
+      <div *ngIf="showAnotherMessage">
+        {{anotherMessage}}
+      </div>
```
xkureck, line 58:
```suggestion
      <div *ngIf="anotherMessage">
        {{anotherMessage}}
      </div>
```

i think that showAnotherMessage = bool; completely useless and we dont need it

xkureck (changes requested):

apps/admin-gui/src/app/shared/components/dialogs/delete-entity-dialog/delete-entity-dialog.component.html
HejdaJakub: Fixed

xkureck (approved):

PR #807, opened 2021-08-10
fix(admin-gui): add constriction to service name

xkureck:
* add regex to service name when creating or editing service
* deleted unused dialog for creating service


xkostka2 (approved):
image

I know it isn't part of this task, but could you do something about the scrollbar on the bottom of the dialog? Doesn't matter if you don't.

PR #810, opened 2021-08-11
fix: Remove reload roles

metodej:
* Removed reload roles from header menu because it is no longer
needed. It was only needed for one demo.


xkostka2 (commented):
You have entered the wrong scope for the commit since the change affects also the admin-gui. I suggest removing the scope from the commit message and PR. 

Apart from that it looks well and can be merged right after the change.

metodej:
> 
> 
> You have entered the wrong scope for the commit since the change affects also the admin-gui. I suggest removing the scope from the commit message and PR.
> 
> Apart from that it looks well and can be merged right after the change.

Good point, thanks. Fixed.

PR #812, opened 2021-08-11
fix(admin-gui): Text style changes in notification editing dialog

bodnara:
* aligned tags text
* changed tag text to be more visible
* moved hide/show button to the bottom of dialog

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.html
+++ b/apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.html
@@ -107,8 +107,13 @@
             <div class="d-flex w-100">
               <div class="pr-2 w-100">
                 <div class="pt-2">
-                  <span class="font-weight-bold pr-2">{{'DIALOGS.NOTIFICATIONS_ADD_EDIT_MAIL.SUBJECT' | translate}}
-                    :</span>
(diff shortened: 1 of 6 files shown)


xkureck (commented):
I dunno, I kind of liked it when the tags was on the side of the dialog, but lets show this to the whole team on the daily scrum and see what they think about it

apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.html
```diff
@@ -120,23 +121,26 @@ <h1 mat-dialog-title>{{'DIALOGS.NOTIFICATIONS_ADD_EDIT_MAIL.TITLE_EDIT' | transl
...
-                <button mat-flat-button (click)="this.showTags = !this.showTags;">
-                  {{showTags ? ('DIALOGS.NOTIFICATIONS_ADD_EDIT_MAIL.HIDE_TAGS' | translate) :
-                  ('DIALOGS.NOTIFICATIONS_ADD_EDIT_MAIL.SHOW_TAGS' | translate)}}
-                </button>
+
                 <mat-form-field class="w-100">
                 <textarea matInput
                           #Textarea
-                          rows="17"
+                          rows="10"
```
xkureck, line 128: so, I would git here more rows than 10, but we will discuss this and everything else on daily scrum

xkureck (approved):

PR #813, opened 2021-08-11
feat(admin): Add create tag to resource page

metodej:
* Button to create a resource tag was added to the assigned tags of
resource page so a user does not have to go back and forth to add
a new tag to resource.


apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-tags/resource-tags.component.ts
```diff
@@ -48,6 +51,10 @@ export class ResourceTagsComponent implements OnInit {
 
     this.route.parent.params.subscribe(params => {
       this.voId = params['voId'];
+      this.voService.getVoById(this.voId).subscribe(vo => {
+        this.vo = vo;
+        this.updateData();
+      });
```
xkureck, line 57:
```suggestion
```
completely remove this, after you have object of resource here you have the voId also and there you can construct objeckt of VO this way:
`this.vo  = {
        id: this.resource.voId,
        beanName: "Vo"
      };`

xkureck (changes requested):

apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-tags/resource-tags.component.ts
metodej: Ok, thanks, I've removed vo and void parameter and moved the vo construction to the `setAuthRights`, since it is the only place it is used.

xkureck (approved):

PR #818, opened 2021-08-17
fix(admin): don't show non-existent objects on dashboard

HejdaJakub:
* Dashboard used to show also objects which were already deleted.
* In this PR was also added tootlip for subgroups on dashboard (tooltip shows whole fullname).

### DIFF
--- a/apps/admin-gui/src/app/users/pages/user-detail-page/user-dashboard/dashboard-recently-viewed-button-field/dashboard-recently-viewed-button-field.component.ts
+++ b/apps/admin-gui/src/app/users/pages/user-detail-page/user-dashboard/dashboard-recently-viewed-button-field/dashboard-recently-viewed-button-field.component.ts
@@ -1,9 +1,15 @@
 import { Component, OnInit } from '@angular/core';
 import { getRecentlyVisitedIds } from '@perun-web-apps/perun/utils';
+import {
+  FacilitiesManagerService,
+  GroupsManagerService,
+  VosManagerService
+} from '@perun-web-apps/perun/openapi';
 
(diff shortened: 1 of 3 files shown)


apps/admin-gui/src/app/users/pages/user-detail-page/user-dashboard/dashboard-recently-viewed-button-field/dashboard-recently-viewed-button-field.component.ts
```diff
@@ -16,14 +22,60 @@ export interface RecentItem {
...
+      }
+    }
+
+    // if no vos or facilities are in recently viewed, post to the backend "-1" to get an empty array
+    if (this.vosIds.length === 0) {
+      this.vosIds.push(-1);
+    }
+    if (this.facilitiesIds.length === 0) {
+      this.facilitiesIds.push(-1);
+    }
```
bodnara, line 62: Add a check for groups as well. Otherwise if no groups are in recently viewed an error is thrown.

bodnara (changes requested):

apps/admin-gui/src/app/users/pages/user-detail-page/user-dashboard/dashboard-recently-viewed-button-field/dashboard-recently-viewed-button-field.component.ts
HejdaJakub: Check for groups added.

bodnara (approved):

PR #824, opened 2021-08-23
refactor: replace oidc-client library by angular-oauth2-oidc

xkostka2:
* Library oidc-client, which was used for authentication, has been replaced by angular-oauth2-oidc, due to end of support for oidc-client
* Response type unified to "id_token token" for all applications


libs/perun/services/src/lib/auth.service.ts
```diff
@@ -144,51 +126,25 @@ export class AuthService {
   }
 
   logout() {
-    this.manager.signoutRedirect()
-      .catch(function(error) {
-        console.log(error);
-      });
+    this.oauthService.revokeTokenAndLogout().catch(function(error) {
+      console.log(error);
+    });
```
HejdaJakub, line 131: There are some problems with logout - application throws an error on revoking token, so the logout is broken.

HejdaJakub (changes requested):

HejdaJakub (approved):

PR #826, opened 2021-08-23
feat(admin): improve recently visited objects dashboard

HejdaJakub:
* Tooltip was added to all entities on dashboard and group tooltip was extended by VO name.
* This PR also fixes group tooltip when user has no recently visited object in local storage.


bodnara (approved):

xkostka2 (changes requested):
I suggest adding spaces around ":" for better readability. Apart from this, I found no other issues, so you can merge it after you fix this.

PR #833, opened 2021-08-24
refactor: unify configuration variables naming convention

bodnara:
* previously we used both camalCase and snake_case to name config
  variables
* names of config variables are now all in snake_case

BREAKING CHANGE: it is necessary to rename variables in config files


HejdaJakub (changes requested):
I found one more camel case variable `backButtonColor` in admin-gui `defaultConfig.json` and this variable is used in component `back-button.component.ts`. Please replace also this variable by snake case on both places. Under this variable in `defaultConfig.json` is also some empty line, so you can delete it.

HejdaJakub (approved):

PR #834, opened 2021-08-25
feat(profile): disable delete Other identities

HejdaJakub:
* Other identities in section Linked identities cannot be deleted now.
* Fixed text in info tooltip.

BREAKING CHANGE: Other identities cannot be deleted any longer.

### DIFF
--- a/apps/user-profile/src/app/pages/identities-page/identities-page.component.html
+++ b/apps/user-profile/src/app/pages/identities-page/identities-page.component.html
@@ -32,7 +32,6 @@
         info_outline
       </mat-icon>
     </h1>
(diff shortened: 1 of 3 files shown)


libs/perun/components/src/lib/user-ext-sources-list/user-ext-sources-list.component.ts
```diff
@@ -48,6 +48,8 @@ export class UserExtSourcesListComponent implements AfterViewInit, OnChanges {
   loginHeader: string;
   @Input()
   disableRouting: boolean;
+  @Input()
+  disableSelect = false;
   @Output()
```
bodnara, line 53: I think this new value is not necessary. You already removed the 'select' column from displayed columns.

HejdaJakub: It's true. I removed this variable.

bodnara (approved):

PR #844, opened 2021-09-06
fix: use better email regex

Vojtech-Sassmann:
* The email regex was too strict, it didn't allow a domain longer than 3 characters.
* Also, we had multiple regexes for email. The regex string is moved to utils
file so it can be reused.


xkureck (changes requested):
missing in generate-sponsored-members-dialog

Vojtech-Sassmann:
> missing in generate-sponsored-members-dialog

Added.

balcirakpeter (approved):

xkureck (approved):

PR #846, opened 2021-09-10
feat(admin): add all assignment statuses to group/resource attribute settings

HejdaJakub:
* On group-resource and resource-group attributes page used to be only groups/resources with active
assignment status. Now there are groups/resources with all statuses (type of assignment status is
visible in search select component).

### DIFF
--- a/apps/admin-gui/src/app/shared/components/two-entity-attribute-page/two-entity-attribute-page.component.ts
+++ b/apps/admin-gui/src/app/shared/components/two-entity-attribute-page/two-entity-attribute-page.component.ts
@@ -14,6 +14,8 @@ import { getDefaultDialogConfig, getRecentlyVisitedIds } from '@perun-web-apps/p
 import { EditAttributeDialogComponent } from '@perun-web-apps/perun/dialogs';
 import { CreateAttributeDialogComponent } from '../dialogs/create-attribute-dialog/create-attribute-dialog.component';
 import { Urns } from '@perun-web-apps/perun/urns';
+import { ResourceWithStatus } from '@perun-web-apps/perun/models';
+import { GroupWithStatus } from '@perun-web-apps/perun/models';
 
 @Component({
   selector: 'app-two-entity-attribute-page',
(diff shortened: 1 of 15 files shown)


libs/perun/components/src/lib/entity-search-select/entity-search-select.component.html
```diff
@@ -22,6 +25,9 @@
       [maxBufferPx]="10 * 48">
       <mat-option *cdkVirtualFor="let entity of filteredEntities | async" [value]="entity">
         {{mainTextFunction(entity)}} <span class="text-muted muted">{{secondaryTextFunction(entity)}}</span>
+        <span *ngIf="displayStatus" class="{{colorByStatus(entity)}}">
+          ({{statusTextFunction(entity)}})
+        </span>
```
xkureck, line 30: what about reusing perun-web-apps-group-resource-status component? with this you must  implement another functionality, which is with this on two places
the problem starts if we change name of the status or add/remove status, then you need to change it in two places
also the way how it looks is not nice

xkureck (changes requested):

libs/perun/components/src/lib/entity-search-select/entity-search-select.component.html
HejdaJakub: I already tried to implement it like this, but I afraid that the icon isn't possible to show in closed search select (it display only name of mat-icon). 

image
image
image

apps/admin-gui/src/app/shared/components/two-entity-attribute-page/two-entity-attribute-page.component.ts
```diff
@@ -15,6 +15,14 @@ import { EditAttributeDialogComponent } from '@perun-web-apps/perun/dialogs';
...
 import { Urns } from '@perun-web-apps/perun/urns';
 
+export interface ResourceWithStatus extends Resource{
+  status?: GroupResourceStatus;
+}
+
+export interface GroupWithStatus extends Group{
+  status?: GroupResourceStatus;
+}
+
```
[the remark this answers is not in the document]
HejdaJakub: This two intarfaces were declared in separate files and usage of old interfaces was replaced.

xkureck (approved):

PR #848, opened 2021-09-13
style: unify info/warn/error alerts

HejdaJakub:
* Design of our ui alerts was unified.
* Fixed wrong usage of info/warn/error alerts in some html files.
* Fixed alert text in publication-systems-list.
* Unified definition of alert type in html files - we mainly use definition alert_type="string", but in some files was definition [alert_type]="'string'" - it was unified to definition alert_type="string".


xkureck:
some of my ideas
image
image
image
image
image

xkureck (approved):

PR #851, opened 2021-09-14
feat(admin-gui): show automatically assigned subgroups/resources

xkureck:
* on page assigned groups and assigned resources we are now showing an
indicator if the group/resource was assigned automatically
* the remove
action is disabled for these groups/resources

### DIFF
--- a/libs/perun/components/src/lib/resources-list/resources-list.component.ts
+++ b/libs/perun/components/src/lib/resources-list/resources-list.component.ts
@@ -47,7 +47,7 @@ export class ResourcesListComponent implements OnInit, OnChanges {
   @Input()
   routingVo = false;
   @Input()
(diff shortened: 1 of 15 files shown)


libs/perun/components/src/lib/resources-list/resources-list.component.ts
```diff
@@ -150,16 +154,18 @@ export class ResourcesListComponent implements OnInit, OnChanges {
     this.dataSource.data = this.resources;
   }
 
+  canBeSelected = (group: GroupWithStatus): boolean => !this.disableSelect(group)
+
```
bodnara, line 158: Perhaps here should be ```ResourceWithStatus``` instead. If so remove the import as well.

bodnara (approved):
Looks good. Small suggestion, maybe put the tooltip above the icon to match tooltip on checkbox.

Also rebase please.

bodnara:
I also realized that the scope should be "admin" not "admin-gui".

bodnara (approved):

PR #853, opened 2021-09-14
fix: add mat ripple to publications and user profile

xkureck:
* add click animation to left menu to publications and user profile


bodnara (changes requested):
In publications app  the ripple effect is not present when "depth" of  page is more than one. E.g. authors -> author-detail is not selected in menu.

xkureck:
@bodnara that was not part of this issue, but I fixed it

bodnara (approved):

PR #855, opened 2021-09-16
feat(admin): improve assigning service packages to resource 

xkostka2:
* improved workflow of assigning service packages to resource

### DIFF
--- a/libs/perun/components/src/lib/entity-search-select/entity-search-select.component.ts
+++ b/libs/perun/components/src/lib/entity-search-select/entity-search-select.component.ts
@@ -110,8 +110,10 @@ export class EntitySearchSelectComponent<T extends PerunBean> implements OnInit,
   }
 
   ngOnChanges(changes: SimpleChanges): void {
+    if (this.entity !== null) {
+      this.entitiesCtrl.setValue(this.entity);
+    }
     if (changes['entities']) {
-
       this.filteredEntities.next(this.entities.slice());
     }
   }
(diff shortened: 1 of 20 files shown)


libs/perun/components/src/lib/entity-search-select/entity-search-select.component.html
```diff
@@ -1,4 +1,4 @@
-<mat-form-field class="w-100">
+<mat-form-field class="w-100 mt-3">
```
xkureck, line 1: with this change you did a change to all entity search selects which could very negatively impact the look on some of the search selects

libs/perun/components/src/lib/entity-search-select/entity-search-select.component.ts
```diff
@@ -22,7 +22,7 @@ import { stringify } from 'querystring';
   templateUrl: './entity-search-select.component.html',
   styleUrls: ['./entity-search-select.component.css']
 })
-export class EntitySearchSelectComponent<T extends PerunBean> implements OnInit, OnChanges, OnDestroy {
+export class EntitySearchSelectComponent<T extends PerunBean> implements OnInit, OnChanges, OnDestroy, OnChanges {
```
xkureck, line 25: we dont need to declade twice the "OnChanges" I guess

xkureck (changes requested):
when I switch tab to "select service package" no option is selected - even the "Not selected" option is not selected, so the question is - do we even need it? or make it work if we want it
image

xkureck (changes requested):
Just a super small thing - 
image
Add some space (mt-1 or mt-2) between the tab and the debounce filter

PR #857, opened 2021-09-20
feat(admin): add table backend paginating to admin users page

xkureck:
* add new table dynamic-users-table that support backend paginating of
data
* new table is used in admin-users page


apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -2053,7 +2053,7 @@
       "SEARCH_INFO": "You can search users by name, email or login or you can list all users without vo.",
       "NO_USERS_FOUND": "No users found.",
       "EMPTY_SEARCH": "Search field cannot be empty",
-      "USERS_WITHOUT_VO": "Find users without organization"
+      "USERS_WITHOUT_VO": "Search in users without organization"
     },
```
bodnara, line 2057: Personaly I would prefer either "Find users ..." or "Search for users ...". "Search in users ..." feels unnatural.

bodnara (commented):
Seems like the duplication problem is still present. We should wait for this to be fixed before merging.
image

bodnara (approved):
We can merge this once backend is fixed.

PR #862, opened 2021-09-22
feat(admin): allow setting member expiration in future

xkostka2:
* if a member is in 'EXPIRED' state, changing his/her expiration to a date in future or to 'never' now changes also member's status to 'VALID'

### DIFF
--- a/apps/admin-gui/src/app/shared/components/member-overview-membership/member-overview-membership.component.ts
+++ b/apps/admin-gui/src/app/shared/components/member-overview-membership/member-overview-membership.component.ts
@@ -76,8 +76,11 @@ export class MemberOverviewMembershipComponent implements OnChanges {
     }
 
     const dialogRef = this.dialog.open(ChangeVoExpirationDialogComponent, config);
-    dialogRef.afterClosed().subscribe(success => {
-      if (success) {
+    dialogRef.afterClosed().subscribe(result => {
+      if (result.success) {
+        if(result.member){
+          this.member = result.member;
+        }
         this.refreshVoExpiration();
       }
     });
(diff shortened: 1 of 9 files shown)


apps/admin-gui/src/app/shared/components/member-overview-membership/member-overview-membership.component.ts
```diff
@@ -78,6 +78,9 @@ export class MemberOverviewMembershipComponent implements OnChanges {
     const dialogRef = this.dialog.open(ChangeVoExpirationDialogComponent, config);
     dialogRef.afterClosed().subscribe(success => {
       if (success) {
+        if(success !== true){
+          this.member = success;
+        }
```
bodnara, line 83: Honestly this is not a good code practice. My suggestion is to return tuple (success, meber) from the dialog.

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1680,7 +1680,8 @@
       "CANCEL": "Cancel",
       "SAVE": "Save",
       "SUCCESS": "Expiration successfully changed",
-      "SAVE_HINT": "You don't have permission to change expiration date."
+      "SAVE_HINT": "You don't have permission to change expiration date.",
+      "STATUS_CHANGE_INFO": "This change will result in changing the status to 'VALID'."
     },
```
bodnara, line 1685: I would change this to be something like "This action will result in ...".

bodnara (changes requested):
When changing expiration of an expired member it is now possible to choose dates both form past and the future. What I undestood was to only make future dates avaliable but maybe I am mistaken.

bodnara (approved):

xkureck (approved):

PR #863, opened 2021-09-24
fix: check password via backend and display help according to namespace

HejdaJakub:
* Unified password form for create serice member, create sponsored member and change password dialogs.
* In these components is password now checked via backend and potential error from backend is displayed in tooltip (error icon in password input).
* Password help is now defined in config file. This help is changing according to namespace.


bodnara (changes requested):
Ther is some null value present in the tooltip.
image

"mu" namespace is missing tooltip.
image

In the change password dialog the confirm button is enabled for a split second before it's checked.
image

The whole section with password change has broken text in the admin-gui. Including the dialog. The translations are missing and are only present in user-profile. However it would maybe be better to fix that in a separate PR, unless you want to make these changes here.

HejdaJakub:
Changes:
- null value was removed from tooltip
- "mu" namespace was tested on idm-test and it works correctly
- confirmation button isn't enabled now before the check is completed
- password-change translation in admin-gui has been fixed

I had to move password-form component and the block of code from perun-utils.ts to the separate module due to circular dependecy, so there are some new changes.

bodnara (approved):

PR #864, opened 2021-09-27
fix(lib): disable deleting persistent ues

HejdaJakub:
* Disable deleting ues, which cannot be deleted.
* Fixed error in development mode - NG0100: Expression has changed after it was checked.


bodnara (changes requested):
I would add some tooltip to inform users as to why are these items disabled.

PR #867, opened 2021-09-29
AdminGui - e2e tests

Vojtech-Sassmann:
* From now, it is possible to run e2e test for the admin-gui.
* To do so, one has to run the `perun-rpc` locally on port 8081, and there has to be a `perun` user with the `test` password.
* Also, some old dependencies were removed.
* Created `README.md` file in the `admin-gui-e2e` app with the description of how to run the e2e tests.
* Some values in the `defaultConfig.json` file contained the configuration of oidc_client from the devel instance. However, we should avoid this. These values were removed or replaced by neutral values.


HejdaJakub (changes requested):
Please fix the lint error in the `side-menu-root-item.component.ts`. Otherwise, it looks good.

HejdaJakub (approved):

xkureck (approved):
its cool

PR #868, opened 2021-09-30
feat: table wrapper extension

xkureck:
* extended table wrapper
* added export button to bottom paginator
also
* setting of pagesize and saving it moved to the component of table
wrapper
* alert directly connected if table is empty or if the filtering
in table has no results moved to table itself


xkostka2 (changes requested):
image

I suggest reverting this change so that the alert is not inside the table

apps/admin-gui/src/app/vos/components/application-form-list/application-form-list.component.html
```diff
@@ -231,11 +231,15 @@
           cdkDrag
           mat-row>
         </tr>
+        <tr *matNoDataRow>
+          <td [attr.colspan]="displayedColumns.length">
+            <app-alert *ngIf="this.applicationFormItems.length === 0 && !loading" alert_type="warn">
+              {{'VO_DETAIL.SETTINGS.APPLICATION_FORM.NO_APPLICATION_FORM' | translate}}
+            </app-alert>
+          </td>
+        </tr>
```
bodnara, line 240: This alert is still inside of the table. This one make sense so, if it is on purpose we can merge this.

PR #869, opened 2021-09-30
feat(profile): force language in URL

bodnara:
* profile reacts to query parameter 'lang' that acts as the mail language


xkureck (changes requested):
I think that @zlamalp wanted to translate also the login page - translating there is not working, also the parameter lang on the login page somehow get lost, also on the login page he wanted to 1. use language from url 2. use the language according to the browser

also small bug 
image
when I got on page with lang cs everything is fine, then I want to change the language for some reason to english, which is fine, then I go on other page, then I come back and the language field is in czech even tho the gui is in english, also when I reload the gui the gui is back to czech -> my point is I user make a his own decision to change the language of the profile shouldnt we respect it? and delete the queryparam lang? @bodnara discuss it with @zlamalp on our daily scrum maybe

libs/perun/services/src/lib/init-auth.service.ts
```diff
@@ -111,9 +111,17 @@ export class InitAuthService {
...
     } else {
       this.setLoginScreen(true);
+      const query = location.search.substr(1).split('&');
+      let queryParams = null;
+      for(const param of query) {
+        const p = param.split('=');
+        if(p[0] === 'lang'){
+          queryParams = {lang: p[1]};
+        }
+      }
```
xkureck, line 121: I think we want all the queryParams to be redirected to login page, not just lang

libs/perun/login/src/lib/login-screen/login-screen.component.ts
```diff
@@ -1,16 +1,29 @@
...
+              private store: StoreService) { }
+
+  ngOnInit() {
+    this.route.queryParams.subscribe(params => {
+      const langs = this.store.get('supported_languages');
+      const urlLang = params['lang'];
+      if(urlLang && langs.includes(urlLang)) {
+        this.translateService.use(urlLang);
+      }
+    });
```
xkureck, line 25: and what about browser preferences? can we somehow get the browser preference for language?

apps/user-profile/src/app/app.component.ts
```diff
@@ -33,11 +35,17 @@ export class AppComponent implements OnInit {
...
+      this.route.queryParams.subscribe(params => {
+        const langs = this.store.get('supported_languages');
+        const urlLang = params['lang'];
+        if(urlLang && langs.includes(urlLang)) {
+          this.translateService.use(urlLang);
+        } else if(prefLang && prefLang.value) {
+          // @ts-ignore
+          this.translateService.use(prefLang.value);
+        }
+      });
```
xkureck, line 48: isnt this block of code repeating throught the application? if so store it in one place and call it from all the places where you need it

PR #880, opened 2021-10-13
feat(admin): rework facility assigned users page

bodnara:
* added additional user information such as email and organization
* added more filter options: allowed/assigned, by facility/resource/service
* table is dynamicly loaded


apps/admin-gui/src/app/facilities/pages/facility-detail-page/facility-allowed-users/facility-allowed-users.component.html
```diff
@@ -1,24 +1,53 @@
...
+    <perun-web-apps-debounce-filter
+      class="mr-2 filter"
+      [placeholder]="'FACILITY_DETAIL.ALLOWED_USERS.FILTER'"
+      (filter)="applyFilter($event)"></perun-web-apps-debounce-filter>
+    <mat-slide-toggle
+      [(ngModel)]="allowed"
+      class="mr-2"
+      labelPosition="before">
+      {{'FACILITY_DETAIL.ALLOWED_USERS.FILTER_ALLOWED' | translate}}
+    </mat-slide-toggle>
```
xkureck, line 17: I think there needs to be better caption of what is going on
if you would do it with checkbox then this label would be enough with some tooltip - I dont think that users know what is the difference between allowed/assigned users
if you want to do this with the slider then I think we need the label on both side of it (just my opinion of course)

xkureck (changes requested):

xkureck:
Please show this on tomorrows daily scrum. The filtering is not working for email - that could be potentially problem. The tooltip of the slider could be something that could be discussed.

apps/admin-gui/src/app/facilities/pages/facility-detail-page/facility-allowed-users/facility-allowed-users.component.html
```diff
@@ -1,17 +1,58 @@
...
+      class="search-select"
+      (serviceSelected)="serviceSelected($event)"
+      [service]="selectedService"
+      [services]="filteredServices">
+    </perun-web-apps-service-search-select>
+  </div>
+  <mat-spinner class="ml-auto mr-auto" *ngIf="loading"></mat-spinner>
+  <perun-web-apps-users-dynamic-list
+    *ngIf="!loading"
+    [noUsersFoundLabel]="'FACILITY_DETAIL.ALLOWED_USERS.NO_USERS' | translate"
```
xkureck, line 46: please make this property work or delete it if you think that the default message is good enought

PR #881, opened 2021-10-13
feat(admin-gui): new section - Service account in My profile section

xkureck:
*  new page on My profile section "Service account"
* new section with
user type: service in My profile section which should be used by user to
manage his service accounts
* in admin section pages "service
identitites" and "associated users" moved from advanced menu
* service
identitites renamed to "service accounts"


bodnara (approved):

bodnara:
I realized the scope should be `admin` not `admin-gui`.

bodnara (approved):

PR #883, opened 2021-10-14
feat(admin): change of status is now allowed from members table

xkureck:
* added change of status from members dynamic list


bodnara (approved):
Changing status from table works fine.

I would just say that this whole action of changing status and as a result expiration as well is a little bit messy. It allows thing such as changing status to `Expired`  and choosing expiration to be `never` which result in status beeing `Active`. Other 'ilogical' results can be achieved (at least temporarily) simpliy by choosing inconsistend status and expiration or canceling the expiration dialog.

PR #891, opened 2021-10-20
fix(admin): fix manual member adding block

bodnara:
* removed blocking for organizations as this is not restricted
* disabled selection of candidates that are not in organization if
  manual adding is blocked

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/add-member-dialog/add-member-dialog.component.html
+++ b/apps/admin-gui/src/app/shared/components/dialogs/add-member-dialog/add-member-dialog.component.html
@@ -34,6 +34,7 @@
         [selection]="selection"
         [members]="members"
         [type]="data.type"
+        [blockManualAdding]="manualAddingBlocked"
         *ngIf="firstSearchDone">
       </app-members-candidates-list>
     </div>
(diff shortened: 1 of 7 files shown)


apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-members/vo-members.component.ts
```diff
@@ -80,12 +76,10 @@ export class VoMembersComponent implements OnInit {
...
-      this.isManualAddingBlocked(voId).then(() => {
-        this.voService.getVoById(voId).subscribe(vo => {
-          this.vo = vo;
-          this.setAuthRights();
-          this.loading = false;
-        });
+      this.voService.getVoById(voId).subscribe(vo => {
+        this.vo = vo;
+        this.setAuthRights();
+        this.loading = false;
```
xkureck, line 82: but we need to check if manual adding to VO is allowed
this is misuderstanding probably - if blockManualMemberAdding(VO attribute) to VO is true, then we need to forbid adding members to VO and GROUP from outside of VO
blockManualMemberAdding(group attribute) is true then we need to forbid adding members overall

PR #896, opened 2021-10-26
feat: Set lang and auth filter in password-reset

bodnara:
* password-reset app now reacts to language and auth filter in url

### DIFF
--- a/apps/password-reset/src/app/app.component.ts
+++ b/apps/password-reset/src/app/app.component.ts
@@ -4,6 +4,8 @@ import { PasswordResetDialogComponent } from './dialogs/password-reset-dialog/pa
 import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
 import { UsersManagerService } from '@perun-web-apps/perun/openapi';
 import { TokenExpiredDialogComponent } from './dialogs/token-expired-dialog/token-expired-dialog.component';
+import { PreferredLanguageService } from '@perun-web-apps/perun/services';
+import { TranslateService } from '@ngx-translate/core';
 
 @Component({
   selector: 'perun-web-apps-root',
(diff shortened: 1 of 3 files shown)


apps/password-reset/src/app/services/password-reset-config.service.ts
```diff
@@ -14,15 +14,29 @@ export class PasswordResetConfigService {
...
+    private authzSevice: AuthzResolverService
   ) { }
 
   loadConfigs(): Promise<void> {
     return this.appConfigService.loadAppDefaultConfig()
       .then(() => this.appConfigService.loadAppInstanceConfig())
       .then(() => this.setApiUrl())
-      .then(() => this.appConfigService.setInstanceFavicon());
+      .then(() => this.appConfigService.setInstanceFavicon())
+      .then(() => this.initAuthService.verifyAuth())
```
xkureck, line 25: the authentication should happen only and only if in the url ISNT "token" parameter

apps/password-reset/src/app/app.component.ts
```diff
@@ -13,18 +15,22 @@ import { TokenExpiredDialogComponent } from './dialogs/token-expired-dialog/toke
...
+    let mode = '';
     if (queryParams.includes('activation')) {
-      mode = 'activation'
+      mode = 'activation';
     } else if(queryParams.includes('token')) {
-      mode = 'reset'
+      mode = 'reset';
     } else {
-      mode = 'change'
+      mode = 'change';
```
xkureck, line 33: we disscuss with pavel that mode change will not be in password reser app
please remove it

xkureck (changes requested):

apps/password-reset/src/app/app.component.ts
bodnara: I don't think this should be part of this PR as this was about setting language and auth filter from query params.

apps/password-reset/src/app/services/password-reset-config.service.ts
```diff
@@ -14,15 +14,36 @@ export class PasswordResetConfigService {
...
   loadConfigs(): Promise<void> {
     return this.appConfigService.loadAppDefaultConfig()
       .then(() => this.appConfigService.loadAppInstanceConfig())
       .then(() => this.setApiUrl())
-      .then(() => this.appConfigService.setInstanceFavicon());
+      .then(() => this.appConfigService.setInstanceFavicon())
+      .then(() => {
+        const queryParams = location.search.substr(1);
+        if(!queryParams.includes('token')) {
+          return this.initAuthService.verifyAuth();
```
xkureck, line 28: the authentication is not working at all, I think the problem is in the thing that you dont wait for the result, there is not a promise in this {}

xkureck (changes requested):

apps/password-reset/src/app/services/password-reset-config.service.ts
xkureck: and rebase please

xkureck (approved):

PR #898, opened 2021-10-26
feat: migration from implicit flow to PKCE

xkureck:
* from implicit flow the default configuration is now code flow
* the
refreshing of access token is setup for code flow only


bodnara (commented):
Fix eslint.

bodnara (approved):

PR #902, opened 2021-11-02
feat(admin): warn before disconnecting users from service accounts

cuadradek:
- Users are now warned when they are disconnecting themselves from
service accounts or when the last identity is about to be disconnected.


bodnara (commented):
Well done. One thing that is up for discussion is whether to show both alerts when removing self as last assosiated user or not.

PR #906, opened 2021-11-04
fix: urls to other applications based on brand configuration

HejdaJakub:
* Now are used all urls to other applications according to brand configuation.
* If no brand fix to domain, then will be used default brand.
* If the url for new app isn't defined for this brand, then will be used url for old gui.

BREAKING CHANGE: deleted "pwd_reset_base_url" property from defaultConfig files

### DIFF
--- a/apps/admin-gui/src/app/core/services/common/admin-gui-config.service.ts
+++ b/apps/admin-gui/src/app/core/services/common/admin-gui-config.service.ts
@@ -112,6 +112,7 @@ export class AdminGuiConfigService {
           return this.initAuthService.loadPrincipal()
             .catch(err => this.handleErr(err))
             .then(() => this.loadPolicies())
+            .then(() => this.appConfigService.loadAppsConfig())
             .then(() => this.guiAuthResolver.loadRolesManagementRules());
         } else {
           return this.initAuthService.handleAuthStart();
--- a/libs/perun/components/src/lib/password-reset/password-reset.component.ts
+++ b/libs/perun/components/src/lib/password-reset/password-reset.component.ts
@@ -1,6 +1,6 @@
 import { Component, OnInit } from '@angular/core';
-import { Attribute, AttributesManagerService} from '@perun-web-apps/perun/openapi';
-import { StoreService } from '@perun-web-apps/perun/services';
+import { Attribute, AttributesManagerService } from '@perun-web-apps/perun/openapi';
+import { OtherApplicationsService, StoreService } from '@perun-web-apps/perun/services';
 import { MatTableDataSource } from '@angular/material/table';
 import { MatDialog } from '@angular/material/dialog';
 import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
@@ -26,7 +26,8 @@ export class PasswordResetComponent implements OnInit {
               private store: StoreService,
               private dialog: MatDialog,
               private route: ActivatedRoute,
-              private router: Router) {
+              private router: Router,
+              private otherApplicationService: OtherApplicationsService) {
   }
 
   ngOnInit(): void {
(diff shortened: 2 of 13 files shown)


libs/perun/components/src/lib/password-reset/password-reset.component.ts
```diff
@@ -53,8 +55,9 @@ export class PasswordResetComponent implements OnInit {
   }
 
   resetPassword(login: string) {
-    const url = this.store.get('pwd_reset_base_url');
-    location.href = `${url}?login-namespace=${login}`;
+    this.utilsService.getAppsConfig().subscribe(config => {
```
xkureck, line 58: you are calling this method every time you need the config, cant the config be stored at something like "StoreService" and load it at the start of the application?

xkureck (changes requested):

libs/perun/services/src/lib/force-router.service.ts
```diff
@@ -53,4 +71,51 @@ export class ForceRouterService {
...
+        case "admin":
+          url += "/gui/"
+          break;
+        case "profile":
+          url += "/profile/"
+          break;
+        case "pwdReset":
+          url += "/pwd-reset/"
+          break;
+      }
```
xkureck, line 111: this code is not working
dunno what you meant by getIdpFilter(
howewe in old application there needs to be before the name of the application the auth type, so something like /fed/

xkureck (changes requested):
please rebase

libs/perun/components/src/lib/password-reset/password-reset.component.ts
HejdaJakub: Good point, I rewrote it.

libs/perun/services/src/lib/force-router.service.ts
HejdaJakub: Fixed

apps/admin-gui/src/app/core/services/common/admin-gui-config.service.ts
```diff
@@ -175,4 +177,13 @@ export class AdminGuiConfigService {
...
   }
+
+  private loadAppsConfig(): Promise<void>  {
+    return new Promise((resolve, reject) => {
+      this.utilsService.getAppsConfig().subscribe( appsConfig => {
+        this.store.setAppsConfig(appsConfig);
+        resolve();
+      }, error => reject(error));
+    });
+  }
```
xkureck, line 188: move this method to the "appConfigService" and also don't forget to call it also from user-profile-config.service!

xkureck: also call it from password reset and publications if you are using the forceRoute service

libs/perun/components/src/lib/password-reset/password-reset.component.ts
```diff
@@ -26,7 +26,8 @@ export class PasswordResetComponent implements OnInit {
               private store: StoreService,
               private dialog: MatDialog,
               private route: ActivatedRoute,
-              private router: Router) {
+              private router: Router,
+              private forceRoute: ForceRouterService) {
```
xkureck, line 30: the name of service is kind of misleading, can it be named better?

xkureck (changes requested):

apps/admin-gui/src/app/core/services/common/admin-gui-config.service.ts
HejdaJakub: Method has been moved and method call has been added to user-profile-config.service.

In publications and password reset applications we don't use links to other applications, so in my opinion, there it's not necessary.

xkureck (approved):

PR #907, opened 2021-11-04
fix(admin): change of service-identity icon

metodej:
* icons for service-identities change to robots


xkureck (changes requested):
image
just add the icon also on this place and it will be fine

metodej:
> 
> 
> image just add the icon also on this place and it will be fine

Fixed, thanks!

xkureck (approved):

PR #908, opened 2021-11-09
feat: Ellipsis in action button text

bodnara:
* added ellipsis to all buttons that start some non trivial action


HejdaJakub (changes requested):
Good job. I found few more places, where ellipsis probably should be added too.

(Admin GUI) My profile -> Advanced settings -> Password reset -> Change password
The same also in user profile -> Change password
image

(Admin GUI) Member -> Send request for password reset
image

There are ellipsis partly over the text (Profile -> Settings -> SSH keys)
image

HejdaJakub (approved):

PR #909, opened 2021-11-09
feat(admin): add option to generate login for service account

Johaney-s:
* in some namespaces login should be generated rather than filled in for service accounts
* this information is now extracted from namespaces rules same as in sponsored user
* changed getting namespaces options from gui config to use backend config, same as for sponsored users


apps/admin-gui/src/app/shared/components/create-service-member-dialog/create-service-member-dialog.component.ts
```diff
@@ -151,11 +176,39 @@ export class CreateServiceMemberDialogComponent implements OnInit {
     }, () => this.processing = false);
   }
 
+  parseNamespaceRules(){
+    for (const rule of this.namespaceRules) {
+      this.namespaceOptions.push(rule.namespaceName);
+
+      const fieldTypes =  {login: 'disabled'};
+      this.parseAttributes(fieldTypes, rule.requiredAttributes, 'required');
+      this.parseAttributes(fieldTypes, rule.requiredAttributes, 'optional');
```
bodnara, line 185: Use `optionalAttributes` with `optional` type.

bodnara (changes requested):
Other than that one small change this looks good. Also creating service account with generated password ends with `PasswordCreationFailedException` but perhaps thats not part of this issue.

apps/admin-gui/src/app/shared/components/create-service-member-dialog/create-service-member-dialog.component.ts
Johaney-s: Thanks, fixed. For the error, I suspect it is caused by missing password manager implementations for specific namespaces in development environment.

bodnara (approved):

xkureck (changes requested):
what is file named "save" and why do we need it in our repository?

Johaney-s:
> what is file named "save" and why do we need it in our repository? otherwise it looks fine, but it needs to be tested on something like idm-test

I removed the file.

xkureck (approved):

PR #910, opened 2021-11-09
fix: center checkboxes in tables

bodnara:
* centered "selection" checkboxes in all tables containing them


xkureck (changes requested):
please add it to string-list.html

xkureck (approved):
looks fine, just rebase please

PR #916, opened 2021-11-10
feat(admin): service identity terminology unification

xkostka2:
* several texts related to service account management were unified

### DIFF
--- a/apps/admin-gui/src/assets/i18n/en.json
+++ b/apps/admin-gui/src/assets/i18n/en.json
@@ -887,7 +887,7 @@
       "PASSWORD_RESET": "Password reset",
       "FACILITIES": "Facilities",
       "RESOURCES": "Resources",
-      "ASSOCIATED_USERS": "Associated users",
+      "ASSOCIATED_USERS": "Account managers",
       "ENTITY": "User",
       "USER_TYPE": "User type",
       "ACCOUNTS": "Accounts"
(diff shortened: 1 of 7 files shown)


apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -2224,19 +2224,19 @@
...
         "TITLE": "Gui Configuration"
       },
       "ASSOCIATED_USERS": {
-        "TITLE": "Associated users",
-        "ADD": "Add",
-        "REMOVE": "Remove",
-        "NO_ASSOCIATED_USERS": "No associated users assigned."
+        "TITLE": "Account managers",
+        "ADD": "Connect account",
+        "REMOVE": "Disconnect account",
```
xkureck, line 2238: I dont think we need the word "account" in those buttons
in the remove button I think the remove its just fine
maybe the add here was also just fine

xkureck (changes requested):
image
on this dialog change title to "connect account"
image
title: "disconect service account"
subtitle: "Following account will be disconnected from you."

be careful this dialog are also used in perun admin section and then we need to change the text a bit

image
in this dialog title: "connect user"
image
this dialog needs also update

also on this dialogs don't forget they are used in perun admin section

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1718,24 +1718,27 @@
...
     },
     "DISCONNECT_IDENTITY": {
-      "TITLE": "Disconnect identity",
+      "TITLE_USER": "Disconnect user",
+      "TITLE_SELF": "Disconnect service account",
       "REMOVE": "Disconnect",
       "CANCEL": "Cancel",
-      "DESCRIPTION": "Following identities will be disconnected from the user.",
+      "DESCRIPTION_USER": "Following accounts will be disconnected from the user.",
+      "DESCRIPTION_SELF": "Following accounts will be disconnected from you.",
```
xkureck, line 1737: in this label dont forget that we are also using the same label in admin section 
use different labels or write the label so there wont be word "you"
image

xkureck (changes requested):

xkureck (approved):

PR #920, opened 2021-11-13
fix(admin): fix assign group/assign resource button auth

xkureck:
* on page resource-groups the button for assign was hidden
for
combination of roles trustedFacilityAdmin+FacilityAdmin
and
resource-selfService + groupAdmin
* on dialogs
assign-group-to-resource and add-group-resource the filtration of
entities shown in table added so the user in table sees only the
entities which he 1. have the priviledge to assign 2. are not already
assigned


HejdaJakub (approved):
Please rebase this PR. Otherwise it looks good.

PR #925, opened 2021-11-16
fix(admin): reserve password for service account with correct rights

Johaney-s:
* reservePasswordForLogin fails for voadmin with insufficient rights
* reservePasswordForUser can distinguish service account and allow voadmin


xkureck (approved):
if its tested and it works I guess I must believe you

PR #928, opened 2021-11-19
feat(admin): improve audit message detail component

HejdaJakub:
* Added "Actor" and "Created at" to this dialog.
* Added new tab for audit message and copy button to both tabs.
* JSON object syntax is now highlighted.


bodnara (changes requested):
Well done. especially the syntax highlighting. Two small changes I would recommend are:
1. Set also the `width` of dialog because it's pretty slim when opened
2. Make the copy button more visible, either put it on top or in the actions panel (or something else). Since right now users must scroll all the way down to find it. This is an issue mostly on the `Objects` tab.

HejdaJakub:
> Well done. especially the syntax highlighting. Two small changes I would recommend are:
> 
> 1. Set also the `width` of dialog because it's pretty slim when opened
> 2. Make the copy button more visible, either put it on top or in the actions panel (or something else). Since right now users must scroll all the way down to find it. This is an issue mostly on the `Objects` tab.

Added minWidth to dialog config and copy button was moved to section mat-dialog-actions.

bodnara (approved):

PR #930, opened 2021-11-23
feat(admin): improve UX of dialogs that use stepper

xkostka2:
* action buttons of stepper like "Back" were moved outside the stepper


apps/admin-gui/src/app/shared/components/dialogs/add-member-to-resource-dialog/add-member-to-resource-dialog.component.html
```diff
@@ -25,8 +25,8 @@ <h1 mat-dialog-title>{{'DIALOGS.ADD_MEMBER_TO_RESOURCE.TITLE' | translate}}</h1>
               <mat-select
                 disableOptionCentering="true"
                 placeholder="{{ 'DIALOGS.ADD_MEMBER_TO_RESOURCE.SELECTED' | translate}}">
-                <mat-option (click)="setResource(resource);
-                           stepper.selected.completed = true" *ngFor="let resource of filteredResources | async"
+                <mat-option (click)="setResource(resource)"
+                            *ngFor="let resource of filteredResources | async"
```
xkureck, line 29: I think that stepper.selected.completed = true were there maybe for reason, because now you can just simply click on tab "Add group" and it will go to that tab even when you don't have any resource selected

xkureck (changes requested):

PR #931, opened 2021-11-24
feat: allow logos in the footer

xkostka2:
* footer configuration has been extended by supporting logos as elements of the columns


xkureck (commented):
So I tried and it works.
There must be always some URL link to the image, is Pavel fine with that?
also the "label" to the image is kind of useless, but I guess its not against anything

zlamalp (changes requested):
It works, but images are forcibly sized to the line height? (20px), which is not desired.

I simply want multiple logos to be stacked inside the column and not forcibly resize column width. If its too complex to handle column width its not necessary to do it. I can already adjust image sizes/spacing etc. using styles within the image tag itself in the config file.

apps/admin-gui/src/assets/config/defaultConfig.json
```diff
@@ -93,6 +93,16 @@
...
+            "link_en": "https://perun-aai.org/",
+            "label_en": ""
```
Vojtech-Sassmann, line 105: I don't think that by default we want for all instances to show these logos. This shouldn't be part of the default config.

Vojtech-Sassmann (changes requested):

apps/admin-gui/src/assets/config/defaultConfig.json
xkostka2: this was just for the review, I will remove it once the review is done

Vojtech-Sassmann: Things like this might get accidentally merged. It would be better to put the logo in a comment to this PR.

xkureck (changes requested):
image
so the behaviour is now that the logos will be in the separate column and in the column can be only logos, @zlamalp are you fine with that?

image
I think we want the "headers" of the column to be on the same line, in this case on the screenshot to be "about" and "support" on the same line as it "help"

and for some reason, the icon next to the "sponsors in organization" completely disappear when some logo is in the footer
image

libs/perun/components/src/lib/perun-footer/perun-footer.component.html
```diff
@@ -1,22 +1,32 @@
 <footer [ngStyle]="{'background': bgColor}" class="footer-container">
   <div class="footer-columns-container">
-    <div *ngFor="let item of items" class="footer-col mx-auto">
+    <div *ngFor="let item of items" [class.my-auto]="containsLogos" class="footer-col mx-auto">
```
HejdaJakub, line 3: Still headers are not at the same position.
image

HejdaJakub (changes requested):
If I add only one logo to the first column, then all columns have the height according to the first (smaller) column.
image

Vojtech-Sassmann (changes requested):
EDIT: sorry, wrong PR

Vojtech-Sassmann (approved):

HejdaJakub (approved):

PR #937, opened 2021-11-26
feat: implementation of password reset application

HejdaJakub:
* Password reset application was implemented according to prototype.
* Use case for password change (with typing old password) was removed from this app.
* Already existing component password-form was reused in this app, but for this purpose it had to be modified.
* To all password inputs was added eye icon which allows user to show/hide password.

BREAKING CHANGE: added new properties in defaultConfig.json and removed old dialogs


HejdaJakub:
Value of `password_reset_logo` should be removed from `defaultConfig.json` before merge. I kept this value only for testing.

Vojtech-Sassmann (changes requested):
There is a missing value for this translation: 
image

HejdaJakub:
> There is a missing value for this translation: image

Translations for `en` and `cs` languages have been added.

bodnara (approved):

Vojtech-Sassmann (changes requested):
In the defaultConfig, we shouldn't keep values for our devel instance. This might cause some problems in production. If instanceConfig would contain some invalid values (or some were missing) we don't want users to be redirected to our devel proxy.

Please, replace `api_url` and `oidc_client` values with similar values which are already present in the `admin-gui` default config.

HejdaJakub:
Vaules in `api_url` and `oidc_client` were replaced and usage of eye icons fixed.
Word `printing` was replaced by `printable` in admin-gui, user-profile and password-reset applications.

Vojtech-Sassmann (changes requested):
I have found one small problem: if I enter the "again field" first and then the first password field, it says that passwords don't match even if they do match.

Another thing - it might be worth discussing if we need the alert with the description of password requirements. IMHO, it is redundant and I don't think it is common. If you provide a detailed explanation, why the provided password is not sufficient, it is redundant.

HejdaJakub:
> I have found one small problem: if I enter the "again field" first and then the first password field, it says that passwords don't match even if they do match.
>
Error `passwords don't match` has been fixed.

> Another thing - it might be worth discussing if we need the alert with the description of password requirements. IMHO, it is redundant and I don't think it is common. If you provide a detailed explanation, why the provided password is not sufficient, it is redundant.
>
As regards password requirements, I discussed it with @balcirakpeter and we can keep it for this moment and we are going to discuss it on review meeting with all stakeholders.

I also removed testing logo.

Vojtech-Sassmann (changes requested):
image

The check icon is not aligned with the rest of the content. 

I would align it with the right side. Also, I would consider using more whitespace above it, so it is visually clear, that it is not part of the login/namespace description.

image

And one more thing. There is a `success` alert in our library. Could it be used here instead? It would make sense to use a single design of such components across our applications.

HejdaJakub:
> And one more thing. There is a `success` alert in our library. Could it be used here instead? It would make sense to use a single design of such components across our applications.
>
Heading tag was replaced by `success` alert and top margin was added too.
image

Vojtech-Sassmann (approved):

PR #941, opened 2021-12-08
fix(admin): group-resource assignment dialog for active/inactive status adjusted

Kuliak:
* dialog changed to inform user that only source group will be assigned with INACTIVE status (if chosen), subgroups should always be assigned with ACTIVE status

### DIFF
--- a/apps/admin-gui/src/assets/i18n/en.json
+++ b/apps/admin-gui/src/assets/i18n/en.json
@@ -1435,7 +1435,7 @@
       "ASYNC_OPT_DESCRIPTION": "If selected, group(s) assignment is executed asynchronously (switched to PROCESSING state and waiting for assignment to run in the background). Otherwise, a synchronous assignment is executed.",
       "AS_ACTIVE": "Assign with active status",
       "ACTIVE_ON_HINT": "Groups will be assigned to the resource with ACTIVE status",
-      "ACTIVE_OFF_HINT": "Groups will be assigned to the resource with INACTIVE status, members will not be able to access services until activation",
+      "ACTIVE_OFF_HINT": "Group(s) will be assigned to the resource with INACTIVE status (if assigning with subgroups, subgroups will be assigned with ACTIVE status), members of INACTIVE group(s) will not be able to access services until activation.",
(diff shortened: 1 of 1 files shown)


apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1435,7 +1435,7 @@
       "ASYNC_OPT_DESCRIPTION": "If selected, group(s) assignment is executed asynchronously (switched to PROCESSING state and waiting for assignment to run in the background). Otherwise, a synchronous assignment is executed.",
       "AS_ACTIVE": "Assign with active status",
       "ACTIVE_ON_HINT": "Groups will be assigned to the resource with ACTIVE status",
-      "ACTIVE_OFF_HINT": "Groups will be assigned to the resource with INACTIVE status, members will not be able to access services until activation",
+      "ACTIVE_OFF_HINT": "Source group(s) will be assigned to the resource with INACTIVE status (subgroups remain with ACTIVE status), members of INACTIVE group(s) will not be able to access services until activation",
```
Johaney-s, line 1438: I think it might not be clear to users what a source group is, because this text is shown even if we're not assigning with subgroups. Also it could be beneficial to clarify when it affects subgroups, I'd suggest something like this:
```suggestion
      "ACTIVE_OFF_HINT": "Goup(s) will be assigned to the resource with INACTIVE status (if assigning with subgroups, subgroups are assigned with ACTIVE status), members of INACTIVE group(s) will not be able to access services until activation",
```

We should also discuss if assigning as inactive option should be visible even if we're not assigning with subgroups, because it no longer makes sense to assign group to resource as INACTIVE on its own I think.

What do you think @Kuliak, @cuadradek ?

cuadradek: Yes, I like @Johaney-s 's text better. And we discussed it and decided we can still show the INACTIVE option if no subgroups are about to be assigned.

Johaney-s (approved):

bodnara:
Please change the scope from `admin-gui` to just `admin`.

bodnara (approved):

PR #942, opened 2021-12-09
fix(profile): adding SSH key fixed

cuadradek:
* added info alert about all supported key formats
* improved validation of new SSH keys


bodnara (changes requested):
image

Right now it is not possible to see all of the possible beginnings at the top (tested on chrome). Also the error message is too clustered in my opinion. I would say show all the beginnings at the top in some user friendly way (maybe show a couple and add a down arrow and on click expand to show all). And in the error message just state that it is incorrect format and refer to the hint above.

Also according to this: 
https://angular.io/guide/component-styles#deprecated-deep--and-ng-deep 
`::ng-deep` is depricated. Please find a work around if the style changes are still needed.

xkureck (approved):

PR #943, opened 2021-12-13
test(admin): added e2e tests for vo management

HejdaJakub:
* E2E tests were created for vo management.
* Selectors were added to all neccesary html components.
* Created new pipe - for dynamically added html tags with multi word selectors.
* As a selector is used data-cy attribute instead of id according to best practises in Cypress documentation (for better isolation from changes).

### DIFF
--- a/apps/admin-gui-e2e/src/integration/vo-management-perun_admin.spec.js
+++ b/apps/admin-gui-e2e/src/integration/vo-management-perun_admin.spec.js
@@ -1,37 +1,291 @@
 /// <reference types="cypress" />
 
 context('Actions', () => {
+  const dbVoName = 'test-e2e-vo-from-db';
+  const dbGroupName = 'test-group-from-db';
+  const dbApplicationItemTextFieldName = 'input-test';
+  const dbVoManager = 'vo-manager-2';
+  const dbUser = 'test';
 
   before(() => {
-    if (Cypress.env('BA_USERNAME')) {
-      sessionStorage.setItem('baPrincipal', '{"name": "test"}');
-      sessionStorage.setItem('basicUsername', Cypress.env('BA_USERNAME'));
-      sessionStorage.setItem('basicPassword', Cypress.env('BA_PASSWORD'));
+    if (Cypress.env('BA_USERNAME_VO_MANAGER')) {
+      sessionStorage.setItem('baPrincipal', '{"name": "voManager"}');
+      sessionStorage.setItem('basicUsername', Cypress.env('BA_USERNAME_VO_MANAGER'));
+      sessionStorage.setItem('basicPassword', Cypress.env('BA_PASSWORD_VO_MANAGER'));
       cy.visit('service-access');
     }
   })
 
   beforeEach(() => {
-    cy.visit('home')
-    cy.intercept('**/vosManager/createVo/**').as('createVo')
+    cy.visit('organizations');
   })
 
   it('test create vo', () => {
 
-    cy.get('#access-item-btn')
+    cy.intercept('**/vosManager/createVo/**').as('createVo')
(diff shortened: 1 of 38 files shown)


apps/admin-gui-e2e/src/integration/vo-management-perun_admin.spec.js
```diff
@@ -12,26 +12,268 @@ context('Actions', () => {
...
         .contains('test-e2e-vo')
         .should('exist')
-  })
+  });
+
+  it('test add attribute and delete', () => {
+
+    cy.intercept('**/attributesManager/setAttributes/vo').as('setAttributes')
+      .intercept('**/attributesManager/removeAttributes/**').as('removeAttributes')
+      .get('[data-cy=test-e2e-vo-from-db]')
```
Vojtech-Sassmann, line 40: I would suggest creating constants with names of objects from db. For example:
```Typescript
const dbVoName = 'test-e2e-vo-from-db';
```
Then you could use it here like this: 
```suggestion
      .get(`[data-cy=${dbVoName}]`)
```
These values should always be defined as contacts. It is then much more clear, where the values are. Also, if you use it in multiple places, you are guaranteed to use the same value.

apps/admin-gui-e2e/src/integration/vo-management-perun_admin.spec.js
```diff
@@ -12,26 +12,268 @@ context('Actions', () => {
...
+        .click()
+      .get('[data-cy=remove-manager-button-dialog]')
+        .click()
+      .wait('@unsetRole')
+      // assert that manager doesn't exist
+      .get('[data-cy=vo-manager-checkbox]')
+        .should('not.exist')
+  });
+
+  it('test delete vo', () => {
```
Vojtech-Sassmann, line 255: This test failed me once because it didn't manage to find the `auto-focused-filter` element. However, I didn't manage to reproduce it. (Therefore, I think it might be a problem with my setup)

Vojtech-Sassmann (changes requested):

apps/admin-gui-e2e/src/integration/vo-management-perun_admin.spec.js
```diff
@@ -1,6 +1,11 @@
...
 
 context('Actions', () => {
+  const dbVoName = 'test-e2e-vo-from-db';
+  const dbGroupName = 'test-group-from-db';
+  const dbApplicationItemTextFieldName = 'input-test';
+  const dbVoManager = 'vo-manager';
+  const dbUser = 'test';
 
   before(() => {
     if (Cypress.env('BA_USERNAME')) {
```
Vojtech-Sassmann, line 11: I have realized one more thing. I think we wanted to do this test with a different principal - a user with only voAdmin role. Could you please implement it? It also requires the user to be created in the DB with the appropriate role.

apps/admin-gui-e2e/src/integration/vo-management-perun_admin.spec.js
```diff
@@ -12,26 +17,268 @@ context('Actions', () => {
...
+      .get('[data-cy=vo-name-link]')
         .contains('test-e2e-vo')
         .should('exist')
-  })
+  });
+
+  it('test add attribute and delete', () => {
+
+    cy.intercept('**/attributesManager/setAttributes/vo').as('setAttributes')
+      .intercept('**/attributesManager/removeAttributes/**').as('removeAttributes')
```
Vojtech-Sassmann, line 44: I think the delete functionality is tested below.

Vojtech-Sassmann (changes requested):

xkostka2 (approved):

PR #944, opened 2021-12-14
feat(profile): change section Authentication in User profile

xkostka2:
* management of MFA has been removed from the User profile, now there is only link to page where users can manage their MFA settings


apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
```diff
@@ -165,65 +152,6 @@ export class SettingsAuthenticationComponent implements OnInit, AfterViewInit {
...
-            this.dataSource.paginator = this.children.first.paginator;
-            this.loading = false;
-          });
-        }
-      }
-    }, () => this.loading = false);
-  }
 }
 
 export interface MFAToken {
```
bodnara, line 157: I believe this is not needed anymore along with the datasource. Otherwise it looks ok.

bodnara (approved):

PR #945, opened 2021-12-23
fix(admin): when user removes himself as manager the authorization is…

xkureck:
… checked

* when user removes himself as manager of entity the authorization
priviledges are checked
* if user dont have the priviledge to stay on
page, he is redirected to page that he have priviledge to see

### DIFF
--- a/apps/admin-gui/src/app/core/services/common/reload-entity-detail.service.ts
+++ b/apps/admin-gui/src/app/core/services/common/reload-entity-detail.service.ts
@@ -1,0 +1,49 @@
+import { EventEmitter, Injectable, Output } from '@angular/core';
+import { EntityStorageService } from '@perun-web-apps/perun/services';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class ReloadEntityDetailService {
+
+  constructor(private entityStorageService: EntityStorageService,) { }
+
+  @Output() groupDetailChange: EventEmitter<void> = new EventEmitter();
+  @Output() voDetailChange: EventEmitter<void> = new EventEmitter();
+  @Output() resourceDetailChange: EventEmitter<void> = new EventEmitter();
+  @Output() facilityDetailChange: EventEmitter<void> = new EventEmitter();
+
+  reloadGroupDetail(): void {
+    this.groupDetailChange.emit();
+  }
+
+  reloadFacilityDetail(): void {
+    this.facilityDetailChange.emit();
+  }
+
+  reloadResourceDetail(): void {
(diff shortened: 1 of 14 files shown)


bodnara (changes requested):
Couple of things regarding the redirection:
-  I believe since the `get*byId...` policies only require the role `SELF` you never get redirected to home page (at least I didn't manage to)
- redirecting to Vo detail form group managers seems unnecessary
- using absolute instead of relative paths would be a better solution even if just for the sake of code readability
- there is a lot of code duplication in the `redirectToAuthRoute` method in `managers-page`, so please refactor this 
- giving users some information as to why they are being redirected would be nice, something like "You removed yourself as a manager and therfore lost access to this *blank* ..."

So if I understand correctly the `reload-entity-service` is needed to force the entity detail to reload itself but only the sidemenu items are incorrect ? Also maybe one emitter would be enough since you only have one open detail at a time.

xkureck (commented):
1. yes, thats true, but once the membership role will be used the using of the policy will make sense finally
2. eg. the user can have role group manager to the other group in the vo, so the user still have some connection to the vo so that's why I redirected him to the vo
maybe the naive solution (when user removes himself as manager just navigate him to the homepage) would be the best?
3. I you think that its more readable then ok, but in the resource case I cannot know if the path is from vo or facility, so I will keep it like that in that case
4. refactored
5. are users that stupid? and if they are, will the notification help?
6. the sidemenu and the entity detail can be wrong too (eg. hide the edit/delete button) 
and yeah, one emmiter would be enough I guess

xkureck (commented):
in the review we agreed that we would like to inform users in the "remove manager" dialog when users is removing himself as some manager

bodnara (commented):
Regarding the redirection to org. detail when removing group manager. The question is if we want consistency (redirect to home) or we want it to be helpful and redirect to org detail which the user may or may not find usefull. Consistent aproach would also make the logistics of redirecting somewhat easier.

I would still appreciate if the info about redirecting would be somehow communicated but leave it up to you.

xkureck:
> Regarding the redirection to org. detail when removing group manager. The question is if we want consistency (redirect to home) or we want it to be helpful and redirect to org detail which the user may or may not find usefull. Consistent aproach would also make the logistics of redirecting somewhat easier.
> 
> I would still appreciate if the info about redirecting would be somehow communicated but leave it up to you.

In the last review, I tried to explain your opinion on these things and stakeholders said it is fine like it is.
But please when we will talk about this task next review please bring this worries up.

bodnara (approved):

PR #949, opened 2022-01-04
feat: advanced filter for assigned users

HejdaJakub:
* Advanced filter according to prototype was added to section assigned users.
* Part of this new component is resuable, so it was placed to libs directory.

### DIFF
--- a/libs/perun/components/src/lib/advanced-filter/advanced-filter.component.ts
+++ b/libs/perun/components/src/lib/advanced-filter/advanced-filter.component.ts
@@ -1,0 +1,31 @@
+import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
+
+@Component({
+  selector: 'perun-web-apps-advanced-filter',
+  templateUrl: './advanced-filter.component.html',
+  styleUrls: ['./advanced-filter.component.scss']
+})
+export class AdvancedFilterComponent implements OnInit {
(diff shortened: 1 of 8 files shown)


bodnara (approved):
Everything works as intended great job.

One thing we could discuss is if it would be helpful/desirable to hide the advanced filter when its not focused anymore. The use case being user sets some parameter in the filter and just wants to look at the data so he click anywhere outside the filter instead of just the button.
Once again this is up to preference it could be both helpful and annoying.

libs/perun/components/src/lib/advanced-filter/advanced-filter.component.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  @Output()
+  clearFilters: EventEmitter<any> = new EventEmitter<any>();
+
+  ngOnInit(): void {
+    this.changeAdvancedFilter.emit(this.advancedFilter);
+  }
+
+  hideAndShowAdvancedFilter() {
+    this.advancedFilter = !this.advancedFilter;
```
bodnara, line 28: Maybe `toggleAdvacedFilter` could be a better description of what this method does.

HejdaJakub: Renamed.

PR #953, opened 2022-01-19
feat(admin): dialog for adding hosts clean-up

xflord:
* removed the name of the facility from the dialog title


xkostka2 (changes requested):
I would suggest also removing unused variable facilityName (also from AddHostDialogData) and making FacilityHostsComponent not pass this value to the dialog.

xkostka2 (approved):

PR #954, opened 2022-01-21
feat(admin): improve list of assigned groups to resource

HejdaJakub:
* Each group is displayed in this table only once now.
* If some group is assigned manually and also automatically, it will have a special icon with info in the tooltip. This group will be also able to remove (then will stay only automatic assignment).
* If some group is assigned only automatically, now will be possible to assign this group also manually in the assign-group-to-resource dialog.

### DIFF
--- a/apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-groups/resource-groups.component.ts
+++ b/apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-groups/resource-groups.component.ts
@@ -55,7 +55,24 @@ export class ResourceGroupsComponent implements OnInit {
         gws.sourceGroupId = g.sourceGroupId;
         return gws;
       });
-      this.groupsToDisable = new Set(this.assignedGroups.filter(group => !!group.sourceGroupId).map(group => group.id));
+
+      // identify groups with multiple assignments
+      const groupsIds = this.assignedGroups.map(group => group.id);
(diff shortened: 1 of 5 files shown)


apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-groups/resource-groups.component.ts
```diff
@@ -55,7 +55,22 @@ export class ResourceGroupsComponent implements OnInit {
         gws.sourceGroupId = g.sourceGroupId;
         return gws;
       });
-      this.groupsToDisable = new Set(this.assignedGroups.filter(group => !!group.sourceGroupId).map(group => group.id));
+
+      // identify groups with multiple assignments
+      const groupsIds = this.assignedGroups.map(group => group.id);
+      const groupsWithMultiAssignments = this.assignedGroups.filter((group, i) =>
+        groupsIds.indexOf(group.id) !== i && (group.sourceGroupId === null || this.assignedGroups[groupsIds.indexOf(group.id)].sourceGroupId === null)).map(group => group.id);
+      const uniqueGroups = this.assignedGroups.filter((group, i) => groupsIds.indexOf(group.id) === i);
```
bodnara, line 63: Please, put the main `.filter` action in a block and format it a little so it's not just a super long line. Also don't calculate `groupsIds.indexOf(group.id)` twice.

bodnara (changes requested):

bodnara (approved):

PR #956, opened 2022-01-21
fix(admin/publications) make refreshing not delete filter value

xflord:
* fixed refresh button removing filter in 'facility - allowed groups'
* fixed refresh button removing filter in 'VO - member - groups'
* fixed refresh button removing filter in 'VO - member - resources'
* fixed refresh button removing filter in 'all publications' and 'my publications'


bodnara (commented):
Well done. One technical thing, usually when multiple scopes are involved such as in this case we ommit the scope specification. Please change the commit message to `fix: make refreshing...`

bodnara (approved):

PR #959, opened 2022-01-31
feat: don't force sign-in on new tab

bodnara:
* authentication flow tries to refresh access token so user doesn't need to sign-in again


xkostka2 (changes requested):
I think you forgot to handle logging out.

xkostka2 (approved):

PR #963, opened 2022-02-07
feat(admin): paginated applications

Johaney-s:
* new component for dynamic table of applications for member, vo and group
* generated latest openApi specification


bodnara (changes requested):
- On vo applications don't display group name column unless group applications are included.
- Sorting by type returns incorrect order, but I assume this is a backend problem.
- Please create one `spinner-container` css class for all dynamic tables to reduce code duplication.
- Please move translations for table header to separate part, it doesn't really belong under `VO_DETAIL.APPLICATION`
- Would it be possible to only define one dynamic table element per page and only change the input values?

Johaney-s:
>     * On vo applications don't display group name column unless group applications are included.
> 
>     * Sorting by type returns incorrect order, but I assume this is a backend problem.
> 
>     * Please create one `spinner-container` css class for all dynamic tables to reduce code duplication.
> 
>     * Please move translations for table header to separate part, it doesn't really belong under `VO_DETAIL.APPLICATION`
> 
>     * Would it be possible to only define one dynamic table element per page and only change the input values?

Thanks, group name and id should be now visible only if checkbox is selected. Sorting by type was indeed a problem on backend side, it will also use userId instead of memberId for member's applications, so I changed it here too. Anyway, if there are no more problems, it should be probably merged after backend PR https://github.com/CESNET/perun/pull/3464 is merged.

bodnara (changes requested):
Just one small thing, please for the translation use something more specific like: `APPLICATION_LIST`.

bodnara (approved):

xkureck (approved):

PR #966, opened 2022-02-09
feat: add shadow to header of the user profile, publications and pass…

xkureck:
…word reset

* add shadow to header of the user profile, publications and password
reset


bodnara (commented):
I assume this was already present in admin-gui. So it would make sense to create one component for header and reuse it in all apps in order to prevent code duplication as well as the need to do the same change for each app.

xkureck:
> I assume this was already present in admin-gui. So it would make sense to create one component for header and reuse it in all apps in order to prevent code duplication as well as the need to do the same change for each app.

I guess in long term this would be better, but this is a quick fix, which Peto would like to have on production
Will you make task in Jira?

PR #983, opened 2022-02-22
feat(admin): Add activate guard

bodnara:
* created routing guard
* added guard for organizations section for some common scenarios
* rest off the section as well as pages need to be added

### DIFF
--- a/apps/admin-gui/src/app/shared/route-auth-guard.service.ts
+++ b/apps/admin-gui/src/app/shared/route-auth-guard.service.ts
@@ -32,14 +32,16 @@ export class RouteAuthGuardService implements CanActivateChild {
 
   private static getBeanName(key: string): string {
     switch (key) {
-      case 'o':
+      case 'organizations':
         return 'Vo';
-      case 'g':
+      case 'groups':
         return 'Group';
-      case 'f':
+      case 'facilities':
         return 'Facility';
-      case 'r':
+      case 'resources':
         return 'Resource';
+      case 'members':
+        return 'Member';
       default:
         return '';
     }
@@ -62,7 +64,7 @@ export class RouteAuthGuardService implements CanActivateChild {
     }
 
     authPair.key = authPair.key.slice(0, authPair.key.length - 1);
-    authPair.entity.beanName = RouteAuthGuardService.getBeanName(authPair.key[0]);
+    authPair.entity.beanName = RouteAuthGuardService.getBeanName(authPair.key.split('-')[0]);
     return authPair;
   }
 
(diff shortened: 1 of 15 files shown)


xkureck (changes requested):
also there needs to be implemented a function which we would call also from components like
vo-overview and sidemenu - from those components we cannot provide ActivatedRouteSnapshot and RouterStateSnapshot

apps/admin-gui/src/app/shared/route-auth-guard.service.ts
```diff
@@ -0,0 +1,96 @@
...
+        break;
+      }
+
+      authPair.key = segment.concat('-', authPair.key);
+    }
+
+    authPair.key = authPair.key.slice(0, authPair.key.length - 1);
+    authPair.entity.beanName = RouteAuthGuardService.getBeanName(authPair.key[0]);
+    return authPair;
+  }
```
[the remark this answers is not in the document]
bodnara: I couldn't find a solution that doesn't feel "hack-y". Parsing the bean name from URL seems like a consistent approach. At least to me.

bodnara:
> also there needs to be implemented a function which we would call also from components like vo-overview and sidemenu - from those components we cannot provide ActivatedRouteSnapshot and RouterStateSnapshot

The logic changed a bit but the solution for side-menu and similar cases stays the same. Just use the `canNavigate` method of  `RoutePolicyService`. Since both the key and entity are known at that point.

apps/admin-gui/src/app/shared/route-auth-guard.service.ts
```diff
@@ -0,0 +1,90 @@
...
+  private static getBeanName(key: string): string {
+    switch (key) {
+      case 'o':
+        return 'Vo';
+      case 'g':
+        return 'Group';
+      case 'f':
+        return 'Facility';
+      case 'r':
+        return 'Resource';
```
xkureck, line 42: ok, I guess this is the best way
but wouldn't be better in that case that there would be whole names of the path, like "organizations" instead of "o"
its better code readability and also when we one day change the URL path in can be easier to identify which pieces of code needs to be changed as well
btw members path is missing here

libs/perun/services/src/lib/route-policy.service.ts
```diff
@@ -0,0 +1,88 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class RoutePolicyService {
+  constructor(private authResolver: GuiAuthResolver) {}
+
+  // Map of page key and function determining if user is authorized to access given page
+  private routePolicies: Map<string, (entity: Entity) => boolean> = new Map<
+    string,
+    (entity: Entity) => boolean
```
xkureck, line 16: you forget about statistics policy

xkureck (changes requested):
You can move the not authorized page to some /lib directory, so we can reuse it in other application?

xkureck (approved):

xkostka2 (approved):

PR #986, opened 2022-02-28
fix(pwd-reset): resolve eslint errors

Johaney-s:
* resolved linter errors in password/reset and password-reset-e2e directories


apps/password-reset/src/main.ts
```diff
@@ -10,4 +10,5 @@ if (environment.production) {
 
 platformBrowserDynamic()
   .bootstrapModule(AppModule)
+  // eslint-disable-next-line no-console
```
bodnara, line 13: This comment can be removed.

PR #988, opened 2022-02-28
Fix eslint in libs

bodnara:
Fixed ESlint errors in all libraries "projects"


libs/perun/utils/src/lib/perun-utils.ts
```diff
@@ -14,6 +14,10 @@ import {
   ApplicationFormItem,
   RichGroup,
   Author,
+  Facility,
+  Vo,
+  Resource,
+  PerunBean,
```
HejdaJakub, line 20: Unused import.
```suggestion
```

libs/perun/utils/src/lib/perun-utils.ts
```diff
@@ -271,13 +283,22 @@ export function addRecentlyVisited(key: string, item: any) {
...
+}
+
 /**
  * Add object that was just visited to 'recent' localStorage.
  *
  * @param item entity that was visited
  * @param voName is used for group tooltip on dashboard (we want to know parent VO of this group)
  */
-export function addRecentlyVisitedObject(item: any, voName?: string) {
+export function addRecentlyVisitedObject(item: Vo & Facility & Group, voName?: string): void {
```
HejdaJakub, line 301:
```suggestion
export function addRecentlyVisitedObject(item: Vo | Facility | Group, voName?: string): void {
```

HejdaJakub (changes requested):
Reviewed everything except `components`, `pipes` and `services` (147 of 264 files).

libs/perun/services/src/lib/sponsored-members-pdf.service.ts
```diff
@@ -1,53 +1,84 @@
...
+
+interface MemberDataTable {
+  heights: number[];
+  widths: number[];
+  body: MemberDataCell[][];
+}
+
+export interface MemberDataTableWrapper {
+  table: MemberDataTable;
+}
```
HejdaJakub, line 31: Do you need this interface if it is the same as `MemberDataTable` just only hidden under another attribute?

libs/perun/services/src/lib/gui-auth-resolver.service.ts
```diff
@@ -1,19 +1,30 @@
...
 } from '@perun-web-apps/perun/openapi';
 import { Role } from '@perun-web-apps/perun/models';
 
+interface Privilege {
+  readAuth: boolean;
+  manageAuth: boolean;
+  modes: string[];
+}
+
+type Entity = Vo & Group & Resource & Member & User;
```
HejdaJakub, line 22: I am not sure if there should be ampersands.
```suggestion
type Entity = Vo | Group | Resource | Member | User;
```

libs/perun/components/src/lib/selection-item-search-select/selection-item-search-select.component.ts
```diff
@@ -153,7 +144,7 @@ export class SelectionItemSearchSelectComponent implements OnInit {
     this.item.value = this.selectedAttribute;
   }
 
-  sortAttributes(attributes: SelectionItem[]) {
+  sortAttributes(attributes: SelectionItem[]): void {
     attributes.sort((a: SelectionItem, b: SelectionItem) => {
```
HejdaJakub, line 148: I think there should be also return type.
```suggestion
  sortAttributes(attributes: SelectionItem[]): void {
    attributes.sort((a: SelectionItem, b: SelectionItem): number => {
```

HejdaJakub (commented):
I realized that the transform methods in `application-form-item-type.pipe.ts` and `application-form-widget-description.pipe.ts` can be typed (parameter and return type - probably `string | Type` will be suitable).

**NOTE:** New application types will be added in PR https://github.com/CESNET/perun-web-apps/pull/995 (it means two new cases in switch statement).

libs/perun/utils/src/lib/perun-utils.ts
bodnara: Here we want the union `&` of types.

libs/perun/services/src/lib/auth.service.ts
```diff
@@ -206,7 +256,7 @@ export class AuthService {
         .loadDiscoveryDocument()
         .then(() => this.oauthService.refreshToken())
         .then(() => Promise.resolve())
-        .catch((err) => err);
+        .catch(() => Promise.resolve());
```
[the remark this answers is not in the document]
bodnara: The error will be handled in `ApiInterceptor` and since I believe we don't react to it in any specific way here, there is no need to return in. Instead empty `void` promise is returned.

libs/perun/services/src/lib/notificator.service.ts
```diff
@@ -49,7 +45,7 @@ export class NotificatorService {
       this.showError(
         this.translate.instant(
           'SHARED_LIB.PERUN.COMPONENTS.NOTIFICATOR.NOTIFICATION.PRIVILEGE_EXCEPTION'
-        ),
+        ) as string,
```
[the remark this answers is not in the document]
bodnara: Couldn't find this method.

libs/perun/services/src/lib/gui-auth-resolver.service.ts
bodnara: Again here we need union of types, since there are properties such as `userId` and etc. that are not present in all of the types.

bodnara:
> I realized that the transform methods in `application-form-item-type.pipe.ts` and `application-form-widget-description.pipe.ts` can be typed (parameter and return type - probably `string | Type` will be suitable).

Since the switch is exhaustive these pipes will always return a `string`.

apps/admin-gui/src/app/shared/pipes/application-form-item-type.pipe.ts
```diff
@@ -9,7 +10,7 @@ export class ApplicationFormItemTypePipe implements PipeTransform {
 
   constructor(private translateService: TranslateService) {}
 
-  transform(value: any): any {
+  transform(value: Type): string {
```
HejdaJakub, line 13: This pipe throws an error "Argument type string is not assignable to parameter type Type" in `add-application-form-item-dialog.component.html`.

apps/admin-gui/src/app/shared/pipes/application-form-widget-description.pipe.ts
```diff
@@ -9,7 +10,7 @@ export class ApplicationFormWidgetDescriptionPipe implements PipeTransform {
 
   constructor(private translateService: TranslateService) {}
 
-  transform(value: any): any {
+  transform(value: Type): string {
```
HejdaJakub, line 13: Again. This pipe throws an error "Argument type string is not assignable to parameter type Type" in `add-application-form-item-dialog.component.html`.

HejdaJakub (commented):
Don't forget to add type `as string` to the translation in `showRouteError()` in `notificator.service.ts` after rebasing.

HejdaJakub (approved):

PR #990, opened 2022-03-02
feat(admin): display form data in dynamic list component

mattjoke:
* added form data column in detailed view
* added button to show all form data
* generated latest openApi specification


apps/admin-gui/src/app/vos/components/applications-dynamic-list/applications-dynamic-list.component.html
```diff
@@ -108,6 +108,29 @@
...
+              {{getFormDataString(data)}}
+            </ng-container>
+          </ng-container>
+          <ng-template #elseBlock>
+            <ng-container *ngFor="let data of application.formData.slice(0,5)">
+              {{getFormDataString(data).substring(0,70)}}...
+            </ng-container>
+          </ng-template>
+          <button
+            mat-raised-button
```
xkureck, line 125:
```suggestion
            mat-flat-button
```

xkureck (changes requested):
1. setup prettier for your IDE(the check on you PR have some issues with prettier) - I have send some instruction on how to do in on our slack channel and repair the problems that the prettier checked
-> with that I think there will be less files with changes that needs to be reviewed
2. 
image
missing translation for the table header
also in the table I think we would like some user friendly data presented instead of whole json object, something like mail: tydyt@gmail.com for one row would be nice I think

xkureck (changes requested):
image
there is really weird bug that when you check the group applications and then click on 'show more details" then this happens

PR #995, opened 2022-03-16
feat(admin): support list and map application form item

bodnara:
* allow creating list and map form items
* properly show them on preview
* allow editing of these items


mattjoke (approved):

apps/admin-gui/src/app/vos/components/application-form-preview/application-form-preview.component.html
```diff
@@ -430,6 +430,60 @@ <h1 class="page-subtitle">
...
+              class="pointer"
+              *ngIf="applicationFormItem.hidden !== 'NEVER'"
+              [matTooltip]="hiddenTooltip(applicationFormItem)"
+              >visibility_off</mat-icon
+            >
+            <span class="ml-2">{{getLocalizedHint(applicationFormItem)}}</span>
+          </div>
+        </div>
+        <div
+          *ngIf="applicationFormItem.type === 'MAP_INPUT_LIST' &&
```
HejdaJakub, line 461: Map item isn't visible on the preview.
```suggestion
          *ngIf="applicationFormItem.type === 'MAP_INPUT_BOX' &&
```

HejdaJakub (changes requested):
I think you should also add these two types to the property `typesWithDisabled` in the `edit-application-form-item-dialog.component.ts`. There are also properties like `typesWithUpdatable` and `dependencyTypes`, but I am not sure if we want the list and map to be updatable or to be a dependency type, but the option for disable them there probably should be.

apps/admin-gui/src/assets/img/PerunWebImages/home.svg
```diff
@@ -1,10 +1,10 @@
...
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 23.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Vrstva_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+     viewBox="0 0 128 128" style="enable-background:new 0 0 128 128;" xml:space="preserve">
+  <style type="text/css">
+    .st0{fill:#222222;}
+  </style>
+  <path class="st0" d="M99.5,104.2l0-41.2l6.3,6.3l7-7L64.1,13.7L64,13.8l-0.1-0.1L15.2,62.3l7,7l6.3-6.3l0,41.2H0V115H128v-10.8H99.5
+	z M64.9,104.2V75.3H49v28.9l-10.5,0l0-51.1L64,27.6l25.5,25.5l0,51.1L64.9,104.2z"/>
+</svg>
```
HejdaJakub, line 10: This probably shouldn't be a part of this PR? Otherwise, it looks good.

HejdaJakub (approved):

PR #999, opened 2022-03-31
fix(admin): Fixed status tab in Accounts

mattjoke:
* fixed status tab error in Perun Admin Accounts


xkureck (approved):
so this is valid fix - at least the page is somehow working
but the problem is not fixed - and that the backend stopped to send the group status even when the frontend requested it

PR #1006, opened 2022-04-05
fix: minor improvements

xkureck:
* page admin owners have correct theme set
* page admin users have autofocus set on input
* added shadow to header in the login screen


bodnara (commented):
The `AddOwnerDialog` used on the admin-owners page has wrong theme as well.

PR #1007, opened 2022-04-06
fix: redirect from login page to home if user logged in

xflord:
* when user is logged in and goes to /login, the login component isn't displayed and the user is redirected to /home instead


bodnara (changes requested):
The solution works, however instead of creating a `<div>` in html to handle the logic make the `LoginScreenComponent` implement `OnInit` and put the logic in there.
Also run the `format:write` script.

bodnara (approved):

PR #1014, opened 2022-04-13
fix: fix refresh token

xkureck:
* fix validity of access token
* fix storing the refresh token right after the login in

### DIFF
--- a/libs/perun/services/src/lib/auth.service.ts
+++ b/libs/perun/services/src/lib/auth.service.ts
@@ -37,6 +37,10 @@ export class AuthService {
 
   getClientConfig(): AuthConfig {
     const filterValue = this.setIdpFilter();
+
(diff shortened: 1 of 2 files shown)


libs/perun/services/src/lib/auth.service.ts
```diff
@@ -291,4 +292,16 @@ export class AuthService {
...
+
+  private refreshAndStoreToken(): Promise<boolean> {
+    return this.oauthService.refreshToken().then(
+      (response) => {
+        localStorage.setItem('refresh_token', response['refresh_token']);
+        return true;
+      },
+      () => {
+        return false;
+      }
```
HejdaJakub, line 304: I think it looks good. There is just one eslint error.
```suggestion
      () => false
```

HejdaJakub (approved):

PR #1018, opened 2022-04-26
feat(admin): add Searcher to the admin section

HejdaJakub:
* Added Searcher page with four tabs (users, members, facilities, resources).
* To members-list and resources-list were added new columns according to the old GUI.
* At this moment, we get the Member object from BE, but member status tooltip and member expiation settings require RichMember, so some other components have to be changed.

### DIFF
--- a/apps/admin-gui/src/app/admin/pages/admin-page/admin-searcher/admin-searcher.component.html
+++ b/apps/admin-gui/src/app/admin/pages/admin-page/admin-searcher/admin-searcher.component.html
@@ -1,0 +1,90 @@
+<h1 class="page-subtitle">{{'ADMIN.SEARCHER.TITLE' | translate}}</h1>
+<mat-tab-group (selectedIndexChange)="tabChanged($event)">
+  <mat-tab class="mb-2">
+    <ng-template matTabLabel>
+      {{'ADMIN.SEARCHER.TAB_USERS' | translate}}
+    </ng-template>
+    <ng-template matTabContent>
+      <perun-web-apps-attribute-search-select
+        *ngIf="!loading"
+        [attributes]="allAttrDefinitions"
+        [attributesForEntity]="'user'"
+        (search)="searchEntities($event, 'user')">
+      </perun-web-apps-attribute-search-select>
+      <app-users-list
+        *ngIf="!loadingEntityData"
+        [tableId]="tableId"
+        [displayedColumns]="['user', 'id', 'name']"
+        [users]="entities">
+      </app-users-list>
+    </ng-template>
+  </mat-tab>
+  <mat-tab>
+    <ng-template matTabLabel>
+      {{'ADMIN.SEARCHER.TAB_MEMBERS' | translate}}
+    </ng-template>
+    <ng-template matTabContent>
+      <div class="mt-3">
+        <perun-web-apps-vo-search-select
+          *ngIf="vos.length > 0"
+          [vos]="vos"
+          (voSelected)="voSelected($event)">
+        </perun-web-apps-vo-search-select>
+      </div>
+      <!-- we search members in searcher by user attributes !!! -->
+      <perun-web-apps-attribute-search-select
+        *ngIf="!loading"
+        [attributes]="allAttrDefinitions"
+        [attributesForEntity]="'user'"
+        (search)="searchEntities($event, 'member')">
+      </perun-web-apps-attribute-search-select>
+      <perun-web-apps-members-list
+        *ngIf="!loadingEntityData"
+        [tableId]="tableId"
+        [displayedColumns]="['id', 'voId', 'userId', 'status']"
+        [disableExpirationChange]="true"
+        (updateTable)="searchEntities(searchInput, 'member')"
+        [members]="entities">
+      </perun-web-apps-members-list>
+    </ng-template>
+  </mat-tab>
+  <mat-tab>
+    <ng-template matTabLabel>
+      {{'ADMIN.SEARCHER.TAB_FACILITIES' | translate}}
+    </ng-template>
+    <ng-template matTabContent>
+      <perun-web-apps-attribute-search-select
+        *ngIf="!loading"
+        [attributes]="allAttrDefinitions"
+        [attributesForEntity]="'facility'"
+        (search)="searchEntities($event, 'facility')">
+      </perun-web-apps-attribute-search-select>
+      <perun-web-apps-facilities-list
+        *ngIf="!loadingEntityData"
+        [tableId]="tableId"
+        [displayedColumns]="['id', 'name', 'description']"
+        [facilities]="entities">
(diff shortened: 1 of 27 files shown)


apps/admin-gui/src/app/admin/pages/admin-page/admin-searcher/admin-searcher.component.html
```diff
@@ -0,0 +1,86 @@
...
+    </ng-template>
+  </mat-tab>
+  <mat-tab>
+    <ng-template matTabLabel>
+      {{'ADMIN.SEARCHER.TAB_MEMBERS' | translate}}
+    </ng-template>
+    <ng-template matTabContent>
+      <div class="mt-3">
+        <perun-web-apps-vo-search-select
+          *ngIf="!loadingEntityData"
```
xkureck, line 28: its kind of weird for user when some member is searched and then this select also disappears

apps/admin-gui/src/app/admin/pages/admin-page/admin-searcher/admin-searcher.component.ts
```diff
@@ -0,0 +1,164 @@
...
+  tabChanged(indexChangeEvent: number): void {
+    this.entities = [];
+    switch (indexChangeEvent) {
+      case 0: // Users
+        this.filterAttrDefinitions('user');
+        this.tableId = TABLE_SEARCHER_USERS;
+        break;
+      case 1: // Members
+        this.loadingEntityData = true;
+        this.filterAttrDefinitions('user'); // we search members in searcher by user attributes !!!
```
xkureck, line 67: and what about member attributes?

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -2135,6 +2136,8 @@
   },
   "MEMBERS_LIST": {
     "ID": "Id",
+    "VO_ID": "Vo Id",
```
xkureck, line 2139:
```suggestion
    "VO_ID": "Organization Id",
```

apps/admin-gui/src/app/admin/pages/admin-page/admin-searcher/admin-searcher.component.ts
```diff
@@ -0,0 +1,164 @@
...
+      this.allAttrDefinitions = attrDefs;
+      this.filterAttrDefinitions('user');
+      this.loading = false;
+    });
+  }
+
+  filterAttrDefinitions(entity: string): void {
+    this.availableAttrDefs = this.allAttrDefinitions
+      .filter((attrDef) => attrDef.entity === entity)
+      .sort(compareFnDisplayName);
```
xkureck, line 55: Maybe this could be part of the attribute search select component

apps/admin-gui/src/app/admin/pages/admin-page/admin-searcher/admin-searcher.component.ts
```diff
@@ -0,0 +1,164 @@
...
+        break;
+      default:
+        break;
+    }
+  }
+
+  loadAllVos(): void {
+    this.loadingEntityData = true;
+    this.voService.getAllVos().subscribe((vos) => {
+      this.vos = vos.sort(compareFnName);
```
xkureck, line 87: question to think about - don't we want to be the organizations sorted like that by default in the vo-search-select-component ?

apps/admin-gui/src/app/admin/pages/admin-page/admin-searcher/admin-searcher.component.html
```diff
@@ -0,0 +1,86 @@
...
+      </div>
+      <perun-web-apps-attribute-search-select
+        *ngIf="!loading"
+        [attributes]="availableAttrDefs"
+        (search)="searchEntities($event, 'member')">
+      </perun-web-apps-attribute-search-select>
+      <perun-web-apps-members-list
+        *ngIf="!loadingEntityData"
+        [tableId]="tableId"
+        [displayedColumns]="['id', 'voId', 'userId', 'status']"
```
xkureck, line 42: add name column :)

xkureck (changes requested):
image
On our add something button in new GUI we don't have any icon

Also to the searcher page add chache so users can come back to the results
`@HostBinding('class.router-component') true;`

apps/admin-gui/src/app/admin/pages/admin-page/admin-searcher/admin-searcher.component.ts
HejdaJakub: The old GUI is implemented in the same way and also javadocs on the backend mention, that members should be searched by user attributes in the Searcher. Moreover, the name of the function is `getMembersByUserAttributes`, so I think, that it's a feature.

apps/admin-gui/src/app/admin/pages/admin-page/admin-searcher/admin-searcher.component.ts
HejdaJakub: This functionality has been moved to the attribute-search-select component.

apps/admin-gui/src/app/admin/pages/admin-page/admin-searcher/admin-searcher.component.ts
HejdaJakub: Good point. FYI in the `notifications-copy-mails-dialog` and `application-form-copy-items-dialog` components, some redundant sort has been removed and in the `user-accounts` has been deleted default selected vo, so now the default vo is managed by `vo-search-select-component` for this component.

apps/admin-gui/src/app/admin/pages/admin-page/admin-searcher/admin-searcher.component.html
HejdaJakub: As I mentioned earlier, we have just a `Member` object instead of a `RichMember` object at this moment, so we can't display the user name for these members. But in the old GUI, there isn't a name column too, so it shouldn't be a problem.
In the future, we can extend the methods for Rich/Enriched entities and display more columns.

xkureck (changes requested):
image
image
some errors in console when you changed behaviour of vo-search-select

bodnara (commented):
There should be a limit to how many "parameters" user can add. Number of attributes avaliable would make sense.
Not passing already used attributes when adding parameter could be useful. However, this would also require adding attribute to all existing search options when removed. So perhaps too much work for a small benefit.

HejdaJakub:
Errors in vo-search-select have been fixed.

The minimum count for available attributes is 69 (for resources), so it maybe doesn't make sense to set a limit with a large number like this. What do you think @bodnara? Is there any benefit to doing it?

xkureck (approved):

PR #1024, opened 2022-05-06
feat(user-profile): downsize antiphishing images

melanger:
shrink image down to 100x100 and try even smaller sizes up to 50x50 until the desired size is reached

Requires more work and testing.


xkureck (changes requested):
image
Be consistent on how you name this functionality - on page title its security image, on dialog title its anti-phishing image

image
same with remove dialog

xkureck (changes requested):
Please just add notifications when the image is saved and removed please and we can merge it

xkureck (approved):

PR #1028, opened 2022-05-10
feat(admin): generate sponsored members - note in CSV output

bodnara:
* extended CSV output when generating sponsored members
* note, if present, is included in CSV output


bodnara:
Should be tested after https://github.com/CESNET/perun/pull/3577 is merged.

HejdaJakub (approved):

xkureck (approved):

PR #1029, opened 2022-05-11
feat(profile): accessibility features

bodnara:
* application title is now also translated
* html lang attribute is set based on user preferred language
* added aria label to menu toggle button
* added title to header logo for screen readers

BREAKING CHANGE: Config variable document_title, was changed from simple string to object containing title for each supported language e.g. `documnet_title: {en: "Example title"}`


mattjoke (approved):

apps/user-profile/src/assets/i18n/cs.json
```diff
@@ -1,4 +1,5 @@
 {
+  "TITLE": "Uživatelskí profil",
```
xkureck, line 2:
```suggestion
  "TITLE": "Uživatelský profil",
```

xkureck (changes requested):

xkureck (approved):

PR #1040, opened 2022-05-19
feat(admin): add member result handling

bodnara:
* chnaged notification behaviour for adding members
* if all request end in success notification is shown
* if some request result in an error a list of candidates and the corespondig error are shown
* separated the container and the presenter part of add member dialog
* refactored files affected by this change


apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1105,20 +1105,17 @@
       "DESCRIPTION": "Search user by name, login or email",
       "SEARCH": "Search",
       "CANCEL": "Cancel",
+      "CLOSE": "Close",
       "CREATE": "Add",
       "INVITE": "Invite",
       "INVITE_IN_LANGUAGE": "Send invitation in",
       "LANGUAGE": "language",
       "SUCCESS": "Selected members were added",
```
xflord, line 1113:
```suggestion
      "SUCCESS_ADD": "Selected members were added",
```
to match usage

xflord (changes requested):
A quick review since I just wanted to see how it all turned out, might look at it more Monday. Also, when searching for users I had some problems with the auto-search, where when I was searching different users in quick succession, my latest result would be overwritten after like a second by a search i did previously. However this was quite rare (Happened 2 times after like 30 minutes of trying to reproduce)

xkureck (changes requested):
image
success notification is not working

image
I just miss something in this dialog - I think this needs some sort of alert - alert about what I should do now

image
For this case maybe would be usable the error dialog.
Also the table should look somehow like a table like we have it in the other parts of the GUI, like this:
image
first column "name" and second "error"
Also I wouldn't put the name of the exception(eg. WrongReferenceAttributeValueException) like normal text - for user its really confusing
But I would put the name of the exception under the error message like this 
image

Some of this may be unclear, or feels no right, so lets talk about it on daily

bodnara:
>Also, when searching for users I had some problems with the auto-search, where when I was searching different users in quick succession, my latest result would be overwritten after like a second by a search i did previously. However this was quite rare (Happened 2 times after like 30 minutes of trying to reproduce)

There was indeed a problem, that some previous request may take too long and override the data. Now each new request should cancel the previous one if it is not finished.

xflord (approved):

xkureck (changes requested):
this is not whole review, just some visuals, I will continue with code, tomorrow

apps/admin-gui/src/app/shared/components/members-candidates-list/members-candidates-list.component.html
```diff
@@ -25,14 +25,14 @@
         <td *matCellDef="let memberCandidate" class="static-column-size align-checkbox" mat-cell>
           <div
             [matTooltip]="getTooltip(memberCandidate) | translate"
-            [matTooltipDisabled]="!isCheckboxDisabled(memberCandidate)"
+            [matTooltipDisabled]="!(memberCandidate | disabledCandidate: blockManualAdding)"
```
xkureck, line 28: you can create new pipe for the get tooltip also, to get rid of the function in the html

apps/admin-gui/src/app/shared/components/dialogs/add-member-dialog/add-member-dialog.component.html
```diff
@@ -1,80 +1,91 @@
...
-      matTooltipPosition="above"
-      [matTooltipDisabled]="selection.selected.length === 0 || this.data.type !== 'group' || (list !== undefined && list.addAuth)">
+    <div mat-dialog-actions>
+      <button mat-flat-button class="ml-auto" (click)="cancel.emit()">
+        {{'DIALOGS.ADD_MEMBERS.CANCEL' | translate}}
+      </button>
+      <span
+        matTooltip="{{'DIALOGS.ADD_MEMBERS.ADD_MEMBER_PERMISSION_TOOLTIP' | translate}}"
+        matTooltipPosition="above"
+        [matTooltipDisabled]="selection.selected.length === 0 || addAuth">
```
xkureck, line 32:
```suggestion
        [matTooltipDisabled]="selection.selected.length === 0 || inviteAuth">
```
this is tooltip for invite button

apps/admin-gui/src/app/shared/components/dialogs/add-member-dialog/add-member-dialog.component.html
```diff
@@ -1,80 +1,91 @@
...
-        [matMenuTriggerFor]="menu"
-        class="ml-2 dropdown-toggle"
+        mat-flat-button
+        class="ml-2"
         color="accent"
-        mat-flat-button>
-        {{'DIALOGS.ADD_MEMBERS.INVITE' | translate}}
+        data-cy="add-button"
+        (click)="add.emit()"
+        [disabled]="selection.selected.length === 0 || !addAuth">
```
xkureck, line 56: maybe we would like some sort of tooltip when this button is disable due to priviledges

apps/admin-gui/src/app/shared/components/dialogs/add-member-dialog/add-member-dialog.component.html
```diff
@@ -1,80 +1,105 @@
...
+              {{failedCandidate.errorName}}
+            </div>
+          </td>
+        </ng-container>
+        <tr *matHeaderRowDef="['userName', 'error']" mat-header-row></tr>
+        <tr *matRowDef="let failedCandidate; columns: ['userName', 'error'];" mat-row></tr>
+      </table>
+      <div mat-dialog-actions *ngIf="!loading">
+        <button mat-flat-button class="ml-auto" (click)="cancel.emit()">
+          {{'DIALOGS.ADD_MEMBERS.CLOSE' | translate}}
```
xkureck, line 100: when user click on this button, the table in the vo/group-members page don't refresh - but In this situation some members might be added but user don't see the added members

xkureck (changes requested):
image
Not the scope of the feature - but please fix this so the disabled checkbox cant be checked by clicking on the select add checkbox - that's easy fix, just use different method/parameters in the method isAllSelected/masterToggle

apps/admin-gui/src/app/shared/components/dialogs/add-member-dialog/add-member-dialog.component.html
```diff
@@ -1,80 +1,105 @@
...
+            {{failedCandidate.errorMsg}}
+            <div class="text-muted">
+              {{failedCandidate.errorName}}
+            </div>
+          </td>
+        </ng-container>
+        <tr *matHeaderRowDef="['userName', 'error']" mat-header-row></tr>
+        <tr *matRowDef="let failedCandidate; columns: ['userName', 'error'];" mat-row></tr>
+      </table>
+      <div mat-dialog-actions *ngIf="!loading">
```
xkureck, line 98: move this div with mat-dialog-actions out of the div with mat-dialog-content - the close button is not on the right spot right now

xkureck (changes requested):
image

after this minor changes it can be merged

xkureck (approved):

PR #1041, opened 2022-05-20
feat(admin): include groups from members organization

xkureck:
* added new feature to include groups from member organizations


bodnara (changes requested):
The rights to include groups from different organizations are evaluated wrongly (checkbox is disabled). Perhaps there is some check that both groups should be from same vo, however this is just a guess I did not investigate further.

xkureck:
> The rights to include groups from different organizations are evaluated wrongly (checkbox is disabled). Perhaps there is some check that both groups should be from same vo, however this is just a guess I did not investigate further.

There were some inconsistency in name of the policies to add/remove relationship
on backend they merged new names for those policies and on the frontend it was just on PR waiting for review
https://github.com/CESNET/perun-web-apps/pull/1044

HejdaJakub (approved):

apps/admin-gui/src/app/shared/components/dialogs/create-relation-dialog/create-relation-dialog.component.ts
```diff
@@ -68,6 +78,25 @@ export class CreateRelationDialogComponent implements OnInit {
     this.dialogRef.close(false);
   }
 
+  getGroupsToInclude(voId: number): void {
```
bodnara, line 81: Change `this.data.group.voId` to `this.data.voId` on all occurences in this method. So there are not three different id references which is a little confusing.

bodnara (approved):

PR #1048, opened 2022-05-30
build(admin): updated OpenAPI Generator to 6.0.0

xkureck:
* updated OpenAPI Generator to 6.0.0


apps/user-profile/src/app/pages/settings-page/settings-alternative-passwords/settings-alternative-passwords.component.ts
```diff
@@ -123,9 +119,11 @@ export class SettingsAlternativePasswordsComponent implements OnInit {
     this.attributesManagerService
       .getUserAttributeByName(this.userId, `urn:perun:user:attribute-def:def:altPasswords:einfra`)
       .subscribe((att) => {
-        this.altPasswordsAttribute = att;
         if (att.value) {
```
bodnara, line 122: Maybe we should also check here that the attribute value is of correct type. Also please use some meaningful varibale name instead of `foo`.

apps/user-profile/src/app/pages/settings-page/settings-data-quotas/settings-data-quotas.component.ts
```diff
@@ -76,15 +76,19 @@ export class SettingsDataQuotasComponent implements OnInit {
     this.attributesManagerService.getResourceAttributes(id).subscribe((atts) => {
       let quotaAttribute = atts.find((att) => att.friendlyName === 'dataQuotas');
       if (quotaAttribute?.value) {
```
bodnara, line 78: Same comment regarding the check.

apps/user-profile/src/app/pages/settings-page/settings-data-quotas/settings-data-quotas.component.ts
```diff
@@ -76,15 +76,19 @@ export class SettingsDataQuotasComponent implements OnInit {
...
-        this.currentQuota = quotaAttribute.value[keys[0]] as string;
+        const values = Object.entries(quotaAttribute.value as { [s: string]: string }).map(
+          (entry) => String(entry[1])
+        );
+        this.currentQuota = values[0];
       } else {
         this.currentQuota = '';
       }
       quotaAttribute = atts.find((att) => att.friendlyName === 'defaultDataQuotas');
       if (quotaAttribute) {
```
bodnara, line 87: Here we don't even check if the value exists.

bodnara (approved):

PR #1050, opened 2022-06-01
feat(profile): authentication settings page can be without security images

HejdaJakub:
* Now is possible to deploy user-profile without security images.
* If the property enable_security_image is set to false, then the Profile->Settings->Authentication page will contain just enforce mfa functionality (without security images).


bodnara (approved):

mattjoke (approved):
Looks good

PR #1051, opened 2022-06-01
fix(admin): Bugfixes of perun admin section

xkureck:
* one bigger bug founded in page user-account
* the other are just 1 line improvements


HejdaJakub (changes requested):
These changes look good, but I noticed, that the status column is displayed on the Groups page (`status` column) and on the Accounts page (`groupStatus` column), but no statuses are displayed.

xkureck:
> These changes look good, but I noticed, that the status column is displayed on the Groups page (`status` column) and on the Accounts page (`groupStatus` column), but no statuses are displayed. 

good point

HejdaJakub (approved):

PR #1052, opened 2022-06-01
fix(admin): bugfixes in vo section

HejdaJakub:
* Service accounts page - fixed parameter which was broking loading of all this page.
* Membership page - fixed correct loading of initial values and fixed behavior while clicking on toggles.


bodnara (changes requested):
~~On the membership page if you disable some additional setting (that was previously enabled) the save button stays disabled.~~

Edit: Apparently this is a feature and not a bug.

bodnara (approved):

PR #1053, opened 2022-06-02
fix(admin): disable checkboxes for members from member VOs

xflord:
* on the members section of a VO, disable checkboxes for members that come from member VOs, which previously was doable and resulted in an error.

### DIFF
--- a/apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-members/vo-members.component.ts
+++ b/apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-members/vo-members.component.ts
@@ -43,6 +43,7 @@ export class VoMembersComponent implements OnInit {
     Urns.USER_DEF_ORGANIZATION,
     Urns.USER_DEF_PREFERRED_MAIL,
     Urns.MEMBER_DEF_EXPIRATION,
+    Urns.MEMBER_LIFECYCLE_ALTERABLE,
   ];
   statuses = new FormControl();
   statusList = ['VALID', 'INVALID', 'EXPIRED', 'DISABLED'];
--- a/libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.html
+++ b/libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.html
@@ -34,8 +34,8 @@
             [aria-label]="checkboxLabel(member)"
             [checked]="selection.isSelected(member)"
             attr.data-cy="{{member.user.firstName | lowercase}}-checkbox"
-            [disabled]="member.membershipType === 'INDIRECT'"
-            [matTooltip]="member.membershipType === 'INDIRECT' ? ('MEMBERS_LIST.CHECKBOX_TOOLTIP' | translate):''"
+            [disabled]="(member | memberListCheckboxDisabled)"
+            [matTooltip]="(member | memberCheckboxLabel)"
             color="primary">
           </mat-checkbox>
         </td>
--- a/apps/admin-gui/src/assets/i18n/en.json
+++ b/apps/admin-gui/src/assets/i18n/en.json
@@ -2195,7 +2195,8 @@
     "NO_MEMBERS_ALERT": "No members present",
     "NO_FILTER_RESULTS_ALERT": "No members are matching your query",
     "INDIRECT_MEMBER": "Indirect member",
-    "CHECKBOX_TOOLTIP": "Indirect members cannot be removed"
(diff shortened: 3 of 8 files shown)


apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-members/vo-members.component.ts
```diff
@@ -43,6 +43,7 @@ export class VoMembersComponent implements OnInit {
     Urns.USER_DEF_ORGANIZATION,
     Urns.USER_DEF_PREFERRED_MAIL,
     Urns.MEMBER_DEF_EXPIRATION,
+    'urn:perun:member:attribute-def:virt:isLifecycleAlterable',
```
xkureck, line 46: add this attribute to the `Urns` class and use it here like the other attributes

libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.html
```diff
@@ -34,8 +34,8 @@
             [aria-label]="checkboxLabel(member)"
             [checked]="selection.isSelected(member)"
             attr.data-cy="{{member.user.firstName | lowercase}}-checkbox"
-            [disabled]="member.membershipType === 'INDIRECT'"
-            [matTooltip]="member.membershipType === 'INDIRECT' ? ('MEMBERS_LIST.CHECKBOX_TOOLTIP' | translate):''"
+            [disabled]="member.membershipType === 'INDIRECT' || !(member | memberLifecycleAlterable)"
```
xkureck, line 37:
```suggestion
            member | memberListCheckboxDisabled"
```
and in that pipe you will make decision to disable the checkbox if the member is indirect or if the attribute is lifecycle alterable is false

libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.html
```diff
@@ -34,8 +34,8 @@
...
+            [matTooltip]="checkboxTooltip(member)"
```
xkureck, line 38: for tooltip also do pipe
when you are using some method in HTML, the method is called everysecond or something like that to update the text
the pipe is called only when the input parameter Is changed

libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.ts
```diff
@@ -34,11 +35,17 @@ import { merge } from 'rxjs';
 import { tap } from 'rxjs/operators';
 import { TableWrapperComponent } from '@perun-web-apps/perun/utils';
 import { TableConfigService } from '@perun-web-apps/config/table-config';
+import { MemberLifecycleAlterablePipe } from '../../../../pipes/src/lib/member-LifecycleAlterable.pipe';
```
xkureck, line 38: add this pipe to file /libs/perun/pipes/src/index.ts and the import path will be much more nicer

libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.ts
```diff
@@ -34,11 +35,17 @@ import { merge } from 'rxjs';
...
 
 @Component({
   selector: 'perun-web-apps-members-dynamic-list',
   templateUrl: './members-dynamic-list.component.html',
   styleUrls: ['./members-dynamic-list.component.css'],
+  providers: [
+    {
+      provide: MemberLifecycleAlterablePipe,
+    },
+  ],
```
xkureck, line 48: dont have to be there

libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.ts
```diff
@@ -163,6 +172,15 @@ export class MembersDynamicListComponent implements AfterViewInit, OnInit, OnCha
     return `${this.selection.isSelected(row) ? 'deselect' : 'select'} row ${row.id + 1}`;
   }
 
+  checkboxTooltip(member: RichMember): string {
+    if (member.membershipType === 'INDIRECT') {
+      return this.translator.instant('MEMBERS_LIST.CHECKBOX_TOOLTIP_INDIRECT');
```
xkureck, line 177:
```suggestion
      return this.translator.instant('MEMBERS_LIST.CHECKBOX_TOOLTIP_INDIRECT') as string;
```

xkureck (changes requested):

xkureck (approved):

PR #1057, opened 2022-06-03
fix(admin): bugfixes in groups and member section

xkostka2:
* fixed moving group to top-level
* fixed expiration settings
* fixed error popping up when dynamic list of applications is empty


apps/admin-gui/src/app/vos/components/expiration-settings/expiration-settings.component.ts
```diff
@@ -341,7 +341,7 @@ export class ExpirationSettingsComponent implements OnInit, OnChanges {
       return null;
     }
 
-    let period = config.specialLoa.toString() + '|';
+    let period = config.specialLoa?.toString() + '|' ?? '';
```
bodnara, line 344: If the `config.specialLoa` is not defined this will be evaluated to `undefined|`. However there is a null check in this method so I am not sure what is this trying to achieve.

bodnara (changes requested):

bodnara (approved):

PR #1060, opened 2022-06-07
feat(admin): new attribute rights system

bodnara:
* attributes rights are set up using collections of policies
* created components to control attribute rights, collections and policies
* refactored components related to this changes


xkureck (changes requested):
image
missing the "OR" and "AND"

personally I would pu there some sort of alert that there are no set rights for this attribute or something like that
image

bodnara:
Should be tested on branch: https://github.com/CESNET/perun/tree/attributeRights

HejdaJakub (approved):
Good job, it looks good.

PR #1068, opened 2022-06-21
feat(admin): add a service search filter on facility resources list

xflord:
* user can now filter resources by services that are assigned to those resources

requires that https://github.com/CESNET/perun/pull/3632 is merged into backend

### DIFF
--- a/apps/admin-gui/src/app/facilities/pages/facility-detail-page/facility-resources/facility-resources.component.ts
+++ b/apps/admin-gui/src/app/facilities/pages/facility-detail-page/facility-resources/facility-resources.component.ts
@@ -2,7 +2,13 @@ import { Component, HostBinding, OnInit } from '@angular/core';
 import { SelectionModel } from '@angular/cdk/collections';
 import { MatDialog } from '@angular/material/dialog';
 import { RemoveResourceDialogComponent } from '../../../../shared/components/dialogs/remove-resource-dialog/remove-resource-dialog.component';
-import { FacilitiesManagerService, Facility, RichResource } from '@perun-web-apps/perun/openapi';
+import {
+  FacilitiesManagerService,
+  ServicesManagerService,
+  Facility,
+  RichResource,
+  Service,
+} from '@perun-web-apps/perun/openapi';
 import { CreateResourceDialogComponent } from '../../../../shared/components/dialogs/create-resource-dialog/create-resource-dialog.component';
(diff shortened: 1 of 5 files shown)


xkostka2 (changes requested):
Filtering by service does not work.

apps/admin-gui/src/app/facilities/pages/facility-detail-page/facility-resources/facility-resources.component.ts
```diff
@@ -109,4 +135,10 @@ export class FacilityResourcesComponent implements OnInit {
       }
     });
   }
+
+  serviceSelected(service: Service): void {
+    this.selectedService = service;
+    this.refreshTable();
+    this.cd.detectChanges();
```
bodnara, line 142: Don't think we need to detect changes here.

bodnara (changes requested):
I think the service select field is a little bit too wide while the filter filed is quite small. Both could be adjusted. Also the margin between select and filter is inconsitent with other elements in the row.
image

The `serviceSlected` method is triggered twice each time service is selected which results in the table refreshing twice as well.

apps/admin-gui/src/app/facilities/pages/facility-detail-page/facility-resources/facility-resources.component.ts
xflord: Rarely a 'Expression has changed after it was checked' error was thrown when refresing the table. I couldn't quite figure out why but this fixed it

bodnara (changes requested):
So the solution for the double trigger is valid and we use this on multiple places across the app. But maybe we could solve the root of this problem so no one else has to suffer. The culprit in this case is inside of `entity-search-select.component.ts`.  Specifically the `entitiesCtrl`, we are subscribing to value changes and immediately emitting new value. A better option is to emit only when the value is different than the previous one using the `distinctUntilChanged` operator.

xflord:
> So the solution for the double trigger is valid and we use this on multiple places across the app. But maybe we could solve the root of this problem so no one else has to suffer. The culprit in this case is inside of `entity-search-select.component.ts`. Specifically the `entitiesCtrl`, we are subscribing to value changes and immediately emitting new value. A better option is to emit only when the value is different than the previous one using the `distinctUntilChanged` operator.

Will do

bodnara (approved):

xkostka2 (approved):

PR #1071, opened 2022-06-23
fix(admin): correctly display group status when listing groups for member

xflord:
* display group status on User accounts page
* display group status on Member VO detail page

### DIFF
--- a/libs/perun/urns/src/lib/perun-urns.ts
+++ b/libs/perun/urns/src/lib/perun-urns.ts
@@ -3,6 +3,7 @@ export class Urns {
   static MEMBER_DEF_EXPIRATION = 'urn:perun:member:attribute-def:def:membershipExpiration';
   static MEMBER_DEF_GROUP_EXPIRATION =
     'urn:perun:member_group:attribute-def:def:groupMembershipExpiration';
+  static MEMBER_GROUP_STATUS = 'urn:perun:member_group:attribute-def:virt:groupStatus';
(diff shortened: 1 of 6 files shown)


xflord:
In order for this to work, I had to create a new AttributeDefinition for groupStatus. This lead me to believe that the attribute wasn't properly initiliazied in backend - hence PR https://github.com/CESNET/perun/pull/3637 . Correct me if I'm wrong.

libs/perun/urns/src/lib/perun-urns.ts
```diff
@@ -3,6 +3,7 @@ export class Urns {
   static MEMBER_DEF_EXPIRATION = 'urn:perun:member:attribute-def:def:membershipExpiration';
   static MEMBER_DEF_GROUP_EXPIRATION =
     'urn:perun:member_group:attribute-def:def:groupMembershipExpiration';
+  static MEMBER_GROUP_GROUP_STATUS = 'urn:perun:member_group:attribute-def:virt:groupStatus';
```
bodnara, line 6:
```suggestion
  static MEMBER_GROUP_STATUS = 'urn:perun:member_group:attribute-def:virt:groupStatus';
```

bodnara:
> In order for this to work, I had to create a new AttributeDefinition for groupStatus. This lead me to believe that the attribute wasn't properly initiliazied in backend - hence PR [CESNET/perun#3637](https://github.com/CESNET/perun/pull/3637) . Correct me if I'm wrong.

I believe you are right.

bodnara (approved):

PR #1073, opened 2022-06-29
fix: use localStorage for storing access token/refresh token

xkureck:
* storing tokens is moved from session storage to local storage because session storage is storing the data only for one tab,
but we want to share tokens across multiple tabs
* refreshing token is now only triggered when access token is expired
* expiration of tokens was moved from 0.75(default) to random number from 0.5 to 0.75 so the refreshing of the token is not triggered by multiple tabs at the same time
* sidemenu items url links members applications and groups fixed

### DIFF
--- a/apps/admin-gui/src/app/app.module.ts
+++ b/apps/admin-gui/src/app/app.module.ts
@@ -30,7 +30,7 @@ import {
 } from 'ngx-perfect-scrollbar';
 import { PerunSharedComponentsModule } from '@perun-web-apps/perun/components';
 import { PerunLoginModule } from '@perun-web-apps/perun/login';
-import { OAuthModule } from 'angular-oauth2-oidc';
+import { OAuthModule, OAuthStorage } from 'angular-oauth2-oidc';
 
 export const API_INTERCEPTOR_PROVIDER: Provider = {
   provide: HTTP_INTERCEPTORS,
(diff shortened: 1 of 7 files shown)


mattjoke:
The code looks good, but the ESLint fails.

apps/admin-gui/src/app/shared/side-menu/side-menu-item.service.ts
```diff
@@ -693,15 +693,15 @@ export class SideMenuItemService {
     if (this.routePolicyService.canNavigate('members-groups', member)) {
       links.push({
         label: 'MENU_ITEMS.MEMBER.GROUPS',
-        url: [`//organizations/${member.voId}/members/${member.id}/groups`],
+        url: [`/organizations/${member.voId}/members/${member.id}/groups`],
```
bodnara, line 696: These side-menu changes are quite out of scope of this PR about token storing.

libs/perun/services/src/lib/init-auth.service.ts
```diff
@@ -8,6 +8,7 @@ import { MatDialog } from '@angular/material/dialog';
 import { UserDontExistDialogComponent } from '@perun-web-apps/general';
 import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
 import { Router } from '@angular/router';
+import { OAuthStorage } from 'angular-oauth2-oidc';
```
bodnara, line 11: Remove unused import.

bodnara (commented):
Well done with this changes, surely a great improvement for users.

PR #1076, opened 2022-06-30
fix(profile): sort vos and resources on the mailing lists page

xflord:
* vos and resources are now sorted alphabetically on the mailing lists opt-out page


xflord:
Sorting of resources wasn't yet tested manually (didn't have time to figure out how to create mailing lists for vos).

PR #1080, opened 2022-07-10
feat(lib): create new translate service

sarkapalkovicova:
Create a new service that extends TranslateService.

### DIFF
--- a/libs/perun/services/src/lib/perun-translate.service.ts
+++ b/libs/perun/services/src/lib/perun-translate.service.ts
@@ -1,0 +1,11 @@
+import { TranslateService } from '@ngx-translate/core';
+import { Injectable } from '@angular/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class PerunTranslateService extends TranslateService {
+  instant(key: string | Array<string>, interpolateParams?: unknown): string {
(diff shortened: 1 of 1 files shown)


libs/perun/services/src/lib/translate.service.ts
```diff
@@ -0,0 +1,7 @@
+import { TranslateService as TranslateServiceNgx } from '@ngx-translate/core';
+
+export class TranslateService extends TranslateServiceNgx {
+  instant(key: string | Array<string>, interpolateParams?: unknown): string {
+    return super.instant(key, interpolateParams) as string;
```
bodnara, line 5:
```suggestion
    return String(super.instant(key, interpolateParams));
```

bodnara (changes requested):
Looks good, I have just a couple of comments.

For a class to be an actual service it needs the `@Injectable` decorator with a information where it is provided, usually `providenIn: root`. You can add this manually but  you can use `ng generate service 'name'` or `ng g s 'name'` in short to create service. 

I would also suggest to use different name so there is no confusion which TranslateService is actually imported. Maybe `PerunTranslateService` but I leave that up to you.

bodnara (approved):

PR #1081, opened 2022-07-11
feat: getter for typed config properties

bodnara:
* created new method `getProperty` in `StoreService`
* old method `get` is marked as deprecated
* created interfaces to represent config properties

### DIFF
--- a/libs/config/src/lib/app-config.service.ts
+++ b/libs/config/src/lib/app-config.service.ts
@@ -5,10 +5,11 @@
    @typescript-eslint/no-unsafe-assignment */
 import { Injectable } from '@angular/core';
 import { HttpClient, HttpHeaders } from '@angular/common/http';
-import { StoreService, PerunConfig } from '@perun-web-apps/perun/services';
+import { StoreService } from '@perun-web-apps/perun/services';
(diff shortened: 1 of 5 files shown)


HejdaJakub (changes requested):
If the method `getProperty()` should be a complete substitute for the `get()` method, it should accept more than one parameter or array - we need it for `theme`, `oidc_client`, `footer`, etc.

bodnara:
> If the method `getProperty()` should be a complete substitute for the `get()` method, it should accept more than one parameter or array - we need it for `theme`, `oidc_client`, `footer`, etc.

It is not a one for one substitute per say. However you can achieve same results using the dot notation on returned value. E.g. `get('footer', 'logo')` can be replaced with `getProperty('footer').logo`.

HejdaJakub:
> > If the method `getProperty()` should be a complete substitute for the `get()` method, it should accept more than one parameter or array - we need it for `theme`, `oidc_client`, `footer`, etc.
> 
> It is not a one for one substitute per say. However you can achieve same results using the dot notation on returned value. E.g. `get('footer', 'logo')` can be replaced with `getProperty('footer').logo`.

Ok, so it looks good!

HejdaJakub (approved):

PR #1089, opened 2022-07-18
feat(admin): replace ngx-perfect-scrollbar by ngx-scrollbar

HejdaJakub:
* Library ngx-perfect-scrollbar has the last release (version 10.0.0) and it will not be updated
anymore, so it's recommended to replace this library with another library such as angular native
ngx-scrollbar.
* Ngx-scrollbar is now using insted of ngx-perfect-scrollbar and the ngx-perfect-scrollbar library has been uninstalled.


bodnara (approved):
It seems to work well, good job.

PR #1090, opened 2022-07-18
feat(admin): add group roles page

Johaney-s:
* added page with group roles accessible via group detail
* refactored user-roles component and extracted common logic


bodnara (commented):
This looks good. One more thing that could make sense it to cache the result of requests so we don't send new request each time the panel is opened. This can be achieved using the `shareReplay` operator.

bodnara (approved):

xkureck (approved):

PR #1091, opened 2022-07-19
test(admin): e2e tests for PERUNADMIN

cuadradek:
- Added e2e tests for PERUNADMIN.


sarkapalkovicova:
It looks good, and all tests passed. Please fix Prettier.

HejdaJakub (approved):
Good job!

NOTE: IMHO in these tests it's not a problem, but sometimes can be problematic when you intercept the response at the beginning of the test and you will get the response from a defined URL several times during this one test. Afterward, the wait method checks the first occurrence of the interception, so the behavior can be different than you expected (you can see that this test doesn't wait for the `@getAttributesDefinition` at the end as you probably expected). But in this case, it's not a problem.
image

PR #1093, opened 2022-07-20
feat(admin): show entities of membership role on user roles page

Johaney-s:
* only IDs of vos, groups, resources and facilities were shown so far
* now tables with these entities where user has membership role are shown


Johaney-s:
First commit contains not yet merged (https://github.com/CESNET/perun-web-apps/pull/1090) changes in roles page.
Membership and sponsorship roles will only show up if approaching roles page via admin gui -> my profile -> roles.
After merging https://github.com/CESNET/perun/pull/3657 it will also be visible in admin gui -> user's detail -> roles.

xkostka2 (changes requested):
For some reason, when I am viewing any user's roles, the description above the tables is missing, while when going through My profile, it is displayed just fine. 
image

xkostka2 (approved):

xkureck (approved):

PR #1095, opened 2022-07-26
feat: server side pagination

bodnara:
Within the group-list
* replaced function calls with pipes
* no redundant reactions to changes
* moved the remove authorization to corresponding components
* removed unused code

Created DynamicDataSource in `perun-models` to handle filter, sort and pagination changes, this data source should be used instead of the existing one. Example usage of new data source was implemented in group-list, it does not require to duplicate table logic as the existing one.

### DIFF
--- a/apps/admin-gui/src/app/vos/pages/group-detail-page/group-subgroups/group-subgroups.component.ts
+++ b/apps/admin-gui/src/app/vos/pages/group-detail-page/group-subgroups/group-subgroups.component.ts
@@ -11,7 +11,8 @@ import { MatSlideToggle } from '@angular/material/slide-toggle';
 import { GroupFlatNode } from '@perun-web-apps/perun/models';
 import { MoveGroupDialogComponent } from '../../../../shared/components/dialogs/move-group-dialog/move-group-dialog.component';
 import { EntityStorageService, GuiAuthResolver } from '@perun-web-apps/perun/services';
-import { GroupsListComponent, GroupsTreeComponent } from '@perun-web-apps/perun/components';
+import { Observable } from 'rxjs';
+import { map, startWith } from 'rxjs/operators';
 
 @Component({
   selector: 'app-group-subgroups',
@@ -23,12 +24,7 @@ export class GroupSubgroupsComponent implements OnInit {
 
   // used for router animation
   @HostBinding('class.router-component') true;
-  @ViewChild('tree', {})
-  tree: GroupsTreeComponent;
-  @ViewChild('list', {})
-  list: GroupsListComponent;
-  @ViewChild('toggle', { static: true })
-  toggle: MatSlideToggle;
+  @ViewChild('toggle', { static: true }) toggle: MatSlideToggle;
 
   group: Group;
   groups: Group[] = [];
@@ -41,6 +37,16 @@ export class GroupSubgroupsComponent implements OnInit {
   createAuth: boolean;
   deleteAuth: boolean;
   routeAuth: boolean;
+  removeAuth$: Observable<boolean> = this.selected.changed.pipe(
+    map((changed) => {
+      return changed.source.selected.reduce(
+        (acc, grp) =>
+          acc && this.guiAuthResolver.isAuthorized('deleteGroup_Group_boolean_policy', [grp]),
+        true
+      );
+    }),
+    startWith(true)
+  );
 
   constructor(
     private dialog: MatDialog,
--- a/apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-groups/vo-groups.component.ts
+++ b/apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-groups/vo-groups.component.ts
@@ -1,49 +1,108 @@
-import { Component, HostBinding, OnInit, ViewChild } from '@angular/core';
+import { ChangeDetectionStrategy, Component, HostBinding, OnInit, ViewChild } from '@angular/core';
 import { MatDialog } from '@angular/material/dialog';
 import { CreateGroupDialogComponent } from '../../../../shared/components/dialogs/create-group-dialog/create-group-dialog.component';
 import { DeleteGroupDialogComponent } from '../../../../shared/components/dialogs/delete-group-dialog/delete-group-dialog.component';
 import { SelectionModel } from '@angular/cdk/collections';
 import { MoveGroupDialogComponent } from '../../../../shared/components/dialogs/move-group-dialog/move-group-dialog.component';
-import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
-import { Group, GroupsManagerService, RichGroup, Vo } from '@perun-web-apps/perun/openapi';
-import { GroupFlatNode } from '@perun-web-apps/perun/models';
+import {
+  downloadData,
+  getDataForExport,
+  getDefaultDialogConfig,
+} from '@perun-web-apps/perun/utils';
+import {
+  Group,
+  GroupsManagerService,
+  GroupsOrderColumn,
+  PaginatedRichGroups,
+  RichGroup,
+  Vo,
+} from '@perun-web-apps/perun/openapi';
+import { GroupFlatNode, GroupWithStatus, PageQuery } from '@perun-web-apps/perun/models';
 import { TABLE_VO_GROUPS } from '@perun-web-apps/config/table-config';
 import { Urns } from '@perun-web-apps/perun/urns';
 import { EntityStorageService, GuiAuthResolver } from '@perun-web-apps/perun/services';
 import { MatSlideToggle } from '@angular/material/slide-toggle';
-import { GroupsListComponent, GroupsTreeComponent } from '@perun-web-apps/perun/components';
+import { BehaviorSubject, merge, Observable } from 'rxjs';
+import { map, startWith, switchMap, tap } from 'rxjs/operators';
+import { GroupUtilsService } from '@perun-web-apps/perun/services';
 
 @Component({
   selector: 'app-vo-groups',
   templateUrl: './vo-groups.component.html',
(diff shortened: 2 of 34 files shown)


libs/perun/components/src/lib/groups-list/groups-list.component.html
```diff
@@ -17,26 +15,26 @@
...
         </th>
-        <td *matCellDef="let row" class="static-column-size align-checkbox" mat-cell>
+        <td *matCellDef="let group" class="static-column-size align-checkbox" mat-cell>
           <span
-            matTooltip="{{getCheckboxTooltipMessage(row) | translate}}"
+            *ngIf="{disable: group | disableGroupSelect: disableMembers : disableGroups : groupsToDisableCheckbox} as vm"
+            matTooltip="{{group | groupCheckboxTooltip: false | translate}}"
             [matTooltipPosition]="'above'"
-            [matTooltipDisabled]="!disableSelect(row)">
+            [matTooltipDisabled]="!vm.disable">
```
xkureck, line 31: also what is 'vm'?

xkureck (changes requested):
image
so for some reason everytime when I go to the vo-groups page even when I have the page size set to 10, I get only 5, and GUI also sends to backend that it only wants 5, even the pagesize is clearly 10
image
There is some failing HTTP request, because you didn't set the query, so maybe that could be connected

Also your changes somehow influenced the tree view in the vo-group page. When the data are loaded initially into the page the user can see the alert that there are no groups until the groups are loaded. Also on this page, when user clicks on the reload button the page doesn't let user know somehow that the operation was made - no feedback that his action had consequences.

image

Also as a Perun admin
image
Also some tooltips
image
image


Also on page vo-groups, the groups are for some reason sorted according to ID instead of name

I appreciate the effort to refactor group-list component.

image
When i try something to filter on vo-group list view

image
When I click on select all and in the table are no disabled checkboxes the selection doesn't happen

On member-groups page: (also on add groups dialog on the same page)
image
Also when you are already refactoring, please make

libs/perun/utils/src/lib/perun-utils.ts
```diff
@@ -723,7 +724,7 @@ export function customDataSourceFilterPredicate<T>(
   return dataStr.toLowerCase().includes(filter);
 }
 
-export function parseAttribute(data: Author, nameOfAttribute: string): string {
+export function parseAttribute<T extends Attributable>(data: T, nameOfAttribute: string): string {
```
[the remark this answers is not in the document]
bodnara: Yes, I think your soulution would make more sense.

libs/perun/components/src/lib/groups-list/groups-list.component.html
bodnara: So the `*ngIf="{some: thing | pipe} as vm"`  is a way of reusing pipe results within the element. To my understanding "vm" (view model) is a common name for this. However we could use more descriptive naming.

xkureck: maybe something like vmThing would be probably more descriptive of what the VM is?

apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-groups/vo-groups.component.ts
```diff
@@ -1,43 +1,92 @@
...
+      setTimeout(() => this.loadingSubject$.next(false), 200);
+    }),
+    startWith({ data: [], totalCount: 0, offset: 0, pageSize: 0 })
+  );
+
   showGroupList = false;
   selected = new SelectionModel<RichGroup>(true, []);
-  loading: boolean;
+  loadingSubject$ = new BehaviorSubject(false);
+  loading$ = merge(this.loadingSubject$, this.nextPage.pipe(mapTo(true)));
```
HejdaJakub, line 64:
```suggestion
  loading$: Observable<boolean> = merge(this.loadingSubject$, this.nextPage.pipe(mapTo(true)));
```
mapTo(...) - deprecated, use map() instead of this. Also if I want to refresh the vo-groups page, on this page appears the spinner, but this spinner will never be hidden again and you can see infinity loading.

HejdaJakub (changes requested):
There is one problem with select boxes (but I am not sure, that there is any simple solution). If I select some checkbox and then somehow change the parameters of this page (click on next page, change filter value, change items per page), then all already selected checkboxes are deselected. This problem appears also on other pages server-side paginated, so it's not just a problem of this PR, but we should probably somehow deal with this issue in the future IMHO (not in this PR).

apps/admin-gui/src/app/vos/pages/group-detail-page/group-subgroups/group-subgroups.component.ts
```diff
@@ -41,6 +37,16 @@ export class GroupSubgroupsComponent implements OnInit {
...
+  removeAuth$: Observable<boolean> = this.selected.changed.pipe(
+    switchMap((changed) => {
+      const authorized = changed.source.selected.reduce(
+        (acc, grp) =>
+          acc && this.guiAuthResolver.isAuthorized('deleteGroup_Group_boolean_policy', [grp]),
+        true
+      );
+      return of(authorized);
+    })
+  );
```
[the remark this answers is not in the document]
bodnara: I realized the map solution makes more sense, but forgot to apply it everywhere. Should be consistent now.

libs/perun/components/src/lib/groups-list/groups-list.component.ts
```diff
@@ -1,105 +1,126 @@
...
+    } else if (!isDynamicDataSource(this.dataSource) && !paginated) {
+      this.dataSource.data = groups;
+    }
+
+    this.updateVoNames();
+  }
+
+  @Input() set filter(value: string) {
+    this.dataSource.filter = value;
+  }
```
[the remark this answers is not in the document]
bodnara: Yes the IDE suggests this for some reason, not sure why. This si not the cause for the error. The error occures when filter input is defined before the group input. This leads me to believe that inputs are initialized in order they are passed in, rather than order they are defined. 
This order would cause an error since datasource will be initialized only after the groups are passed:
```
<perun-web-apps-groups-list
        [filter]="filterValue"
        [groups]="groups">
</perun-web-apps-groups-list>
```

xkureck (changes requested):
When the data are loaded in the vo - groups page - the shadow that should be just on the page is on the whole page.
Also we on the default are sorting the group tables according to names - and you are sorting it according to the id by default
image
When the treeview is the users chosen option the vo-group page don't load at all and this is what user see

Also the loading on vo-groups page don't work fine - there is difference how we want to handle it in the group-tree component and the group-list - if we have chosen the treeview initially to show then when the data are loaded the on the background there is still the alert with "No results match the given search criteria." which looks ugly

image
Bad tooltip when creating relation - should be that the user dont have permission

image
Bad authorization when I want to remove included groups - I have right to do that but GUI tells me that i dont have rights to do that

sarkapalkovicova:
The existing way of obtaining data for export worked when we didn't paginate the list of groups. Now it needs to be edited.

libs/perun/components/src/lib/groups-list/groups-list.component.html
```diff
@@ -113,12 +111,13 @@
         </th>
         <td *matCellDef="let group" mat-cell>
           <i
-            class="material-icons {{getStatusAttribute(group) | groupStatusIconColor}}"
-            matTooltip="{{getStatusAttribute(group)}}"
+            *ngIf="{status: group.attributes | findAttribute: 'groupStatus'} as groupStatus"
+            class="material-icons {{groupStatus.status | memberStatusIconColor}}"
```
HejdaJakub, line 115: I think there should be `groupStatusIconColor` instead of `memberStatusIconColor` or the `memberStatusIconColor` pipe has to be changed, because now it requires RichMember as an argument, not a string.

HejdaJakub (changes requested):
On the vo-groups page there is still the alert behind the spinner (during the loading of some data for tree view and also for list view):
image

As @sarkapalkovicova said the export is still not working. Maybe we want to fix it in another PR? But there is no feedback about this issue.

libs/perun/components/src/lib/groups-list/groups-list.component.html
bodnara: I have used the groupStatusIconColor as is it on master.

HejdaJakub (changes requested):
Member-groups page is broken:
image

libs/perun/components/src/lib/groups-list/groups-list.component.html
```diff
@@ -115,12 +113,13 @@
         </th>
         <td *matCellDef="let group" mat-cell>
           <i
-            class="material-icons {{getStatusAttribute(group) | groupStatusIconColor}}"
-            matTooltip="{{getStatusAttribute(group)}}"
+            *ngIf="{status: group.attributes | findAttribute: 'groupStatus'} as groupStatus"
+            class="material-icons {{groupStatus.status | groupStatusIconColor}}"
```
HejdaJakub, line 117: I am not sure if the pipe will fail due to it, but there s an error: `Invalid number of arguments, expected 2..3 `

bodnara:
> Member-groups page is broken
I believe it was because of the new changes to isAllSelected and applied the same fix (ngIf always true). It seems to fix this.

libs/perun/components/src/lib/groups-list/groups-list.component.html
bodnara: It may cause problems in some specific circumstance. However I won't be fixing is here as it is already on master. Could be looked at in different issue.

HejdaJakub (approved):
I think that now it looks good.

PR #1096, opened 2022-07-27
feat: Account linking applications

xkureck:
* consolidator application
* linker application
* lib-linker library
* user profile implemented

How to test this: https://perunaai.atlassian.net/wiki/spaces/STRIBOG/pages/51314739/Implementa+n+detaily+a+testov+n+nov+ho+konsolid+toru#Testov%C3%A1n%C3%AD-frontend-aplikac%C3%AD(user-profile,-linker,-consolidator)-na-stroji-cloud4


apps/linker/src/app/app.module.ts
```diff
@@ -1,15 +1,105 @@
...
+  multi: true,
+};
+
+export function createTranslateLoader(http: HttpClient): TranslateHttpLoader {
+  return new TranslateHttpLoader(http, './assets/i18n/', '.json');
+}
+
+export function apiConfigFactory(store: StoreService): Configuration {
+  const params: ConfigurationParameters = {
+    basePath: store.get('api_url') as string,
```
bodnara, line 43:
```suggestion
    basePath: store.getProperty('api_url'),
```

apps/linker/src/app/components/show-result/show-result.component.html
```diff
@@ -0,0 +1,91 @@
+<div class="position-relative">
```
bodnara, line 1: Here again, perhaps a more general solution would help us reduce code duplicity.

apps/linker/src/app/components/show-result/show-result.component.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  closeWindow(): void {
+    window.close();
+  }
+
+  loginAgain(): void {
+    let queryParams = location.search.substring(1);
+    let index = 1;
+    if (this.linkerResult === 'IDENTITY_UNKNOWN') {
+      for (let i = 1; i < 10; i++) {
```
bodnara, line 55: This feels hacky, at least provide some comment as to what is happening here.

apps/linker/src/app/components/show-result/show-result.component.ts
```diff
@@ -0,0 +1,133 @@
...
+    }
+    sessionStorage.setItem('auth:queryParams', queryParams);
+    this.oauthService.initLoginFlow();
+  }
+
+  sendMessageToSupport(): void {
+    this.loading = true;
+    const currentUser = this.storeService.getPerunPrincipal().user;
+
+    const queryParams = location.search.substr(1);
```
bodnara, line 73:
```suggestion
    const queryParams = location.search.substring(1);
```

apps/linker/src/app/components/show-result/show-result.component.ts
```diff
@@ -0,0 +1,133 @@
...
+      this.rtMessages
+        .sentMessageToRTWithQueue(
+          'perun',
+          'Account linking: The accounts could not be automatically linked.',
+          this.getFullEmailBody(currentUser, formerUser)
+        )
+        .subscribe(
+          () => {
+            this.linkerResult = 'MESSAGE_SENT_TO_SUPPORT';
+            (window.opener as Window).postMessage(this.linkerResult, '*');
```
bodnara, line 86: There are multiple instances of accessing `window.opener` and casting it to `Window`, is it possible to save this to a variable and prevent this need to cast repeatedly.

apps/linker/src/app/components/show-result/show-result.component.ts
```diff
@@ -0,0 +1,133 @@
...
+            this.loading = false;
+            this.linkerResult = 'UNKNOWN_ERROR';
+            (window.opener as Window).postMessage(this.linkerResult, '*');
+          }
+        );
+    });
+  }
+
+  getFullEmailBody(currentUser: User, formerUser: User): string {
+    const instance = this.storeService.get('config') as string;
```
bodnara, line 99:
```suggestion
    const instance = this.storeService.getProperty('config');
```

apps/linker/src/app/components/show-result/show-result.component.ts
```diff
@@ -0,0 +1,133 @@
...
+    let text =
+      'Account linking: The accounts could not be automatically linked.\n' +
```
bodnara, line 101: Are we assuming that there will never be a requirement to localize these texts ?

apps/linker/src/app/service/linker-config.service.ts
```diff
@@ -0,0 +1,52 @@
...
+          });
+        } else {
+          return this.initAuthService.handleAuthStart();
+        }
+      });
+  }
+
+  private setApiUrl(): Promise<void> {
+    return new Promise((resolve) => {
+      this.authzSevice.configuration.basePath = this.storeService.get('api_url') as string;
```
bodnara, line 48:
```suggestion
      this.authzSevice.configuration.basePath = this.storeService.getProperty('api_url');
```

apps/linker/src/assets/config/defaultConfig.json
```diff
@@ -0,0 +1,23 @@
+{
```
bodnara, line 1: Please add config description for both applications to documentation.

apps/user-profile/src/app/pages/identities-page/add-identity-button/add-identity-button.component.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (this.storage.getProperty('use_new_consolidator')) {
+      let linkerUrl = this.otherApplicationsService.getUrlForOtherApplication('linker');
+      //next 'if' is there for testing purposes, when you want to use localhost url of linker
+      if (this.storage.getProperty('use_localhost_linker_url')) {
+        linkerUrl = 'http://localhost:4201/';
+      }
+      const openedWindow = openLinkerWindow(linkerUrl, this.oauthService.getAccessToken());
+      const event = this.refreshTable;
+      const notify = this.notificator;
+      const translator = this.translate;
```
bodnara, line 44: If you define the interval function with `() =>` this is not needed.

apps/user-profile/src/app/pages/identities-page/add-identity-button/add-identity-button.component.ts
```diff
@@ -0,0 +1,81 @@
...
+
+      let dialogRef: MatDialogRef<FocusOnLinkerDialogComponent, unknown>;
```
bodnara, line 46:
```suggestion
      let dialogRef: MatDialogRef<FocusOnLinkerDialogComponent, void>;
```

apps/user-profile/src/app/pages/identities-page/add-identity-button/add-identity-button.component.ts
```diff
@@ -0,0 +1,81 @@
...
+import { MatDialog } from '@angular/material/dialog';
+import { RegistrarManagerService } from '@perun-web-apps/perun/openapi';
+import { MatDialogRef } from '@angular/material/dialog/dialog-ref';
+
+@Component({
+  selector: 'perun-web-apps-add-identity-button',
+  templateUrl: './add-identity-button.component.html',
+  styleUrls: ['./add-identity-button.component.css'],
+})
+export class AddIdentityButtonComponent {
```
bodnara, line 20: Not sure why we would need a whole component for this button, feels a little strange.

libs/lib-linker/src/lib/lib-linker.ts
```diff
@@ -0,0 +1,52 @@
...
+  | 'OK'
+  | 'IDENTITY_UNKNOWN'
+  | 'IDENTITY_REGISTERED_ALREADY'
+  | 'IDENTITY_LINKED'
+  | 'IDENTITY_IDENTICAL'
+  | 'TOKEN_EXPIRED'
+  | 'UNKNOWN_ERROR'
+  | 'MESSAGE_SENT_TO_SUPPORT';
+
+export const LinkerResult = {
```
bodnara, line 11: I think with the power of TS you can cast all properties of object to given type automatically.

libs/lib-linker/src/lib/lib-linker.ts
```diff
@@ -0,0 +1,52 @@
...
+  let url = linkerUrl + '?formerToken=' + formerToken;
+  if (idpFilter) {
+    if (idpFilter.length !== 0) {
+      url += '&idphint=' + idpFilter.toString();
+    }
+  }
+  const width = 600;
+  const height = 600;
+
+  const top = (window.top as Window).outerHeight / 2 + (window.top as Window).screenY - height / 2;
```
bodnara, line 36: Create a variable for the `window.top` so there is no need to cast multiple times.

libs/lib-linker/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+  "include": [],
+  "references": [
+    {
+      "path": "./tsconfig.lib.json"
+    },
+    {
+      "path": "./tsconfig.spec.json"
+    }
+  ],
+  "compilerOptions": {
```
bodnara, line 13: I believe we don't set `compilerOptions` or `angularCompilerOptions` in this file for other libs.

libs/perun/login/src/lib/login-screen-base/login-screen-base.component.html
```diff
@@ -14,6 +14,9 @@
   <div *ngIf="application === 'user-profile'" class="mt-auto mb-auto">
     <a class="logo-container" [innerHTML]="logo"></a>
   </div>
+  <div *ngIf="application === 'consolidator'" class="mt-auto mb-auto">
+    <div [innerHTML]="logo" class="logo-container"></div>
+  </div>
```
bodnara, line 19: It looks like we only handle the publications app diferently, maybe other cases can be merged.

libs/perun/services/src/lib/ApiInterceptor.ts
```diff
@@ -133,4 +136,11 @@ export class ApiInterceptor implements HttpInterceptor {
     rpcError.payload = req.body as unknown as object;
     return rpcError;
   }
+
+  private isNotConsolidatorOrLinker(): boolean {
+    return !(
+      this.store.getProperty('application') === 'Linker' ||
+      this.store.getProperty('application') === 'Consolidator'
+    );
```
bodnara, line 144:
```suggestion
    const application = this.store.getProperty('application');
    return !(application  === 'Linker' || application ) === 'Consolidator');
```

libs/perun/services/src/lib/auth.service.ts
```diff
@@ -72,7 +82,7 @@ export class AuthService {
       scope: this.store.get('oidc_client', 'oauth_scopes') as string,
       clockSkewInSec: clockSkewInSec,
       timeoutFactor: timeoutFactor,
-      // sessionChecksEnabled: true,
+      userinfoEndpoint: this.store.get('oidc_client', 'user_info_endpoint_url') as string,
```
bodnara, line 85: Add to `OidcClient` if needed.
```suggestion
      userinfoEndpoint: this.store.getProperty('oidc_client').user_info_endpoint_url,
```

apps/linker/src/app/components/show-result/show-result.component.ts
xkureck: This text is sent to user support- so that would be kind of sad if they don't speak english

libs/perun/services/src/lib/ApiInterceptor.ts
xkureck: In my opinion the condition `!(application  === 'Linker' || application ) === 'Consolidator')` seems super not human readable, so I would like to stay it the way it is

apps/consolidator/src/app/components/list-of-identities/list-of-identities.component.ts
```diff
@@ -0,0 +1,144 @@
...
+  constructor(
+    private oauthService: OAuthService,
+    private userService: UsersManagerService,
+    private storeService: StoreService,
+    private initService: InitAuthService
+  ) {}
+
+  ngOnInit(): void {
+    this.loading = true;
+    void this.oauthService.loadUserProfile().then((userProfile) => {
```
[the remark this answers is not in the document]
xkureck: https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse
User info response can look super different and also if some property is null, then it is not present in the response
We just cannot guarantee that some property is there, also for different proxy's the response might and probably will look different

apps/consolidator/src/app/components/header/header.component.css
```diff
@@ -0,0 +1,31 @@
...
+  max-height: 64px;
+  top: 0;
+}
+
+.logo-container {
+  padding: 0 8px 0 8px;
+}
+
+.logo-container ::ng-deep img,
+.logo-container ::ng-deep svg {
```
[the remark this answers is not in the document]
xkureck: https://stackoverflow.com/questions/47024236/what-to-use-in-place-of-ng-deep

libs/lib-linker/src/lib/consolidation-result/consolidation-result.component.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  greenColor = '#e0ffd4';
+  blueColor = '#d4f2ff';
+  redColor = '#ffd4d4';
+  warningIcon = 'warning';
+  infoIcon = 'info';
+  messageSent = 'mark_email_read';
+
+  ngOnChanges(): void {
+    switch (this.result) {
```
bodnara, line 23: I would prefer if the `result` input would be a setter and this logic would be inside of it. Since this only needs to be evaluated when the result changes not on every change.

apps/consolidator/src/app/components/list-of-identities/list-of-identities.component.ts
```diff
@@ -0,0 +1,144 @@
...
+import { Attribute, UsersManagerService } from '@perun-web-apps/perun/openapi';
+import { MatTableDataSource } from '@angular/material/table';
+import { getAttribute } from '@perun-web-apps/perun/utils';
+import { Urns } from '@perun-web-apps/perun/urns';
+
+export interface TableInfo {
+  name: string;
+  value: string;
+}
+
```
bodnara, line 13:
```suggestion
interface ProfileInfo {
  name?: string;
  given_name?: string;
  family_name?: string;
  preferred_username?: string;
  nickname?: string;
  email?: string;
}

interface UserProfile {
  info: ProfileInfo;
}
```

apps/consolidator/src/app/components/list-of-identities/list-of-identities.component.ts
```diff
@@ -0,0 +1,144 @@
...
+    private oauthService: OAuthService,
+    private userService: UsersManagerService,
+    private storeService: StoreService,
+    private initService: InitAuthService
+  ) {}
+
+  ngOnInit(): void {
+    this.loading = true;
+    void this.oauthService.loadUserProfile().then((userProfile) => {
+      this.setIdpInfo(userProfile);
```
bodnara, line 39:
```suggestion
      this.setIdpInfo(userProfile as UserProfile);
```

apps/consolidator/src/app/components/list-of-identities/list-of-identities.component.ts
```diff
@@ -0,0 +1,144 @@
...
+    if (name !== null && name !== undefined && name !== '') {
+      this.info.push({ name: 'Name', value: name });
+    }
+    const login = this.getUserLogin(userProfile);
+    if (login !== null && login !== undefined && login !== '') {
+      this.info.push({ name: 'Login', value: login });
+    }
+    if ((userProfile['info'] as object)['email'] as string) {
+      this.info.push({ name: 'Email', value: (userProfile['info'] as object)['email'] as string });
+    }
```
bodnara, line 83:
```suggestion
  setIdpInfo(userProfile: UserProfile): void {
    const name = this.getUserName(userProfile);
    if (name) {
      this.info.push({ name: 'Name', value: name });
    }
    const login = this.getUserLogin(userProfile);
    if (login) {
      this.info.push({ name: 'Login', value: login });
    }
    if (userProfile.info.email) {
      this.info.push({ name: 'Email', value: userProfile.info.email });
    }
```

apps/consolidator/src/app/components/list-of-identities/list-of-identities.component.ts
```diff
@@ -0,0 +1,144 @@
...
+      }
+    }
+    const nickname = (userProfile['info'] as object)['nickname'] as string;
+    if (nickname) {
+      if (nickname !== '') {
+        return nickname;
+      }
+    }
+    return '';
+  }
```
bodnara, line 143:
```suggestion
  getNestedItem(userProfile: UserProfile, path: string[]): string {
    let currentItem = userProfile.info as object;
    for (const segment of path) {
      currentItem = currentItem[segment] as object;
      if (Array.isArray(currentItem)) {
        currentItem = currentItem[0] as object;
      }
    }
    return String(currentItem);
  }

  getUserName(userProfile: UserProfile): string {
    const name = userProfile.info.name;
    if (name) {
      return name;
    }
    const firstName = userProfile.info.given_name;
    const lastName = userProfile.info.family_name;
    if (firstName && lastName) {
      return firstName + ' ' + lastName;
    }
    return '';
  }

  getUserLogin(userProfile: UserProfile): string {
    const preferredUserName = userProfile.info.preferred_username;
    if (preferredUserName) {
      return preferredUserName;
    }
    const nickname = userProfile.info.nickname;
    if (nickname) {
      return nickname;
    }
    return '';
  }
```

bodnara (changes requested):

apps/linker/src/app/components/show-result/show-result.component.ts
```diff
@@ -0,0 +1,132 @@
...
+    let queryParams = location.search.substring(1);
+    if (this.linkerResult === 'IDENTITY_UNKNOWN') {
+      let index = 0;
+      let formAccessToken = 'yes';
+      while (formAccessToken !== '') {
+        index++;
+        formAccessToken = parseQueryParams('formerToken' + index.toString(), queryParams);
+      }
+      queryParams =
+        queryParams + '&formerToken' + index.toString() + '=' + this.oauthService.getAccessToken();
```
HejdaJakub, line 62:
```suggestion
      queryParams += '&formerToken' + index.toString() + '=' + this.oauthService.getAccessToken();
```

HejdaJakub (commented):
When I select one identity to connect and this identity is already linked, I cannot select another identity from the same IdP. I know that something similar (select account option) is mentioned in the user testing document, but I think it is not the same. In my case, I am not able to "log in" with another identity even if I close the dialog, so maybe just don't "caching" already used identities would be helpful? I think that it is more user-friendly when we allow users to type the same identity twice (and he/she will get the same response) rather than we don't allow them to type another identity from the same identity provider. But it can be part of the new task (which is in this Sprint).
image

libs/lib-linker/src/lib/consolidation-result/consolidation-result.component.ts
xkureck: So the only input this component have is `this.result` so the `ngOnChanges` method will happen when the result changes which is something what we want

bodnara: Yes, you are technically correct, as long as no other input is added.

bodnara (approved):

PR #1099, opened 2022-08-01
feat(profile): add advanced MFA settings

HejdaJakub:
* To the authentication page was added advanced MFA settings. Now is possible to select categories
and services (rps) which requires mfa.
* If enforce mfa is set to true, then all categories are selected and they are not allowed for disabling.
* If enforce mfa (for all services) is set to false, then user can select particular categories and services.

BREAKING CHANGE: For this functionality is required new oauh scope - authn_details in instance
config. Also new version of MFA API which supports this functionality is required.

### DIFF
--- a/apps/user-profile/src/app/pages/settings-page/settings-authorization/mfa-settings/mfa-settings.component.ts
+++ b/apps/user-profile/src/app/pages/settings-page/settings-authorization/mfa-settings/mfa-settings.component.ts
@@ -1,0 +1,428 @@
+/* eslint-disable
+   @typescript-eslint/no-explicit-any,
+   @typescript-eslint/explicit-module-boundary-types,
+   @typescript-eslint/no-unsafe-member-access,
+   @typescript-eslint/no-unsafe-assignment */
+
+import { Component, OnInit, ViewChild } from '@angular/core';
+import { now } from 'moment-timezone';
+import { AttributesManagerService } from '@perun-web-apps/perun/openapi';
+import { AuthService, MfaApiService, StoreService } from '@perun-web-apps/perun/services';
+import { OAuthService } from 'angular-oauth2-oidc';
+import { HttpClient } from '@angular/common/http';
+import { MatSlideToggle } from '@angular/material/slide-toggle';
+import { TranslateService } from '@ngx-translate/core';
+
+@Component({
+  selector: 'perun-web-apps-mfa-settings',
+  templateUrl: './mfa-settings.component.html',
+  styleUrls: ['./mfa-settings.component.scss'],
+})
+export class MfaSettingsComponent implements OnInit {
+  @ViewChild('toggle') toggle: MatSlideToggle;
+
+  mfaAvailable = false;
+  loadingMfa = false;
+
+  enforceMfa: boolean;
+  showDetail = false;
+  loadingCategories = false;
+  includeCategories: string[] = [];
+  excludeRps: string[] = [];
+  allCategories = false;
+  unchangedSettings = true;
+  unchangedEnforce = true;
+
+  categories = {};
+
+  constructor(
+    public translate: TranslateService,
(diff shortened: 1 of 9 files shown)


xkostka2 (changes requested):
After I changed the settings, clicked save, and then, after the page reloads save button is still enabled, I think it should be disabled after a successful save when there are no new changes.

Another thing, it is a bit inconvenient when I want to enable just one or two services from a category to be forced to manually disable all the other services within that category.

I don't know whether it should be part of this task or not, but some filtering options (search bar would be enough) would also be nice to have, in my opinion.

bodnara (changes requested):
I would actaully prefer the "manage token" button (currently bottom) to be on top. Mainly because if users do not have any token, they cannot interact with the settings. Also make the layout responsive.
image

Another thing is, as I mentioned, I would suggest using the `HttpClient` rather than `fetch`. However create a service for this purpose, components should not handle such logic.

Lastly I thing the whole part with MFA setting deserves to be a separate component.

apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
```diff
@@ -217,9 +206,314 @@ export class SettingsAuthenticationComponent implements OnInit, AfterViewInit {
...
+    sessionStorage.setItem('settings_mfa', body);
+  }
+
+  /**
+   * This method is used when we want to do some PUT operation, but we need new (not expired) id token
+   */
+  reAuthenticate(): void {
+    sessionStorage.setItem('mfa_route', '/profile/settings/auth');
+    localStorage.removeItem('refresh_token');
+    this.oauthService.logOut(true);
```
xkureck, line 424: since we are storing all the tokens in the localStorage the method logOut already remove refresh token
```suggestion
    this.oauthService.logOut(true);
```

xkureck (changes requested):
I agree with Adam that the MFA logic needs its own service and we should use the angular HTTP service instead of fetch

HejdaJakub:
Done:
- enable/disable button fixed
- toggles are not disabled now, just parent toggles react to the children's toggles
- "manage tokens" button moved above mfa settings
- fetch() replaced by HttpClient methods
- MFA setting moved to the new component

apps/user-profile/src/app/pages/settings-page/settings-authorization/mfa-settings/mfa-settings.component.ts
```diff
@@ -0,0 +1,428 @@
...
+  excludeRps: string[] = [];
+  allCategories = false;
+  unchangedSettings = true;
+  unchangedEnforce = true;
+
+  categories = {};
+
+  constructor(
+    private attributesManagerService: AttributesManagerService,
+    private translate: TranslateService,
```
bodnara, line 40: Since it is used in html
```suggestion
    public translate: TranslateService,
```

bodnara (commented):
TLDR: Works, but is ugly. 

I have checked this functionality in user profile and it seems to work, which is nice. However I must say that the mfa settings components is quite confusing. As we discussed we want methods to have single resposibility. Here it feels like each method has multiple side effects with different conditions.  Most methods call another method to form quite a long chain and it is almost imposible to navigate.

apps/user-profile/src/app/pages/settings-page/settings-authorization/mfa-settings/mfa-settings.component.ts
HejdaJakub: I agree that this is NOT the most beautiful component that I've ever seen :/. I tried to not duplicate the code and also not write infinity functions, but handling these nested toggles generated according to the "unstructured" object from API is not really good solution.

xkureck (changes requested):

PR #1105, opened 2022-08-08
fix(admin): cache multiple component instances

bodnara:
* components are now cached per page, before only one instance was cached and could result in mismatch between data and entity


HejdaJakub (changes requested):
There is still one inconsistency for the use case from the task in Jira: I am on `Facility1`, I click on `Facilities management` and select `Facility2`, then I click on `Allowed groups`. If I click on the back button to get back on `Facility1` and try to click on `Allowed groups` here, then the groups are correct for Facility1, but the selected entity switch to `Facility2` - in the side menu and in URL.

HejdaJakub (approved):

PR #1109, opened 2022-08-10
feat(admin): mark critical actions on attributes

HejdaJakub:
* In Perun admin -> Attribute definitions we can we can mark in the dialog READ/WRITE operations for
existing (or newly created) attributes as critical or uncritical.


libs/perun/services/src/lib/attribute-rights.service.ts
```diff
@@ -68,6 +73,23 @@ export class AttributeRightsService {
     };
   }
 
+  updateAttributeAction(
```
bodnara, line 76: As per my comment, this is probably not needed.

apps/admin-gui/src/app/shared/components/dialogs/create-attribute-definition-dialog/create-attribute-definition-dialog.component.ts
```diff
@@ -84,11 +88,30 @@ export class CreateAttributeDefinitionDialogComponent {
...
+          return zip(of(attDef.id), of(this.collections));
+        }),
         this.attributeRightsService.addAttributeId(),
         this.attributeRightsService.filterNullInPolicy(),
         switchMap((collections) =>
           this.attributeService.setAttributePolicyCollections({ policyCollections: collections })
+        ),
+        switchMap(() =>
+          this.attributeRightsService.updateAttributeAction(
+            this.finalReadOperations,
```
bodnara, line 102: I realized that if none of this calls depend on each other or have to be in order, we could use `mergeMap` to start all of them at once. If order is important then `concatMap` is probably better solution than chaining switch maps.

bodnara (commented):
~~If I understand correctly, the critical flag is updated only if it was truly changed. In my opinion this introduces a lot of "noise" in the code (duplicated vars, multiple checks ...) for negligible outcome.~~

EDIT: Turns out backend throws an error if same value is passed, so this is in fact necessary.

HejdaJakub:
According to a conversation with @bodnara, I just resolved conflicts in the last force push.

PR #1115, opened 2022-08-16
feat: refactor of login workflow

xkostka2:
* code regarding initial authentication has been moved to init-auth service and put into more compact form
* adjusted documentation

### DIFF
--- a/apps/publications/src/assets/config/defaultConfig.json
+++ b/apps/publications/src/assets/config/defaultConfig.json
@@ -17,6 +17,7 @@
     "en"
   ],
   "auto_auth_redirect": true,
+  "allow_service_access": true,
   "footer": {
     "columns": [
       {
--- a/apps/publications/src/assets/config/defaultConfig.json
+++ b/apps/publications/src/assets/config/defaultConfig.json
@@ -17,7 +17,6 @@
     "en"
   ],
   "auto_auth_redirect": true,
-  "allow_service_access": true,
   "footer": {
     "columns": [
       {
--- a/apps/user-profile/src/assets/config/defaultConfig.json
+++ b/apps/user-profile/src/assets/config/defaultConfig.json
@@ -7,6 +7,7 @@
   },
   "instance_favicon": false,
   "auto_auth_redirect": true,
+  "allow_service_access": true,
   "supported_languages": [
     "en"
   ],
--- a/apps/user-profile/src/assets/config/defaultConfig.json
+++ b/apps/user-profile/src/assets/config/defaultConfig.json
@@ -7,7 +7,6 @@
   },
   "instance_favicon": false,
   "auto_auth_redirect": true,
-  "allow_service_access": true,
   "supported_languages": [
     "en"
   ],
--- a/libs/perun/services/src/lib/auth.service.ts
+++ b/libs/perun/services/src/lib/auth.service.ts
@@ -35,7 +35,7 @@ export class AuthService {
   }
 
   loadOidcConfigData(): void {
-    this.oauthService.configure(this.getOidcClientConfig());
+    this.oauthService.configure(this.getClientConfig());
   }
 
   logout(): void {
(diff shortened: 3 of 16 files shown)


libs/perun/services/src/lib/init-auth.service.ts
```diff
@@ -106,38 +130,108 @@ export class InitAuthService {
...
         queryParams: queryParams,
         queryParamsHandling: 'merge',
       });
     }
   }
+
+  /**
+   *  Set up check whether access token expires, when it does, it gets automatically refreshed (if possible)
+   */
+  startRefreshToken(isLoggedIn: boolean): void {
```
xkureck, line 163: I dunno but I my opinion this method doesn't need isLoggedIn input - you can check right in the method if user is logged in

libs/perun/services/src/lib/init-auth.service.ts
```diff
@@ -106,38 +130,108 @@ export class InitAuthService {
...
     } else {
       this.setLoginScreen(true);
       localStorage.setItem('routeAuthGuard', window.location.pathname);
-      const query = location.search.substr(1).split('&');
+      const query = location.search.substring(1).split('&');
       const queryParams = {};
       for (const param of query) {
         const p = param.split('=');
         queryParams[p[0]] = p[1];
       }
```
xkureck, line 152: I am I weird or is this code totally useless
you can just on the this.router.navigate() use options of `queryParamsHandling: 'merge'` and not to do any of this?

libs/perun/services/src/lib/init-auth.service.ts
```diff
@@ -106,38 +130,108 @@ export class InitAuthService {
...
+  /**
+   * This method is used to handle oauth callbacks.
+   *
+   * First, it finishes the authentication and then redirects user to the url
+   * he wanted to visit.
+   *
+   */
+  handleAuthCallback(): Promise<boolean> {
+    return this.oauthService.loadDiscoveryDocumentAndTryLogin();
+  }
```
xkureck, line 188: also not used method

libs/perun/services/src/lib/auth.service.ts
```diff
@@ -3,14 +3,13 @@ import { filter } from 'rxjs/operators';
...
+import { AuthConfig, OAuthService } from 'angular-oauth2-oidc';
 import { parseQueryParams } from '@perun-web-apps/perun/utils';
 
 @Injectable({
   providedIn: 'root',
 })
 export class AuthService {
-  loggedIn = false;
   filterShortname: string;
   redirectUrl: string;
```
xkureck, line 14: check if this property is used, if not remove it

libs/perun/services/src/lib/auth.service.ts
```diff
@@ -166,10 +51,6 @@ export class AuthService {
     }
```
xkureck, line 51: remove from logout method `localStorage.removeItem('refresh_token');` because this is already done by the this.oauthService.logOut();

libs/perun/services/src/lib/auth.service.ts
```diff
@@ -184,24 +65,20 @@ export class AuthService {
...
-  handleAuthCallback(): Promise<boolean> {
-    return this.oauthService.loadDiscoveryDocumentAndTryLogin();
-  }
-
   redirectToOriginDestination(): Promise<boolean> {
     const mfaRoute = sessionStorage.getItem('mfa_route');
     if (mfaRoute) {
       return this.router.navigate([mfaRoute], { replaceUrl: true });
     }
+
```
xkureck, line 73: In this method (redirectToOriginDestination())
Is this pile of code:
    let params: string[] = [];
    if (storageParams) {
      params = storageParams.split('&');
    }
    const queryParams: Params = {};
    params.forEach((param) => {
      const elements = param.split('=');
      queryParams[elements[0]] = elements[1];
    });
useless? Because when you navigating you can just mergeQueryParams?

Also this pile of code:
if (queryParams['idpFilter']) {
      this.filterShortname = String(queryParams['idpFilter']);
    }
Is already done in the constructor of the service, so do we also need to do it here?

libs/perun/services/src/lib/auth.service.ts
```diff
@@ -244,96 +116,82 @@ export class AuthService {
...
+    const randomSalt = Math.random() * 0.25;
+    //Defines when the token_timeout event should be raised.
+    //Expiration of tokens was moved from 0.75(default) to random number from 0.5 to 0.75
+    //So the refreshing of the token is not triggered by multiple tabs at the same time
+    const timeoutFactor = 0.5 + randomSalt;
+
+    const customQueryParams = !filterValue ? {} : { acr_values: filterValue };
+
+    if (
+      (this.store.get('oidc_client', 'oauth_scopes') as string)
```
xkureck, line 134: you can use new method `getProperty`

libs/perun/services/src/lib/auth.service.ts
```diff
@@ -244,96 +116,82 @@ export class AuthService {
...
+        .split(' ')
+        .includes('offline_access') &&
+      this.store.get('oidc_client', 'oauth_offline_access_consent_prompt')
```
xkureck, line 137: you can use new method `getProperty`

libs/perun/services/src/lib/auth.service.ts
```diff
@@ -244,96 +116,82 @@ export class AuthService {
...
+      requestAccessToken: true,
+      issuer: this.store.get('oidc_client', 'oauth_authority') as string,
+      clientId: this.store.get('oidc_client', 'oauth_client_id') as string,
+      redirectUri: this.store.get('oidc_client', 'oauth_redirect_uri') as string,
+      postLogoutRedirectUri: this.store.get(
+        'oidc_client',
+        'oauth_post_logout_redirect_uri'
+      ) as string,
+      responseType: this.store.get('oidc_client', 'oauth_response_type') as string,
+      scope: this.store.get('oidc_client', 'oauth_scopes') as string,
```
xkureck, line 163: you can use new method `getProperty`

libs/perun/services/src/lib/auth.service.ts
```diff
@@ -244,96 +116,82 @@ export class AuthService {
...
-    if (localStorage.getItem('refresh_token') && !this.isLoggedIn()) {
-      return this.refreshAndStoreToken()
-        .then(() => Promise.resolve())
-        .catch(() => Promise.resolve());
-    } else {
-      return Promise.resolve();
+  private setIdpFilter(): string {
+    const queryParams = location.search.substring(1).split('&');
+    this.filterShortname = null;
+    const filters: Map<string, string> = this.store.get('oidc_client', 'filters') as Map<
```
xkureck, line 174: you can use new method `getProperty`

xkureck (changes requested):
image

Could we somehow get rid of this code in the method loadPrincipal and move it to some more appropriate place?
` const previousUrl = localStorage.getItem('routeAuthGuard');
          if (previousUrl) {
            localStorage.removeItem('routeAuthGuard');
            void this.router.navigate([previousUrl]);
          }`


In auth.service delete `private dialog: MatDialog,` from constructor

apps/publications/src/assets/config/defaultConfig.json
```diff
@@ -17,6 +17,7 @@
     "en"
   ],
   "auto_auth_redirect": true,
+  "allow_service_access": true,
```
HejdaJakub, line 20: Unused property, you can delete it.

apps/user-profile/src/assets/config/defaultConfig.json
```diff
@@ -7,6 +7,7 @@
   },
   "instance_favicon": false,
   "auto_auth_redirect": true,
+  "allow_service_access": true,
```
HejdaJakub, line 10: Unused property, you can delete it.

HejdaJakub (changes requested):
I am not able to test all use cases manually, because the login is broken. I see just infinite spinner and I can't log into the admin gui.

From a task definition, we want to have some property that allows us to "enforce" service access on a particular instance without typing `/service-access` to the URL - then you can probably use it in the `handleAuthStart()` method.

HejdaJakub (changes requested):
I tested all use cases from the task definition (excluding revoking access token and the refreshing of tokens - because we have still a bug here in our GUI ... and also excluding MFA in user-profile Authentication section, because there are some CORS errors now) and I cannot find any problems, so it looks good.

So just fix the comments below, please.

HejdaJakub (approved):

PR #1116, opened 2022-08-16
feat(admin): warning on consents hub page when global consent enforcing is disabled

xflord:
* if enforce_consents is false on the instance, display a warning informing the user that any changes on the consent hubs page will have no effect
* when consents globally disabled, disabled the 'Evaluate consents' button and added a tooltip

### DIFF
--- a/apps/admin-gui/src/app/admin/pages/admin-page/admin-consent-hubs/admin-consent-hubs.component.html
+++ b/apps/admin-gui/src/app/admin/pages/admin-page/admin-consent-hubs/admin-consent-hubs.component.html
@@ -1,16 +1,22 @@
 <div>
   <h1 class="page-subtitle">{{'ADMIN.CONSENT_HUBS.TITLE' | translate}}</h1>
-
+  <perun-web-apps-alert *ngIf="!globalForceConsents" alert_type="warn">
+    {{'ADMIN.CONSENT_HUBS.GLOBAL_DISABLED' | translate}}
+  </perun-web-apps-alert>
   <perun-web-apps-refresh-button (refresh)="refreshTable()"></perun-web-apps-refresh-button>
-  <button
-    *ngIf="authResolver.isPerunAdmin()"
-    (click)="evaluateConsents()"
-    [disabled]="selection.selected.length === 0"
-    color="accent"
-    class="action-button mr-2"
-    mat-flat-button>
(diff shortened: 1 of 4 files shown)


xflord:
Wasn't quite sure whether to retrieve the global consent enforcing property this way or create a new backend method that would return it. Got inspired by 'ConsentRelatedAttributePipe' where it's done this way.

apps/admin-gui/src/app/admin/pages/admin-page/admin-consent-hubs/admin-consent-hubs.component.html
```diff
@@ -1,11 +1,15 @@
...
+    {{'ADMIN.CONSENT_HUBS.GLOBAL_DISABLED' | translate}}
+  </perun-web-apps-alert>
   <perun-web-apps-refresh-button (refresh)="refreshTable()"></perun-web-apps-refresh-button>
   <button
     *ngIf="authResolver.isPerunAdmin()"
     (click)="evaluateConsents()"
-    [disabled]="selection.selected.length === 0"
+    [disabled]="!globalForceConsents || selection.selected.length === 0"
+    [matTooltipDisabled]="globalForceConsents"
+    matTooltip="{{'ADMIN.CONSENT_HUBS.TOOLTIP' | translate}}"
```
bodnara, line 12: The tooltip on disabled elements will not show up on some browsers. Our usuall workaround is to wrap the button with span and put the tooltip on that.

libs/perun/models/src/lib/ConfigProperties.ts
```diff
@@ -128,6 +128,7 @@ export interface PerunConfig {
   // Optional
   auto_auth_redirect?: boolean;
   supported_languages?: string[];
+  enforce_consents?: boolean;
```
bodnara, line 131: Please, also add this property to our config documentation here: https://perunaai.atlassian.net/wiki/spaces/STRIBOG/pages/3440714/Configuration+of+new+GUI+applications

bodnara (changes requested):
I think this solution is correct, just fix the tooltip.

libs/perun/models/src/lib/ConfigProperties.ts
xflord: done

bodnara (approved):

PR #1117, opened 2022-08-16
test: upgrade cypress to v10

xkureck:
* upgrade cypress to v10


apps/admin-gui-e2e/cypress.config.ts
```diff
@@ -0,0 +1,30 @@
...
+      BA_PASSWORD: 'test',
+      BA_USERNAME_VO_MANAGER: 'voManager',
+      BA_PASSWORD_VO_MANAGER: 'test',
+      BA_USERNAME_GROUP_MANAGER: 'groupManager',
+      BA_PASSWORD_GROUP_MANAGER: 'test',
+      BA_USERNAME_FACILITY_MANAGER: 'facilityManager',
+      BA_PASSWORD_FACILITY_MANAGER: 'test',
+      BA_USERNAME_RESOURCE_MANAGER: 'resourceManager',
+      BA_PASSWORD_RESOURCE_MANAGER: 'test',
+      BA_PASSWORD_TOP_GROUP_CREATOR: 'topGroupCreator',
```
HejdaJakub, line 25: **Not a part of your PR, so ignore it.**
We should use USERNAME/PASSWORD for consistency, we should use it also in the test file, but for this, we need to change `perun.passwd` file in gitlab repo, so we can change it further.
```suggestion
      BA_USERNAME_TOP_GROUP_CREATOR: 'topGroupCreator',
      BA_PASSWORD_TOP_GROUP_CREATOR: 'test',
```

HejdaJakub (changes requested):
Just one things:
- top-group-creator-perun_admin.**spec**.js -> top-group-creator-perun_admin.**cy**.js (rename it like all other tests please, otherwise this file is skipped)
image

apps/admin-gui-e2e/cypress.config.ts
xkureck: I agree

HejdaJakub (approved):

PR #1121, opened 2022-08-17
feat(linker): consolidator user testing feedback

HejdaJakub:
* Forkflow changed - when user tries to send message to user support, new dialog will be opened. In
this dialog is textarea with general message which can be changed by the user. This message will be merged with some technical info before sending to user support.
* Some texts were too.

### DIFF
--- a/apps/linker/src/app/components/send-message-dialog/send-message-dialog.component.html
+++ b/apps/linker/src/app/components/send-message-dialog/send-message-dialog.component.html
@@ -1,0 +1,30 @@
+<div>
+  <div mat-dialog-title>
+    <h1 class="page-subtitle">
+      {{'SHARED_LIB.CONSOLIDATOR.SEND_MESSAGE_TO_SUPPORT_DIALOG.TITLE' | translate}}
+    </h1>
+  </div>
+  <mat-spinner *ngIf="loading" class="mr-auto ml-auto"></mat-spinner>
+  <div *ngIf="!loading" mat-dialog-content>
+    <perun-web-apps-alert alert_type="info">
+      {{'SHARED_LIB.CONSOLIDATOR.SEND_MESSAGE_TO_SUPPORT_DIALOG.SUBTITLE' | translate}}
+    </perun-web-apps-alert>
+    <mat-form-field class="w-100">
+      <textarea
+        [(ngModel)]="message"
+        matInput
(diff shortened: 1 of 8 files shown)


apps/linker/src/app/components/send-message-dialog/send-message-dialog.component.html
```diff
@@ -0,0 +1,30 @@
+<div>
+  <div mat-dialog-title>
+    <h2 class="page-subtitle">
```
xkureck, line 3: add css style for the page-subtitle class - in linker it is not defined

libs/lib-linker/src/lib/consolidation-result/consolidation-result.component.html
```diff
@@ -12,9 +12,16 @@
       <h1>
         {{('SHARED_LIB.CONSOLIDATOR.' + result + '_TITLE') | translate}}
       </h1>
-      <h5>
+      <h5 *ngIf="!(userMail && result === 'OK')">
         {{('SHARED_LIB.CONSOLIDATOR.' + result + '_SUBTITLE') | translate}}
-        <span *ngIf="result === 'UNKNOWN_ERROR'"> {{supportMail}} </span>
+        <span *ngIf="result === 'UNKNOWN_ERROR' || result === 'MESSAGE_SENT_TO_SUPPORT'">
+          {{supportMail}}
+        </span>
```
xkureck, line 19: oh sorry, you misunderstood me
I wanted to show the user to what HIS email will support contact him - so not the perun@cesnet.cz but -> eg. luckureckova@gmail.com

xkureck (changes requested):
image

apps/linker/src/app/components/send-message-dialog/send-message-dialog.component.html
HejdaJakub: Css class added.

libs/lib-linker/src/lib/consolidation-result/consolidation-result.component.html
HejdaJakub: Oh my bad, I changed the email and text for this use case (MESSAGE_SENT_TO_SUPPORT).

xkureck (approved):

PR #1124, opened 2022-08-19
feat(admin): e2e tests for TRUSTEDFACILITYADMIN

xflord:
* Added e2e tests for the TRUSTEDFACILITYADMIN role.

### DIFF
--- a/apps/admin-gui-e2e/cypress.config.ts
+++ b/apps/admin-gui-e2e/cypress.config.ts
@@ -23,6 +23,8 @@ export default defineConfig({
       BA_USERNAME_RESOURCE_MANAGER: 'resourceManager',
       BA_PASSWORD_RESOURCE_MANAGER: 'test',
       BA_PASSWORD_TOP_GROUP_CREATOR: 'topGroupCreator',
+      BA_USERNAME_TRUSTED_FACILITY_ADMIN: 'trustedFacilityAdmin',
+      BA_PASSWORD_TRUSTED_FACILITY_ADMIN: 'test',
     },
     specPattern: 'src/e2e/**/*.cy.{js,jsx,ts,tsx}',
     supportFile: false,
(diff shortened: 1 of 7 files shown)


apps/admin-gui-e2e/src/integration/trusted-facility-admin-perun_admin.spec.js
```diff
@@ -0,0 +1,259 @@
...
+      .wait('@getGroupAssignments')
+
+      //  assert that group was removed
+      .get(`[data-cy=${dbGroupToRemove}-checkbox]`)
+      .should('not.exist')
+  })
+
+  it('test deactivate group assignment', () => {
+
+    cy.intercept('**/resourcesManager/getGroupAssignments**').as('getGroupAssignments')
```
HejdaJakub, line 94: Again move the interception further in the test - for example before `.wait()`.

apps/admin-gui-e2e/src/integration/trusted-facility-admin-perun_admin.spec.js
```diff
@@ -0,0 +1,259 @@
...
+      .get('[data-cy=remove-manager-button-dialog]')
+      .click()
+      .wait('@getRichAdmins')
+
+      // assert that manager was removed
+      .get(`[data-cy=${dbUserToRemoveRoles}-checkbox]`)
+      .should('not.exist')
+  })
+
+  it('test remove resource observer', () => {
```
HejdaJakub, line 209: You should change the role filter at the end of this test because now you are checking that `Test8 User17` is not Resource Admin NOT Resource Observer. After that, you should move intercept() and wait() functions after the role filter change.

apps/admin-gui-e2e/src/integration/trusted-facility-admin-perun_admin.spec.js
```diff
@@ -0,0 +1,259 @@
...
+      .get('[data-cy=remove-manager-button-dialog]')
+      .click()
+      .wait('@getRichAdmins')
+
+      // assert that manager was removed
+      .get(`[data-cy=${dbUserToRemoveRoles}-checkbox]`)
+      .should('not.exist')
+  })
+
+  it('test remove resource self service', () => {
```
HejdaJakub, line 233: The same, you are checking Resource Admin role NOT Resource self service role.

HejdaJakub (changes requested):
Rebase please:
- Due to cypress v10 the config file is changed (cypress.json deleted and cypress.config.ts added) - move the username and password to this file, please
- In this new version is also required another naming of the test files, so rename it, please: trusted-facility-admin-perun_admin.**spec**.js -> trusted-facility-admin-perun_admin.**cy**.js

PR #1128, opened 2022-08-26
feat(admin): update of the Table-Checkbox service

mattjoke:
* updated TableCheckbox Service
* merged isAllSelected and isAllSelected isAllSelectedWithDisabledCheckbox
* changed the call for the service to be used
  in a pipe

### DIFF
--- a/apps/admin-gui/src/app/shared/components/blacklist-list/blacklist-list.component.html
+++ b/apps/admin-gui/src/app/shared/components/blacklist-list/blacklist-list.component.html
@@ -15,13 +15,15 @@
       matSortActive="id"
       matSortDirection="asc"
       matSortDisableClear>
-      <ng-container matColumnDef="select">
+      <ng-container
+        matColumnDef="select"
+        *ngIf="(dataSource | isAllSelected: selection.selected.length) as selected">
         <th *matHeaderCellDef class="align-checkbox" mat-header-cell>
           <mat-checkbox
             (change)="$event ? masterToggle() : null"
             [aria-label]="checkboxLabel()"
-            [checked]="selection.hasValue() && isAllSelected()"
-            [indeterminate]="selection.hasValue() && !isAllSelected()"
+            [checked]="selection.hasValue() && selected"
+            [indeterminate]="selection.hasValue() && !selected"
             color="primary">
           </mat-checkbox>
         </th>
--- a/libs/perun/pipes/src/lib/is-all-selected.pipe.ts
+++ b/libs/perun/pipes/src/lib/is-all-selected.pipe.ts
@@ -1,0 +1,18 @@
+import { MatTableDataSource } from '@angular/material/table';
+import { TableCheckbox } from '@perun-web-apps/perun/services';
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'isAllSelected',
+})
+export class IsAllSelectedPipe implements PipeTransform {
+  constructor(private tableCheckboxService: TableCheckbox) {}
+
+  transform<T>(
+    ds: MatTableDataSource<T>,
+    numberSelected: number,
+    canBeSelected?: (item: T) => boolean
+  ): boolean {
+    return this.tableCheckboxService.isAllSelected(numberSelected, ds, canBeSelected);
+  }
+}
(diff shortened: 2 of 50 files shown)


apps/admin-gui/src/app/shared/components/blacklist-list/blacklist-list.component.html
```diff
@@ -19,8 +19,8 @@
           <mat-checkbox
             (change)="$event ? masterToggle() : null"
             [aria-label]="checkboxLabel()"
-            [checked]="selection.hasValue() && isAllSelected()"
-            [indeterminate]="selection.hasValue() && !isAllSelected()"
+            [checked]="selection.hasValue() && (dataSource | isAllSelected: selection.selected.length :canBeSelected)"
+            [indeterminate]="selection.hasValue() && !(dataSource | isAllSelected: selection.selected.length :canBeSelected)"
```
HejdaJakub, line 23: Remove no existing method `canBeSelected`. This problem is in many places, so fix it in all HTML files, where this problem occurs, please.
image

```suggestion
            [checked]="selection.hasValue() && (dataSource | isAllSelected: selection.selected.length)"
            [indeterminate]="selection.hasValue() && !(dataSource | isAllSelected: selection.selected.length)"
```

libs/perun/pipes/src/lib/is-all-selected.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+@Pipe({
+  name: 'isAllSelected',
+})
+export class IsAllSelectedPipe implements PipeTransform {
+  constructor(private tableCheckboxService: TableCheckbox) {}
+
+  transform(
+    ds: MatTableDataSource<unknown>,
+    numberSelected: number,
+    canBeSelected: (T) => boolean
```
HejdaJakub, line 14: Make this param optional, so you can skip it in HTML files, where this method does not exist.
```suggestion
    canBeSelected?: (T) => boolean
```

HejdaJakub (changes requested):
I don't know why, but replacing methods in HTML with pipes didn't fix the problem. We can discuss it on a daily standup.

libs/perun/pipes/src/lib/is-all-selected.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+export class IsAllSelectedPipe implements PipeTransform {
+  constructor(private tableCheckboxService: TableCheckbox) {}
+
+  transform(
+    ds: MatTableDataSource<unknown>,
+    numberSelected: number,
+    canBeSelected?: (T) => boolean
+  ): boolean {
+    return this.tableCheckboxService.isAllSelected(numberSelected, ds, canBeSelected);
+  }
```
bodnara, line 17:
```suggestion
  transform<T> (
    ds: MatTableDataSource<T>,
    numberSelected: number,
    canBeSelected?: (item: T) => boolean
  ): boolean {
    return this.tableCheckboxService.isAllSelected(numberSelected, ds, canBeSelected);
  }
```

libs/perun/services/src/lib/table-checkbox.service.ts
```diff
@@ -17,26 +17,38 @@ export class TableCheckbox {
...
+  isAllSelected<T>(
     rowsSelected: number,
-    filter: string,
-    pageSize: number,
-    nextPage: boolean,
-    pageIndex: number,
     dataSource: MatTableDataSource<T>,
-    sort: MatSort,
-    canBeSelected: (T) => boolean
+    canBeSelected?: (T) => boolean
```
bodnara, line 23: If we provide a default select function that just returns true, there is no need for the `if(!canBeSelected)` block I believe.
```suggestion
    canBeSelected: (item: T) => boolean = (): boolean => true
```

bodnara (changes requested):
Some tables have console errors and some do not show properly at all, I did not go through all of them but e.g. the member-list.

Also perhaps it was not specified properly but the masterToggle function can have the number of parameters reduces as well.

bodnara (approved):
Nice job on this one.

HejdaJakub (approved):
I have bad news guys because I tested it and the function `isAllSelected()` from the tableCheckbox service is still called many times on every mouse move across the table. Then I found out that we call in these HTML files function `checkboxLabel()`, which calls `isAllSelected()` function and here we go again. But I believe that this part deserves a separate task and a separate PR so that I can approve this PR.

PR #1131, opened 2022-09-07
feat: catch mfa required error and handle mfa

HejdaJakub:
* Now our applications catch new MFA required error.
* If this error occurs, then the first dialog is opened that tells the user, that MFA is required.
* If the user wants to proceed with additional verification, then is opened a new window, which handles MFA.
* Original window is locked until the second window is opened.

### DIFF
--- a/libs/perun/multi-factor-authentication/src/lib/mfa-required-dialog/mfa-required-dialog.component.html
+++ b/libs/perun/multi-factor-authentication/src/lib/mfa-required-dialog/mfa-required-dialog.component.html
@@ -1,0 +1,14 @@
+<h1 mat-dialog-title>{{'SHARED_LIB.PERUN.COMPONENTS.MFA_REQUIRED_DIALOG.TITLE' | translate}}</h1>
+<div class="dialog-container" mat-dialog-content>
+  <perun-web-apps-alert alert_type="warn">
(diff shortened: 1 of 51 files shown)


xkostka2 (approved):

libs/perun/multi-factor-authentication/src/lib/mfa-required-dialog/mfa-required-dialog.component.html
```diff
@@ -0,0 +1,14 @@
+<h1 mat-dialog-title>{{'SHARED_LIB.PERUN.COMPONENTS.MFA_REQUIRED_DIALOG.TITLE' | translate}}</h1>
+<div class="dialog-container" mat-dialog-content>
+  <perun-web-apps-alert alert_type="warn">
+    {{'SHARED_LIB.PERUN.COMPONENTS.MFA_REQUIRED_DIALOG.INFO' | translate}}
+  </perun-web-apps-alert>
+</div>
+<div mat-dialog-actions>
+  <button (click)="onCancel()" class="ml-auto" mat-flat-button>
```
bodnara, line 8: I believe one of angular stylistic suggestions is to not prefix with on.

bodnara (changes requested):
Overall very well done. Just some small things to change.

bodnara (approved):

PR #1133, opened 2022-09-13
feat(admin): e2e tests for RESOURCEOBSERVER

xflord:
* Added e2e tests for the RESOURCEOBSERVER role


HejdaJakub (approved):
If we want to skip these tests at this moment due to insufficient rights:
- view resources on facility
- view assigned members
- view assigned groups

... then I can approve this PR. So just comment these tests and unused constants, please. Other tests look good.

HejdaJakub (approved):

PR #1135, opened 2022-09-14
fix(admin): set service to entity storage service

bodnara:
* service is now correctly saved on service detail page
* other pages on service detail can now get the service from entity storage


HejdaJakub (changes requested):
You can then delete `this.entityStorageService.setEntity({ id: service.id, beanName: service.beanName });` in refresh() method.

mattjoke (approved):

HejdaJakub (approved):

PR #1136, opened 2022-09-14
fix(admin): add missing refresh button

bodnara:
* added refresh button to notification pages (vo, group)
* refresh button on ext sources page was changed to be consistent


HejdaJakub (changes requested):
I think that stakeholders told that they want the refresh button also on the Perun admin -> Users page. It's the only place where I miss the refresh button.

HejdaJakub (approved):

PR #1140, opened 2022-09-15
feat(admin): scroll to top of table on page change

bodnara:
* changing page from bottom paginator scrolls user to top of table

### DIFF
--- a/libs/perun/utils/src/lib/table-wrapper/table-wrapper.component.ts
+++ b/libs/perun/utils/src/lib/table-wrapper/table-wrapper.component.ts
@@ -1,4 +1,12 @@
-import { Component, EventEmitter, Input, OnInit, Output, ViewChild } from '@angular/core';
+import {
+  Component,
+  ElementRef,
+  EventEmitter,
+  Input,
+  OnInit,
+  Output,
+  ViewChild,
+} from '@angular/core';
 import { MatPaginator, PageEvent } from '@angular/material/paginator';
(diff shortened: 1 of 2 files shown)


libs/perun/utils/src/lib/table-wrapper/table-wrapper.component.ts
```diff
@@ -46,5 +55,6 @@ export class TableWrapperComponent implements OnInit {
     if (this.tableId) {
       this.tableConfigService.setTablePageSize(this.tableId, event.pageSize);
     }
+    this.topNav.nativeElement.scrollIntoView(false); // scroll to top of table on page changes
```
HejdaJakub, line 58:
```suggestion
    this.topNav.nativeElement.scrollIntoView(true); // scroll to top of table on page changes
```
I would prefer the `true` parameter because if you have more tables on one page, it looks better with `true` option IMHO (you can test it on /myProfile/roles -> Membership section).

Maybe there is one better option that looks the best for me. 
```suggestion
    // scroll to top of table on page changes
    this.topNav.nativeElement.scrollIntoView({
      behavior: 'smooth',
      block: 'start',
      inline: 'nearest',
    });
```

bodnara: I like this option the most  as well. Initialy I thought scrolling would be annoying with 100 items per page but the behaviour is truly "smooth".

HejdaJakub (approved):

PR #1144, opened 2022-09-23
feat(admin): show backend version

mattjoke:
* the footer contains information about latest backend version

### DIFF
--- a/libs/perun/components/src/lib/perun-footer/perun-footer.component.html
+++ b/libs/perun/components/src/lib/perun-footer/perun-footer.component.html
@@ -60,15 +60,21 @@
         class="text-decoration-none"
         target="_blank"
         >{{index ? ', ' + item.name : item.name}}</a
-      >
-    </span>
-    Version
+      > </span
+    >, web application:
     <a
       [href]="githubRepository"
       [ngStyle]="{'color': copyrightTextColor}"
       class="text-decoration-none"
       target="_blank"
-      >{{version}}</a
+      >{{guiVersion}}</a
+    >, server:
+    <a
+      [href]="githubBackendRepository"
+      [ngStyle]="{'color': copyrightTextColor}"
+      class="text-decoration-none"
+      target="_blank"
+      >{{backendVersion}}</a
     >
   </div>
 </footer>
--- a/libs/perun/components/src/lib/perun-footer/perun-footer.component.ts
+++ b/libs/perun/components/src/lib/perun-footer/perun-footer.component.ts
@@ -9,6 +9,7 @@ import { StoreService } from '@perun-web-apps/perun/services';
 import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
 import { TranslateService } from '@ngx-translate/core';
 import { TranslatedElem } from '@perun-web-apps/perun/pipes';
+import { UtilsService } from '@perun-web-apps/perun/openapi';
 declare let require: any;
 
 interface FooterElement extends TranslatedElem {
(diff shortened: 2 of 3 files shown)


libs/perun/components/src/lib/perun-footer/perun-footer.component.ts
```diff
@@ -55,14 +56,19 @@ export class PerunFooterComponent implements OnInit {
...
     private translateService: TranslateService,
+    private utilsService: UtilsService,
     private dialog: MatDialog
   ) {}
 
   ngOnInit(): void {
     this.translateService.onLangChange.subscribe((lang) => {
       this.language = lang.lang;
     });
-    this.version = require('../../../../../../package.json').version as string;
```
xflord, line 65: It's a bit confusing that now we display the version of back-end, however the link in the footer still leads to GUI releases. I would probably update the version in package.json and then display both the version of GUI and BE.

mattjoke: Yeah, you're right about the links. The following commit addresses this; the footer will now display both information (i.e. GUI and BE version) and their respective links

xflord (commented):
Good job, a few more visual details - I think the footer would look better without spaces between 'v.' and the version numbers. Also, 'RPC' instead of 'Backend' is a better fit, I believe. What do @HejdaJakub and @bodnara think?

libs/perun/components/src/lib/perun-footer/perun-footer.component.html
```diff
@@ -62,13 +62,21 @@ <h6 [ngStyle]="{'color': headersTextColor}" class="font-weight-bold">
         >{{index ? ', ' + item.name : item.name}}</a
       >
     </span>
-    Version
+    , GUI: v.
```
xkureck, line 65:
```suggestion
    , web application: v.
```

libs/perun/components/src/lib/perun-footer/perun-footer.component.html
```diff
@@ -62,13 +62,21 @@ <h6 [ngStyle]="{'color': headersTextColor}" class="font-weight-bold">
...
     <a
       [href]="githubRepository"
       [ngStyle]="{'color': copyrightTextColor}"
       class="text-decoration-none"
       target="_blank"
-      >{{version}}</a
+      >{{guiVersion}}</a
+    >
+    , Backend: v.
```
xkureck, line 73:
```suggestion
    , core: v.
```

xkureck (commented):
more user friendly, users dont know what gui or backend is

libs/perun/components/src/lib/perun-footer/perun-footer.component.html
HejdaJakub, line 73: There are suggestions like `Backend`,  `RPC` or `core`, so I also would like to suggest my option. I am not sure what is commonly used across applications, but the `server` is used at least in our old GUI, so our users should be familiar with it.
```suggestion
    , server: v.
```

libs/perun/components/src/lib/perun-footer/perun-footer.component.ts
```diff
@@ -46,23 +47,37 @@ export class PerunFooterComponent implements OnInit {
...
+
+    this.guiVersion = require('../../../../../../package.json').version as string;
+    this.utilsService.getPerunStatus().subscribe((val) => {
+      const versionString = val[0];
+      this.backendVersion = versionString.substring(versionString.indexOf(':') + 2);
+      this.backendVersion = '17.2.0';
+      this.guiVersion = '17.2.0';
```
HejdaJakub, line 78: You probably forgot to delete these two lines.
```suggestion
```

mattjoke: Yes, thank you for pointing it out.

HejdaJakub (approved):

PR #1145, opened 2022-09-26
fix(admin): facility resources page double load

xflord:
* fixed a bug where the facility assigned resources table would refresh on page open


HejdaJakub (changes requested):
image
There is an error now due to `loading` value inconsistency before and after view initialization. You can easily fix it with ChangeDetectorRef.
```
ngAfterViewInit(): void {
    this.cd.detectChanges();
}
```

HejdaJakub (approved):

PR #1147, opened 2022-09-27
fix: replace get() with getProperty()

sarkapalkovicova:
All usages of get() method were replaced by calling getProperty() method.

### DIFF
--- a/libs/perun/components/src/lib/perun-footer/perun-footer.component.ts
+++ b/libs/perun/components/src/lib/perun-footer/perun-footer.component.ts
@@ -8,26 +8,9 @@ import { ReportIssueDialogComponent } from '../report-issue-dialog/report-issue-
 import { StoreService } from '@perun-web-apps/perun/services';
 import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
 import { TranslateService } from '@ngx-translate/core';
-import { TranslatedElem } from '@perun-web-apps/perun/pipes';
 import { UtilsService } from '@perun-web-apps/perun/openapi';
+import { CopyrightItem, FooterColumn } from '@perun-web-apps/perun/models';
 declare let require: any;
-
-interface FooterElement extends TranslatedElem {
-  logo?: string;
-  icon?: string;
-  dialog?: string;
-  link_en: string;
-}
-
-interface FooterColumn extends TranslatedElem {
-  elements: FooterElement[];
-  logos: string[];
-}
-
-interface FooterCopyrightItems {
-  name: string;
-  url: string;
-}
 
 @Component({
   selector: 'perun-web-apps-footer',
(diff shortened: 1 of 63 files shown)


HejdaJakub (changes requested):
You can also delete `deprecated` get(...) method from `store.service.ts`.

libs/perun/login/src/lib/login-screen-base/login-screen-base.component.ts
```diff
@@ -25,11 +25,11 @@ export class LoginScreenBaseComponent implements OnInit {
...
   ngOnInit(): void {
-    this.headerBackgroundColor = this.storeService.get('theme', 'nav_bg_color') as string;
-    this.headerTextColor = this.storeService.get('theme', 'nav_text_color') as string;
-    this.logo = this.sanitizer.bypassSecurityTrustHtml(this.storeService.get('logo') as string);
+    this.headerBackgroundColor = this.storeService.getProperty('theme').nav_bg_color;
+    this.headerTextColor = this.storeService.getProperty('theme').nav_text_color;
+    this.logo = this.sanitizer.bypassSecurityTrustHtml(this.storeService.getProperty('logo'));
     this.textColor = this.headerTitle
-      ? (this.storeService.get('theme', 'header_text_color') as string)
+      ? this.storeService.getProperty('theme').header_text_color
```
HejdaJakub, line 32: You can delete `header_text_color` from `ConfigProperties.ts`.
```suggestion
      ? this.storeService.getProperty('theme').nav_text_color
```

libs/perun/components/src/lib/perun-footer/perun-footer.component.ts
```diff
@@ -8,51 +8,29 @@ import { ReportIssueDialogComponent } from '../report-issue-dialog/report-issue-
...
-    'footer',
-    'github_backend_releases'
-  ) as string;
-  iconColor: string = this.storeService.get('theme', 'footer_icon_color') as string;
-  bgColor: string = this.storeService.get('theme', 'footer_bg_color') as string;
+  headersTextColor: string = this.storeService.getProperty('theme').footer_headers_text_color;
+  linksTextColor: string = this.storeService.getProperty('theme').footer_links_text_color;
+  githubRepository: string = this.storeService.getProperty('footer').github_releases;
+  githubBackendRepository: string = this.storeService.getProperty('footer').github_backend_releases;
+  iconColor: string = this.storeService.getProperty('theme').footer_icon_color;
```
HejdaJakub, line 31: I couldn't find any history of the `footer_icon_color` property. Variable `iconColor` is NOT used in HTML, so I think you can delete this line and the property `footer_icon_color` can be deleted from `ConfigProperties.ts`.
```suggestion
```

libs/perun/services/src/lib/store.service.ts
```diff
@@ -62,73 +62,41 @@ export class StoreService {
...
-          }
-          currentValue = currentValue[keys[i]];
-        }
+  addMissingValuesToProperty<T extends keyof PerunConfig>(
+    value: PerunConfig[T],
+    defaultValue: PerunConfig[T]
+  ): PerunConfig[T] {
+    if (typeof value === 'object' && !Array.isArray(value) && value) {
+      for (const key of Object.keys(defaultValue)) {
+        defaultValue[key] = this.addMissingValuesToProperty(value[key], defaultValue[key]);
```
HejdaJakub, line 95: I think you can use something like this to fix eslint errors.
```suggestion
        defaultValue[key] = this.addMissingValuesToProperty(
           value[key] as PerunConfig[T], 
           defaultValue[key] as PerunConfig[T]
        );
```

libs/perun/services/src/lib/store.service.ts
```diff
@@ -62,73 +62,41 @@ export class StoreService {
...
-            // console.error('Missing value in default config: ' + keys);
-            break;
-          }
-          currentValue = currentValue[keys[i]];
-        }
+  addMissingValuesToProperty<T extends keyof PerunConfig>(
+    value: PerunConfig[T],
+    defaultValue: PerunConfig[T]
+  ): PerunConfig[T] {
+    if (typeof value === 'object' && !Array.isArray(value) && value) {
```
HejdaJakub, line 93: There is a little "problem" with an array of objects properties like the `custom_labels` property. On the other hand properties of this type are just about labels (`ProfileAttribute[]`, `ProfileExtService[]` and `ProfileCustomLabel[]`) so I don't think that it is a big problem at all. IMHO it's more about a bad configuration, if someone put just a part of this config type to the `instanceConfig`, what do you think (@xkostka2)?

HejdaJakub (changes requested):

libs/perun/services/src/lib/store.service.ts
bodnara: I already said this to @sarkapalkovicova but the issue here is that on consecutive passes the values are no longer item of PerunConfig but some nested item. So the definition of function has to be changed.

libs/perun/services/src/lib/store.service.ts
xkostka2: I agree, default values in array-of-objects properties are usually just examples of how to fill those properties.

bodnara (approved):

PR #1150, opened 2022-10-03
feat: add ssh keys and certificates to admin-gui

HejdaJakub:
* My Profile - Authentication section was refactored. There is new submenu with two new sections (SSH Keys and Certificates).
* Component SSH Keys was already used in user-profile, so this component was extracted to to libs folder and reused in admin-gui.
* Page Certificates is created as a MVP, because we are not sure yet, if this page will be used by users (but now we want to move this functionality to the new gui from old gui). This page now displays table with only one attribute. Value of the certificate isn't displayed user-friendly, but as a MVP it should be sufficient.


libs/perun/components/src/lib/identities-page/identities-page.component.ts
```diff
@@ -129,12 +137,16 @@ export class IdentitiesPageComponent implements OnInit {
           location.reload();
         } else if (result === 'OK') {
           this.notificator.showSuccess(
-            this.translate.instant('IDENTITIES.SUCCESSFULLY_ADDED') as string
+            this.translate.instant(
```
bodnara, line 140: Use the perun translate service instead.

libs/perun/components/src/lib/settings-ssh-keys/settings-ssh-keys.component.ts
```diff
@@ -43,33 +43,37 @@ export class SettingsSSHKeysComponent implements OnInit {
...
     translateService
-      .get('SSH_KEYS.HEADER_COLUMN')
+      .get('SHARED_LIB.PERUN.COMPONENTS.SSH_KEYS.HEADER_COLUMN')
       .subscribe((value: string) => (this.headerColumnText = value));
   }
 
   ngOnInit(): void {
-    this.userId = this.store.getPerunPrincipal().userId;
+    if (window.location.pathname.startsWith('/myProfile')) {
+      this.userId = Number(window.location.pathname.split('/')[3]);
```
bodnara, line 61: Same question as before, can we use entity storage ?

bodnara (changes requested):
When I use add button on the certificats page in admin-gui, it redirect to page not found.

libs/perun/components/src/lib/identities-page/identities-page.component.ts
```diff
@@ -53,8 +54,15 @@ export class IdentitiesPageComponent implements OnInit {
   ) {}
 
   ngOnInit(): void {
-    this.userId = this.storage.getPerunPrincipal().userId;
-    this.displayCertificates = this.storage.get('display_identity_certificates') as boolean;
+    if (window.location.pathname.startsWith('/myProfile')) {
+      this.userId = Number(window.location.pathname.split('/')[3]);
```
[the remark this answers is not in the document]
HejdaJakub: The changes on the identities page were restored due to another final solution of the certificates page.

libs/perun/components/src/lib/identities-page/identities-page.component.ts
HejdaJakub: The changes on the identities page were restored due to another final solution of the certificates page.

libs/perun/components/src/lib/settings-ssh-keys/settings-ssh-keys.component.ts
HejdaJakub: Good point. Fixed.

bodnara (approved):

PR #1151, opened 2022-10-04
feat(admin): fix non-existing form

mattjoke:
* fixed infinite loader when form does not exist


HejdaJakub (commented):
Just remove one unused import in application-form-copy-items-dialog.component.ts, please.
`HttpErrorResponse' is defined but never used`
Looks good otherwise.

HejdaJakub (approved):

PR #1152, opened 2022-10-05
feat(admin): add mailing lists opt-out and data quota settings pages to service account overview

xflord:
* added page for mailing lists opt-out for service identities
* added page for data quota management for service identities
* refactored some components


apps/user-profile/src/app/app.module.ts
```diff
@@ -82,6 +79,9 @@ import { SettingsLocalAccountComponent } from './pages/settings-page/settings-lo
 import { ActivateLocalAccountDialogComponent } from './components/dialogs/activate-local-account-dialog/activate-local-account-dialog.component';
 import { PerunNamespacePasswordFormModule } from '@perun-web-apps/perun/namespace-password-form';
 import { MfaSettingsComponent } from './pages/settings-page/settings-authorization/mfa-settings/mfa-settings.component';
+import { PerunDialogsModule } from '@perun-web-apps/perun/dialogs';
```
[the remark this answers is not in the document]
xflord: Turns out the module doesn't need to be imported. A bunch of errors were thrown before during the build, which I then fixed with importing the module, but I guess it was also fixed by something else I did afterwards as well (indexing perhaps?).
Removed the import.

bodnara (approved):

libs/perun/components/src/lib/settings-mailing-lists/mailing-lists.component.ts
```diff
@@ -0,0 +1,246 @@
...
+      this.getOptOutAttribute(resource);
+    }
+    if (this.changeOptOut) {
+      if (this.changeOptOut === 'subscribe') {
+        this.subscribe();
+      } else if (this.changeOptOut === 'unsubscribe') {
+        this.unsubscribe();
+      }
+      this.changeOptOut = null;
+    }
```
HejdaJakub, line 95: I would move this block of code to the if the block below (`if (!this.isService){...}`) because this way it would be possible to subscribe/unsuscribe the mailing list via the URL, but I think we don't want it in the admin GUI at this time.

apps/user-profile/src/assets/i18n/en.json
```diff
@@ -223,16 +223,6 @@
       "CLOSE": "Close",
       "COPY": "Copy"
     },
-    "REQUEST_DATA_QUOTA_CHANGE": {
```
HejdaJakub, line 226: You forgot to delete also duplication of `DATA_QUOTAS` and `OPT_OUT_MAILING_LISTS` in this file.

HejdaJakub (changes requested):
You should add these sections also to the side menu.

NOTE: In the task definition was written that we want to add these sections to the `Advanced Settings`, but I think it looks good on this "level", so for me, it's not a problem.

HejdaJakub (commented):
I am not sure what is going on here, but if I click on `Mailing lists opt-out` in the side menu, and then when I click on another item in this side menu (for service user `Data quotas`, `Authentication` etc.), the activated regex is somehow broken for this first click - the background of the selected item doesn't reflect the changes, but the whole pages (including URL) are correctly changed.

xflord:
> I am not sure what is going on here, but if I click on `Mailing lists opt-out` in the side menu, and then when I click on another item in this side menu (for service user `Data quotas`, `Authentication` etc.), the activated regex is somehow broken for this first click - the background of the selected item doesn't reflect the changes, but the whole pages (including URL) are correctly changed.

ngOnDestroy of the mailing lists component used for the url params was the problem

HejdaJakub (approved):

PR #1153, opened 2022-10-05
feat(admin): consent status in assigned users facility table

xflord:
* display consent users' status for the facility on the assigned users page
* also added filter for consent status
* updated openapi


libs/perun/components/src/lib/users-dynamic-list/users-dynamic-list.component.ts
```diff
@@ -100,11 +109,15 @@ export class UsersDynamicListComponent implements OnInit, OnChanges, AfterViewIn
     if (!this.authResolver.isPerunAdminOrObserver()) {
       this.displayedColumns = this.displayedColumns.filter((column) => column !== 'id');
     }
+    if (this.includeConsents) {
+      this.displayedColumns.push('consentStatus');
```
HejdaJakub, line 113: You should add the column `consentStatus` also to the function `getExportDataForColumn` for the correct export of data. I am not sure if it is the best solution, but you can make the function `getExportDataForColumn` non-static and an arrow function:
```
getExportDataForColumn = (data: RichUser, column: string): string => {
```
If you change it in this way, you can then use this reference in the function and you can add this case to switch (you need to add your pipe to the constructor and providers):
```
case 'consentStatus':
        return this.consentStatusPipe.transform(data, this.consents);
```
After that, you have to change `UsersDynamicListComponent.getExportDataForColumn` reference to `this.getExportDataForColumn` in this file.

As I wrote I am not sure if it is a 100 % correct solution, but IMHO it works. If you find a better solution, I will be happy.

HejdaJakub (changes requested):

libs/perun/openapi/src/lib/model/models.ts
```diff
@@ -50,6 +50,7 @@ export * from './extSourceObject';
 export * from './facility';
 export * from './facilityPropagationState';
 export * from './facilityState';
+export * from './facilityWithAttributes';
```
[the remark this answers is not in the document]
xflord: whoops

libs/perun/components/src/lib/users-dynamic-list/users-dynamic-list.component.ts
xflord: Other solution that comes to mind is to create a new object that will hold a User with his corresponding consent status and then get the status for each user in the export methods before calling getDataForExport. Would be a bit more code, but getDataForColumn would remain static (which is imperative i guess?)

HejdaJakub: Yes, the new object should resolve this use case. You just have to be careful during the refactoring since two components currently use this list. It shouldn't be a problem, just saying to don't forget to test both of them.

HejdaJakub (changes requested):

bodnara (changes requested):

apps/admin-gui/src/app/facilities/pages/facility-detail-page/facility-allowed-users/facility-allowed-users.component.ts
```diff
@@ -48,6 +51,12 @@ export class FacilityAllowedUsersComponent implements OnInit {
   filteredServices: Service[] = [this.emptyService];
   selectedService: Service = this.emptyService;
 
+  globalForceConsents: boolean;
+  facilityForceConsents: boolean;
+  consentStatusesList = ['UNSIGNED', 'GRANTED', 'REVOKED'];
+  selectedConsentStatuses: ConsentStatus[] = [];
+  statuses = new UntypedFormControl();
```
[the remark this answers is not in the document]
xflord: Done, also added translation to displaySelectedStatuses()

apps/admin-gui/src/app/facilities/pages/facility-detail-page/facility-allowed-users/facility-allowed-users.component.ts
```diff
@@ -67,14 +77,21 @@ export class FacilityAllowedUsersComponent implements OnInit {
     private resourceService: ResourcesManagerService,
     private authResolver: GuiAuthResolver,
     private storeService: StoreService,
-    private entityStorageService: EntityStorageService
+    private entityStorageService: EntityStorageService,
+    private consentService: ConsentsManagerService,
+    private translate: TranslateService
```
bodnara, line 82: Please use the PerunTranslateService instead. After that remove type casting.

libs/perun/pipes/src/lib/consent-status-icon.pipe.ts
```diff
@@ -0,0 +1,12 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { Consent, RichUser } from '@perun-web-apps/perun/openapi';
+
+@Pipe({ name: 'consentStatusIcon' })
+export class ConsentStatusIconPipe implements PipeTransform {
+  transform(user: RichUser, consents: Consent[]): string {
+    if (!consents) return;
+    const usersConsents = consents.filter((consent) => consent.userId === user.id);
+    // return latest consent (one with the highest id)
+    return usersConsents.reduce((a, b) => (a.id > b.id ? a : b)).status;
```
bodnara, line 10: This can throw an error if the array is empty.
image

bodnara (changes requested):

HejdaJakub (approved):

PR #1154, opened 2022-10-05
Service access fix

xkostka2:


HejdaJakub (changes requested):
If you don't have initially `/service-access` in the URL, `auto_service_access_redirect` is set to `true` and some queryParams are in the URL (f.e. `?idpFilter=social`), then the queryParams are lost during the process when the URL is replaced by `/service-access`.

HejdaJakub (changes requested):
You should run `prettier --write` on `perun-footer.component.ts` to fix the prettier check. You should also rebase this PR because there is a merge conflict due to the newly non-existing `get(...) `method in `storeService` (it is replaced by `getProperty(...)`). 
There are still missing `queryParams`.

HejdaJakub (approved):

PR #1155, opened 2022-10-06
feat(admin): connect creating of the service account and sponsoring

HejdaJakub:
* New workflow - there is a new final step in creating the service account process. Now we allow users to sponsor a newly created service account directly in this workflow.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -1619,8 +1619,5 @@
       },
       "tags": []
     }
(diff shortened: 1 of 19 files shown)


Johaney-s (approved):
Good job, really like the solution of disabling the NEXT button until adding another owner is finished :+1:

PR #1157, opened 2022-10-07
feat: created pipe for checkbox label

bodnara:
* created pipes for master and common checkbox aria label
* added localization to apps
* replaced all usages with new implementation

### DIFF
--- a/libs/perun/components/src/lib/groups-list/groups-list.component.html
+++ b/libs/perun/components/src/lib/groups-list/groups-list.component.html
@@ -20,7 +20,7 @@
           <mat-checkbox
             *ngIf="!disableHeadCheckbox"
             (change)="$event ? masterToggle() : null"
-            [aria-label]="checkboxLabel()"
+            [aria-label]="selected.all | masterCheckboxLabel | translate"
             [checked]="selection.hasValue() && selected.all"
             [indeterminate]="selection.hasValue() && !selected.all"
             color="primary">
@@ -35,6 +35,7 @@
             <mat-checkbox
               (change)="$event ? itemSelectionToggle(group) : null"
               (click)="$event.stopPropagation()"
+              [aria-label]="selection.isSelected(group) | checkboxLabel | translate: {name: group.name}"
               [checked]="selection.isSelected(group)"
               [disabled]="groupStatus.disable"
               attr.data-cy="{{group.name}}-checkbox"
--- a/libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.html
+++ b/libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.html
@@ -18,13 +18,15 @@
       matSortActive="fullName"
       matSortDirection="asc"
       matSortDisableClear>
-      <ng-container matColumnDef="checkbox">
+      <ng-container
+        matColumnDef="checkbox"
+        *ngIf="{all: dataSource | isAllSelected: selection.selected.length :canBeSelected} as selected">
         <th *matHeaderCellDef mat-header-cell class="align-checkbox">
           <mat-checkbox
             (change)="$event ? masterToggle() : null"
-            [aria-label]="checkboxLabel()"
-            [checked]="selection.hasValue() && isAllSelected()"
-            [indeterminate]="selection.hasValue() && !isAllSelected()"
+            [aria-label]="selected.all | masterCheckboxLabel | translate"
+            [checked]="selection.hasValue() && selected.all"
+            [indeterminate]="selection.hasValue() && !selected.all"
             color="primary">
           </mat-checkbox>
         </th>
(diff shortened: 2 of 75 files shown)


bodnara:
After rebase this fixes a bug whith checkboxes in group-list.

libs/perun/components/src/lib/vos-list/vos-list.component.html
```diff
@@ -15,12 +15,12 @@
...
         <th *matHeaderCellDef mat-header-cell class="align-checkbox"></th>
-        <td *matCellDef="let member" class="static-column-size align-checkbox" mat-cell>
+        <td *matCellDef="let vo" class="static-column-size align-checkbox" mat-cell>
           <mat-checkbox
-            (change)="$event ? selection.toggle(member) : null"
+            (change)="$event ? selection.toggle(vo) : null"
             (click)="$event.stopPropagation()"
-            [aria-label]="checkboxLabel(member)"
-            [checked]="selection.isSelected(member)"
+            [aria-label]="selection.isSelected(vo) | checkboxLabel | translate: {name: vo.name ?? vo.vo.name}"
```
HejdaJakub, line 22: It's also a little strange that the name parameter is cut after a few letters. For example instead of `Select item Vo with a very very long name` there is a label `Select item Vo with a very ver` (it occurs not only on vos list - attributes have usually long names). IMHO it's not a big problem, but it is strange :D.

libs/perun/components/src/lib/users-dynamic-list/users-dynamic-list.component.ts
```diff
@@ -142,11 +142,8 @@ export class UsersDynamicListComponent implements OnInit, OnChanges, AfterViewIn
     return numSelected === numRows;
   }
 
-  checkboxLabel(row?: RichUser): string {
-    if (!row) {
-      return `${this.isAllSelected() ? 'select' : 'deselect'} all`;
-    }
-    return `${this.selection.isSelected(row) ? 'deselect' : 'select'} row ${row.id + 1}`;
+  checkboxLabel(): string {
+    return `${this.isAllSelected() ? 'select' : 'deselect'} all`;
```
HejdaJakub, line 146: Why do you use this method instead of `masterCheckboxLabel` pipe?

libs/perun/components/src/lib/groups-list/groups-list.component.html
```diff
@@ -20,7 +20,7 @@
           <mat-checkbox
             *ngIf="!disableHeadCheckbox"
             (change)="$event ? masterToggle() : null"
-            [aria-label]="checkboxLabel()"
```
HejdaJakub, line 23: Remove `checkboxLabel()` method also from .ts file please.

libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.ts
```diff
@@ -170,11 +170,8 @@ export class MembersDynamicListComponent implements AfterViewInit, OnInit, OnCha
     return numSelected === numRows;
   }
 
-  checkboxLabel(row?: RichMember): string {
-    if (!row) {
-      return `${this.isAllSelected() ? 'select' : 'deselect'} all`;
-    }
-    return `${this.selection.isSelected(row) ? 'deselect' : 'select'} row ${row.id + 1}`;
+  checkboxLabel(): string {
+    return `${this.isAllSelected() ? 'select' : 'deselect'} all`;
```
HejdaJakub, line 174: The same. Can we use `masterCheckboxLabel` instead of this function?

HejdaJakub (changes requested):
On the dynamic list (f.e. groups-list) the initial value of the master label doesn't work correctly. After the change, it starts to work (`selected.all` seems to be true when the `dataSource` is empty). 
image

libs/perun/components/src/lib/vos-list/vos-list.component.html
bodnara: I am not sure why is the text cut. As far as I know the aria-label and translation pipe is not restricted by lenght.

libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.html
```diff
@@ -18,21 +18,23 @@
...
+            [indeterminate]="selection.hasValue() && !selected.all"
             color="primary">
           </mat-checkbox>
         </th>
         <td *matCellDef="let member" class="static-column-size align-checkbox" mat-cell>
           <mat-checkbox
             (change)="$event ? selection.toggle(member) : null"
             (click)="$event.stopPropagation()"
-            [aria-label]="checkboxLabel(member)"
+            [aria-label]="selection.isSelected(member) | checkboxLabel | translate: {name: member | userFullName}"
```
HejdaJakub, line 37:
```suggestion
            [aria-label]="selection.isSelected(member) | checkboxLabel | translate: {name: member.user | userFullName}"
```

HejdaJakub (changes requested):
Last small suggestion. It looks good, otherwise.

HejdaJakub (approved):

PR #1160, opened 2022-10-11
fix(admin): disable remove of indirect member in groups view

mattjoke:
* indirect member in group table in member profile section, no longer allows removal of indirect members


bodnara (changes requested):
The indirect tooltip is present even on places where it shouldn't.
image

mattjoke:
It would be a good idea, to test the solution after the merge https://github.com/CESNET/perun/pull/3750

bodnara (approved):

xflord (approved):

PR #1161, opened 2022-10-12
fix: resize window on sidenav reopening

Johaney-s:
* when sidenav is hidden and reopened, the below window's width is not recalculated
* currently this is only a case in admin-gui, the fix is applied also to publications and profile as a precaution


bodnara (approved):
Great solution!

HejdaJakub (approved):

PR #1163, opened 2022-10-13
feat(gui): Minimum password length requirements messages added

Kuliak:
* default configs updated to contain newly added requirement for minimum password length (8 characters)
* CS version of messages were updated as well, if present


HejdaJakub (commented):
I would maybe shorten it and use only the new message `Password must be at least 8 characters long` or at least I would move it to the start of this message, what do you think? The default check on the backend can only fail on a password length (or empty string), am I right? The previous message was only a common recommendation to avoid using accented characters I think (not a condition).

HejdaJakub (approved):

xflord (approved):

PR #1164, opened 2022-10-14
feat(admin): Manage roles on role-page

bodnara:
* it is now possible to (un)set roles for self, user or group on roles page

### DIFF
--- a/apps/admin-gui/src/app/shared/components/roles-list/roles-page.component.html
+++ b/apps/admin-gui/src/app/shared/components/roles-list/roles-page.component.html
@@ -1,4 +1,12 @@
 <h1 class="page-subtitle d-flex">{{'ROLES.TITLE'| translate}}</h1>
+<button
+  *ngIf="rules.length !== 0"
+  mat-flat-button
+  color="accent"
+  class="mb-3 mr-2"
+  (click)="addRole()">
+  {{'ROLES.ADD'| translate}}
+</button>
 <mat-spinner *ngIf="outerLoading" class="ml-auto mr-auto"></mat-spinner>
 <mat-accordion *ngIf="!outerLoading && roles.size > 0">
   <mat-expansion-panel
@@ -87,7 +95,10 @@
   </mat-expansion-panel>
 
   <div *ngFor="let role of ['GROUPADMIN', 'GROUPOBSERVER', 'GROUPMEMBERSHIPMANAGER']">
-    <mat-expansion-panel (opened)="getGroups(roles.get(role).get('Group'))" *ngIf="roles.has(role)">
+    <mat-expansion-panel
+      (opened)="getGroups(roles.get(role).get('Group')); selectedRole = role"
+      (closed)="selection.clear()"
+      *ngIf="roles.has(role)">
       <mat-expansion-panel-header>
         <mat-panel-title>
           <mat-icon class="mr-2" mat-card-avatar svgIcon="perun-group-black"></mat-icon>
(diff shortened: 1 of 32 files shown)


bodnara:
Opening the add role dialog might throw some Priviledge exeption for now. You can ignore it for now, later on all methods of type "getAllEntities" should be callable by SELF.

The issue with role being assigned inderictly through group and therefore cannot be removed for user is still present.

libs/perun/pipes/src/lib/manageable-entities.pipe.ts
```diff
@@ -0,0 +1,22 @@
...
+export class ManageableEntitiesPipe implements PipeTransform {
+  constructor(private authResolver: GuiAuthResolver) {}
+
+  transform(entities: PerunBean[], role: string): PerunBean[] {
+    return entities.filter((e) => this.canSetRoleForEntity(e, role));
+  }
+
+  private canSetRoleForEntity(entity: PerunBean, role: string): boolean {
+    const manageRights = new Map<string, AuthPrivilege>();
+    this.authResolver.setRolesAuthorization([role], entity, manageRights);
```
HejdaJakub, line 18: There are several problems that relate to the nested `function fetchAllRelatedObjects(...)` and the "Facility" object. Firstly at a `roles-page.component.ts` you put as an argument Rich variants of objects, but the Rich variant for the `Facility` is `EnrichedFacility` which doesn't extend PerunBean and has a facility as a nested property. Secondly, if you correctly parse the Facility object, the parameter of the `fetchAllRelatedObjects(...)` method is `Entity[]` type (Vo & Group & Resource & Member & User), so we probably want to add also Facility to this `Entity` type.

You can test this problem when you want to check some facility checkbox under the facility admin expansion panel.

apps/admin-gui/src/app/shared/components/roles-list/roles-page.component.html
```diff
@@ -99,16 +110,32 @@ <h1 class="page-subtitle d-flex">{{'ROLES.TITLE'| translate}}</h1>
...
         </div>
-
+        <span
+          matTooltip="{{'ROLES.REMOVE_DISABLED_TOOLTIP'| translate}}"
+          matTooltipPosition="left"
+          [matTooltipDisabled]="selection.selected.length === 0 || !disableRemove">
+          <button
+            (click)="openConfirmDialog(role)"
+            [disabled]="selection.selected.length === 0 || disableRemove"
+            mat-flat-button
```
HejdaJakub, line 120: NITPICK: Maybe something like this looks better?
image
```suggestion
            class="mt-2"
            mat-flat-button
```

instead of this:
image

HejdaJakub (changes requested):
If I try to set/unset some roles to a group, changes are displayed only after manual reloading. (for users it works better).

It's also a little strange that a group can be its own manager (but it is a general problem if it is problem, not related to this PR). 
image

libs/perun/pipes/src/lib/manageable-entities.pipe.ts
bodnara: Enriched facility should now be parsed at all places.

bodnara:
The refreshing of roles after adding / removing should now be fixed. 
When adding a role entities that are already assigned with given role are now filtered out.

> It's also a little strange that a group can be its own manager (but it is a general problem if it is problem, not related to this PR)

I think this is something we want, so you can give some role in the group for each member.

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -3061,6 +3074,11 @@
           "REMOVE_KEY": "Remove selected key",
           "REMOVE_DIALOG_TITLE": "Remove ssh key",
           "REMOVE_DIALOG_DESCRIPTION": "Selected ssh key will be removed."
+        },
+        "ROLE_SEARCH_SELECT": {
+          "SELECT_ROLE": "Select role",
+          "FIND_ROLE": "Find role...",
+          "NO_ROLE_FOUND": "No matching role find"
```
HejdaJakub, line 3081:
```suggestion
          "NO_ROLE_FOUND": "No matching role found"
```

HejdaJakub (changes requested):
- Some tables (facilities-list and vos-list) are without master toggle (I know that in the original lists the master toggles probably are NOT implemented yet, but here it maybe makes sense - at least for consistency with other tables - so maybe implement it or remove from other tables)
- Master toggles at (for example) groups-list under group roles or resources-list under resource roles don't work correctly
image
- Some translation doesn't work correctly.
image
- If I am not a Perun Admin, I don't have permission to manage some of my roles. Is it a feater or a bug?
image

HejdaJakub:
And what do you think about this point?
> 3. Perun admin - I understand the use case when I want to un/set Perun admin to someone else, but I would maybe hide this button or maybe add "DELETE" string condition when I want to unset this role at My Profile

bodnara:
> * Some tables (facilities-list and vos-list) are without master toggle (I know that in the original lists the master toggles probably are NOT implemented yet, but here it maybe makes sense - at least for consistency with other tables - so maybe implement it or remove from other tables)
> * Master toggles at (for example) groups-list under group roles or resources-list under resource roles don't work correctly

All tables on this page only allow single selection, since we do not support unsetting role on multiple entities. So for consistency all master toggle options are hidden.


> * Some translation doesn't work correctly.

Added missing translations.

> * If I am not a Perun Admin, I don't have permission to manage some of my roles. Is it a feater or a bug?

This was a bug, since RichResource as well as EnrichedFacility have property facility. RichResources were then mapped to nested facility. By changing the order of the checks this should no longer happen.

HejdaJakub (commented):
It's an edge use case, but there is a real chance that the user will have some role of `ignoredRoles` like f.e. `CABINETADMIN`. Then the label looks a little ugly and there is no option how to remove this role. In some ways, it's really an edge use case, but on the other hand, we maybe should discuss which roles we want to display, if we want to fix at least labels for `ignoredRoles`, or if we want to handle some default option. The same situation maybe can occur for `AUDITCONSUMERADMIN` role?
image

HejdaJakub:
NOTE: Roles like `CABINETADMIN`, `AUDITCONSUMERADMIN`, or instance-specific roles like `CYCLOPADMIN` will be fixed in separate task (PR).

HejdaJakub (approved):

PR #1165, opened 2022-10-17
feat: better workflow for MFA dialog with MfaRolePrivilegeException

HejdaJakub:
* If there is the MfaRolePrivilegeException, we want to hide the cancel button.
* Previously after clicking on the cancel button there appeared another dialog with a refresh button, which only redirects the user back to the first dialog, so the cancel button is useless for this use case with role exception.


bodnara (approved):

HejdaJakub:
The description in the dialog for this use case changed.

bodnara (approved):

PR #1166, opened 2022-10-17
feat: reaction on invalid access token

HejdaJakub:
* When invalid access token occurs (for example revoked token), then new dialog with better information is opened. In this dialog is also "Sign in" button which start authentication flow.

### DIFF
--- a/libs/perun/services/src/lib/ApiInterceptor.ts
+++ b/libs/perun/services/src/lib/ApiInterceptor.ts
@@ -19,6 +19,7 @@ import { MatDialog, MatDialogRef } from '@angular/material/dialog';
 import { SessionExpirationDialogComponent } from '@perun-web-apps/perun/session-expiration';
 import { InitAuthService } from './init-auth.service';
 import { MfaHandlerService } from './mfa-handler.service';
+import { OAuthService } from 'angular-oauth2-oidc';
 
 @Injectable()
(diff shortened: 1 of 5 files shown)


bodnara (commented):
Coulnd't we reuse the SessionExpirationDialog ? It seems to serve the same purpose and I would say is a better message to users who might not know what access token refering to.

HejdaJakub:
> Coulnd't we reuse the SessionExpirationDialog ? It seems to serve the same purpose and I would say is a better message to users who might not know what access token is refering to.

You're right, this message sounds better. I replaced the dialog by SessionExpirationDialog.

libs/perun/services/src/lib/ApiInterceptor.ts
```diff
@@ -61,6 +63,7 @@ export class ApiInterceptor implements HttpInterceptor {
         sessionStorage.setItem('auth:redirect', location.pathname);
         sessionStorage.setItem('auth:queryParams', location.search.substring(1));
         void this.initAuthService.handleAuthStart();
+        this.reauthenticate();
```
bodnara, line 66: There was a recent change to this and I believe you can remove the hanldeAuthStart but place it inside of the reauthenticate instead of startAuth.

bodnara (approved):

PR #1167, opened 2022-10-19
fix(admin): e2e repair

sarkapalkovicova:
Fixing e2e tests that failed.

### DIFF
--- a/apps/admin-gui-e2e/src/e2e/resource-observer-perun_admin.cy.js
+++ b/apps/admin-gui-e2e/src/e2e/resource-observer-perun_admin.cy.js
@@ -17,13 +17,19 @@ context('Actions', () => {
   });
 
   beforeEach(() => {
-    cy.visit('organizations')
+    cy.visit('home')
+      .get(`[data-cy=access-item-button]`)
+      .click()
+      .get('[data-cy=auto-focused-filter]')
+      .type(dbVoName)
       .get(`[data-cy=${dbVoName}]`)
       .click()
-      .get('[data-cy=resources]')
+      .get(`[data-cy=resources]`)
       .click()
-      .get('[data-cy=resource-list]')
+      .get(`[data-cy=resource-list]`)
       .click()
+      .get('[data-cy=unfocused-filter]')
+      .type(dbResourceName)
       .get(`[data-cy=${dbResourceName}]`)
       .click();
   });
@@ -35,6 +41,8 @@ context('Actions', () => {
       .click()
 
       // group should be visible
+      .get('[data-cy=filter-input]')
+      .type(dbGroupName)
       .get(`[data-cy=${dbGroupName}]`)
       .should('exist')
   })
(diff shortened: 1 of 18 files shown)


xflord (commented):
The changes seem fine, everything that needs to be filtered gets filtered. However, when running the tests I get this error:
Screenshot from 2022-10-24 13-14-40
Probably an issue with some new changes to backend and not with the changes you made, but would be nice to have it fixed.

xflord:
Sometimes, a checkbox is ticked BEFORE the new filter value causes the list to refresh, which unticks the checkbox and ultimately causes the test to fail:
Actions -- test remove group member (failed)


https://user-images.githubusercontent.com/97669007/198873938-863a72cc-68b1-4817-ba7b-5459dc7aaa44.mp4

xflord:
Also 'test list assigned users' seems to be failing, because due to to recent changes to new gui an existing consent hub is expected for every facility (which isn't created in the e2e container database).
Screenshot from 2022-11-01 10-48-54

xflord (approved):
Everything seems to pass now, good job

apps/admin-gui-e2e/src/e2e/resource-observer-perun_admin.cy.js
```diff
@@ -17,13 +17,7 @@ context('Actions', () => {
...
 
   beforeEach(() => {
-    cy.visit('organizations')
-      .get(`[data-cy=${dbVoName}]`)
-      .click()
-      .get('[data-cy=resources]')
-      .click()
-      .get('[data-cy=resource-list]')
-      .click()
+    cy.visit('home')
```
HejdaJakub, line 20: I have just one suggestion. In several tests, we have `cy.visit('home').get('something')` in the `beforeEach()` section, but I would maybe prefer to use more concrete navigation, similar to what you deleted here. If we extend these tests in the future and the user will have more objects, we won't be able to be sure that the object that we want is shown between recently viewed objects. 
Another possible solution is to navigate to the object in this way at least in the `before()` section (then we will know that this object will be shown between recently viewed objects) and then use the shortened version in `beforeEach()` section.
What do you think about it @sarkapalkovicova @xflord ?

xflord: Starting from 'home' and clicking through the side menu would probably safest and cover more of the app, no?
Also a side-note, the `cy.visit('organizations')` line is causing problems (at least that's what i deduced) in the e2e tests (see here: https://gitlab.ics.muni.cz/perun/e2e-perun/-/jobs/650843/artifacts/file/perun-rpc/cypress/videos/resource-self-service-perun_admin.cy.js.mp4). 
The test visits 'service-access' in the 'before' phase, starts 'beforeEach' where it's rerouted to 'organizations' but then a redirect to 'home' happens, which (I believe) is from the authorization flow.
 I remembered it now that you mentioned this and what Sarka did here would theoretically fix it.

HejdaJakub: Good point, I agree that probably the best solution is to start from `home` and continue through the side menu. So we should avoid direct redirecting (like for example `cy.visit('organizations')`) in `beforeEach()` section and we should also avoid continuing from home through recently viewed objects (cards).

HejdaJakub (approved):

PR #1168, opened 2022-10-24
fix(admin-gui): Friendly name is visible

mattjoke:
* friendly name is visible in attribute edit
* added copy button for friendly name


sarkapalkovicova (commented):
In my opinion, it looks good. For clarity, I just want to suggest renaming `copyFriendlyName` to `copyUrn` and the same for variables in assets.

sarkapalkovicova (approved):

bodnara (approved):

PR #1169, opened 2022-10-25
feat(publications): improve Publications application for a production usage

HejdaJakub:
* Checked rules for different roles (PERUNADMIN, CABINETADMIN, SELF) and fixed some use cases for these roles (un/lock publication, side-menu, routing).
* Publications filter fixed - now it is possible to filter by code (isbn, issn, doi), in a date-picker are used only years, not months/days, start year is not selected by default, and also after Clear filter action is the start year also again unselected.
* Create publication guide improved - fixed some labels, button positioning, fixed the finding of similar publications, a label of a similar publication in a tab menu, and navigation between steps.
* According to an old GUI there is a new alert when a user tries to create a publication where he/she is not selected as an author.
* Fixed some other labels, translations, icons, and some other CSS magic has been involved.

### DIFF
--- a/apps/publications/src/app/components/add-authors/add-authors.component.ts
+++ b/apps/publications/src/app/components/add-authors/add-authors.component.ts
@@ -1,11 +1,11 @@
-import { Component, Input, OnInit } from '@angular/core';
+import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
 import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
 import { AddAuthorsDialogComponent } from '../../dialogs/add-authors-dialog/add-authors-dialog.component';
 import { UniversalConfirmationItemsDialogComponent } from '@perun-web-apps/perun/dialogs';
 import { Author, CabinetManagerService, PublicationForGUI } from '@perun-web-apps/perun/openapi';
 import { SelectionModel } from '@angular/cdk/collections';
 import { MatDialog } from '@angular/material/dialog';
-import { NotificatorService } from '@perun-web-apps/perun/services';
+import { NotificatorService, StoreService } from '@perun-web-apps/perun/services';
 import { TranslateService } from '@ngx-translate/core';
 import { TABLE_PUBLICATION_AUTHORS } from '@perun-web-apps/config/table-config';
 
@@ -17,6 +17,9 @@ import { TABLE_PUBLICATION_AUTHORS } from '@perun-web-apps/config/table-config';
 export class AddAuthorsComponent implements OnInit {
   @Input() publication: PublicationForGUI;
   @Input() selection: SelectionModel<Author> = new SelectionModel<Author>(true, []);
+  @Input() disableRouting = false;
+  @Input() similarityCheck = false;
+  @Output() yourselfAsAnAuthor: EventEmitter<boolean> = new EventEmitter<boolean>();
 
   filterValue = '';
   loading = false;
@@ -27,7 +30,8 @@ export class AddAuthorsComponent implements OnInit {
     private dialog: MatDialog,
     private cabinetService: CabinetManagerService,
     private notificator: NotificatorService,
-    private translate: TranslateService
+    private translate: TranslateService,
+    private storeService: StoreService
   ) {}
 
   ngOnInit(): void {
@@ -37,6 +41,9 @@ export class AddAuthorsComponent implements OnInit {
   refresh(): void {
     this.loading = true;
     this.cabinetService.findAuthorsByPublicationId(this.publication.id).subscribe((authors) => {
+      this.yourselfAsAnAuthor.emit(
+        authors.some((author) => author.id === this.storeService.getPerunPrincipal().userId)
+      );
       this.publication.authors = authors;
       this.selection.clear();
       this.loading = false;
--- a/apps/publications/src/app/components/publication-detail-list/publication-detail-list.component.ts
+++ b/apps/publications/src/app/components/publication-detail-list/publication-detail-list.component.ts
@@ -6,9 +6,8 @@ import {
   Publication,
   PublicationForGUI,
 } from '@perun-web-apps/perun/openapi';
-import { SelectionModel } from '@angular/cdk/collections';
 import { Moment } from 'moment';
-import { UntypedFormControl, Validators } from '@angular/forms';
+import { FormControl, Validators } from '@angular/forms';
 import { DateAdapter, MAT_DATE_FORMATS, MAT_DATE_LOCALE } from '@angular/material/core';
 import { MomentDateAdapter } from '@angular/material-moment-adapter';
 import * as _moment from 'moment';
@@ -46,18 +45,18 @@ export const YEAR_MODE_FORMATS = {
 export class PublicationDetailListComponent implements OnInit {
   @Input() publication: PublicationForGUI;
   @Input() categories: Category[] = [];
+  @Input() similarityCheck = false;
   @Output() edited: EventEmitter<boolean> = new EventEmitter<boolean>();
   loading = false;
   dataSource: MatTableDataSource<{ key; value }> = null;
   displayedColumns = ['key', 'value'];
-  isChanging = new SelectionModel<{ key: string; value: string }>(true, []);
   keys: string[];
   values: string[];
   map: { key: string; value: string }[] = [];
-  yearControl: UntypedFormControl;
-  categoryControl: UntypedFormControl;
-  rankControl: UntypedFormControl;
-  titleControl: UntypedFormControl;
+  yearControl: FormControl<Moment>;
+  categoryControl: FormControl<string>;
+  rankControl: FormControl<number>;
+  titleControl: FormControl<string>;
   editing = false;
   maxYear: Moment;
 
@@ -97,10 +96,10 @@ export class PublicationDetailListComponent implements OnInit {
     }
     this.dataSource = new MatTableDataSource<{ key; value }>(this.map);
 
-    this.titleControl = new UntypedFormControl(this.publication.title, Validators.required);
-    this.yearControl = new UntypedFormControl(moment().year(this.publication.year));
-    this.categoryControl = new UntypedFormControl(this.publication.categoryName);
-    this.rankControl = new UntypedFormControl(this.publication.rank, [
+    this.titleControl = new FormControl(this.publication.title, Validators.required);
+    this.yearControl = new FormControl(moment().year(this.publication.year));
+    this.categoryControl = new FormControl(this.publication.categoryName);
+    this.rankControl = new FormControl(this.publication.rank, [
       Validators.pattern(/^[0-9]+(\.[0-9])?$/),
       Validators.required,
     ]);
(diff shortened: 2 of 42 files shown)


apps/publications/src/app/components/add-authors/add-authors.component.ts
```diff
@@ -37,6 +40,10 @@ export class AddAuthorsComponent implements OnInit {
   refresh(): void {
     this.loading = true;
     this.cabinetService.findAuthorsByPublicationId(this.publication.id).subscribe((authors) => {
+      this.yourselfAsAnAuthor.emit(
+        authors.filter((author) => author.id === this.storeService.getPerunPrincipal().userId)
+          .length === 1
+      );
```
bodnara, line 46:
```suggestion
      this.yourselfAsAnAuthor.emit(
        authors.some((author) => author.id === this.storeService.getPerunPrincipal().userId)
      );
```

apps/publications/src/app/components/publications-list/publications-list.component.ts
```diff
@@ -25,7 +25,7 @@ import {
   TABLE_ITEMS_COUNT_OPTIONS,
   TableWrapperComponent,
 } from '@perun-web-apps/perun/utils';
```
bodnara, line 27: `'parseFullName' is defined but never used`

apps/publications/src/app/components/year-range/year-range.component.html
```diff
@@ -5,14 +5,13 @@
     [matDatepicker]="sdp"
     [max]="startMaxYear"
     [formControl]="startYear"
-    class="disable"
-    readonly />
```
bodnara, line 9: I think the readonly property should be kept, the end year date picker still has it.

apps/publications/src/app/pages/create-publication-page/create-publication-page.component.ts
```diff
@@ -7,6 +7,8 @@ import { Router } from '@angular/router';
   styleUrls: ['./create-publication-page.component.scss'],
 })
 export class CreatePublicationPageComponent {
+  displayTooltip = false;
```
bodnara, line 10: I believe this can be removed ?

apps/publications/src/app/pages/create-publication-page/create-single-publication-page/create-single-publication-page.component.html
```diff
@@ -176,26 +181,20 @@ <h5 class="ml-auto mr-auto font-weight-bold">
...
-          mat-stroked-button
-          type="button"
-          matStepperPrevious>
+          class="ml-2"
+          mat-flat-button
+          (click)="redirect(['create-publication'])"
+          type="button">
+          {{'CREATE_SINGLE_PUBLICATION.CANCEL' | translate}}
+        </button>
+        <button class="ml-auto" mat-flat-button type="button" matStepperPrevious>
```
bodnara, line 192: The back button does not work on step 2 if no similiar publications are found.

apps/publications/src/app/pages/publication-detail/publication-detail.component.ts
```diff
@@ -51,6 +53,10 @@ export class PublicationDetailComponent implements OnInit {
 
   ngOnInit(): void {
     this.initLoading = true;
+    this.disableRouting = this.disableRouting
+      ? this.disableRouting
+      : !this.authResolver.isCabinetAdmin();
```
bodnara, line 58: Wouldn't this work as well ?
```suggestion
    this.disableRouting = this.disableRouting || !this.authResolver.isCabinetAdmin();
```

apps/publications/src/assets/i18n/en.json
```diff
@@ -100,15 +98,17 @@
...
     "NOT_BELOW": "If NO, continue with Create button below.",
-    "IS_BELOW": "If YES, click on it to see details and add yourself as an author. Afterwards continue with the Return button. If desired publication is locked for changes, notify administrators about your request.",
+    "IS_BELOW": "If YES, click on it to see details and add yourself as an author. Afterwards continue with the Cancel button. If desired publication is locked for changes, notify administrators about your request.",
     "CHECK_PERFORM": "Checking for similar publications",
     "NO_SIMILAR": "Publication similarity check passed",
     "SUCCESS": "Publication was successfully created.",
-    "PUB_CREATED": "Publication was created, however it is not fully set up. You can finish setting up your publication by completing all steps. Additionally you can always edit publications on their detail."
+    "PUB_CREATED": "Publication was created, however it is not fully set up. You can finish setting up your publication by completing all steps. Additionally you can always edit publications on their detail.",
+    "NOT_AN_AUTHOR_DIALOG_TITLE": "You are not set as an author",
```
bodnara, line 110: Just so it's not the same as start of the alert. Maybe you can think of even better title.
```suggestion
    "NOT_AN_AUTHOR_DIALOG_TITLE": "Not an author",
```

bodnara (changes requested):
I would appreciate if would could change all form controles to the typed form in all the affected components. I believe there are couple already changed in this PR but it would be nice if all of them would.

apps/publications/src/app/components/year-range/year-range.component.html
HejdaJakub: I used to want it like this because the end year shouldn't be empty (there should be at least max value - current year), but for the start year, it makes sense to me, so I wanted to allow the user to delete the start year manually. On the other hand, it looks inconsistently now and the user can clear the filter which deletes the state year, so I agree that `readonly` attribute should probably be here.

HejdaJakub:
> I would appreciate if would could change all form controles to the typed form in all the affected components. I believe there are couple already changed in this PR but it would be nice if all of them would.

You're right. After a quick look, I decided to change all `UntypedFormControl` in the publications folder to typed, so now it should be changed everywhere in this folder.

apps/publications/src/app/components/publication-detail-list/publication-detail-list.component.ts
```diff
@@ -8,7 +8,7 @@ import {
 } from '@perun-web-apps/perun/openapi';
 import { SelectionModel } from '@angular/cdk/collections';
```
bodnara, line 9:
```suggestion

```

apps/publications/src/app/components/publication-filter/publication-filter.component.ts
```diff
@@ -1,10 +1,12 @@
 import { Component, EventEmitter, OnInit, Output } from '@angular/core';
 import { CabinetManagerService, Category } from '@perun-web-apps/perun/openapi';
-import { UntypedFormControl } from '@angular/forms';
+import { FormControl } from '@angular/forms';
 import * as _moment from 'moment';
 import { DateAdapter, MAT_DATE_FORMATS, MAT_DATE_LOCALE } from '@angular/material/core';
 import { MomentDateAdapter } from '@angular/material-moment-adapter';
 import { formatDate } from '@angular/common';
+import { EMPTY } from 'rxjs';
```
bodnara, line 8:
```suggestion

```

apps/publications/src/app/components/year-range/year-range.component.ts
```diff
@@ -1,33 +1,35 @@
...
   startMaxYear: Date;
   endMinYear: Date;
   endMaxYear: Date;
 
   ngOnInit(): void {
     this.endMaxYear = new Date();
-    this.startMaxYear = this.endYear.value as Date;
-    this.endMinYear = this.startYear.value as Date;
+    this.startMaxYear = this.endYear.value.toDate();
+    this.startYear.setValue(null);
```
bodnara, line 22: By setting this to null, it can cause this and parent component to call `.toDate()` on null

apps/publications/src/app/components/publication-filter/publication-filter.component.ts
```diff
@@ -83,8 +85,8 @@ export class PublicationFilterComponent implements OnInit {
     this.code.setValue('');
     this.selectedMode = 'isbn/issn';
     this.selectedCategory = 'no_value';
-    this.startYear = new UntypedFormControl(moment());
-    this.endYear = new UntypedFormControl(moment());
+    this.startYear = new FormControl(moment());
+    this.endYear = new FormControl(moment());
```
bodnara, line 89: Can we just set value some default value for the year controls as well ? Instead of creating new objects.

apps/publications/src/app/dialogs/add-category-dialog/add-category-dialog.component.ts
```diff
@@ -29,15 +29,16 @@ export class AddCategoryDialogComponent implements OnInit {
...
       Validators.required,
       Validators.pattern('^[\\w.-]+( [\\w.-]+)*$'),
       Validators.maxLength(128),
     ]);
-    this.rankCtrl = new UntypedFormControl(null, [
+    this.rankCtrl = new FormControl(0, [
       Validators.required,
       Validators.pattern('^[0-9]+(\\.[0-9])?$'),
     ]);
+    this.rankCtrl.setValue(null);
```
bodnara, line 41:
```suggestion
    this.rankCtrl = new FormControl<number>(null, [
      Validators.required,
      Validators.pattern('^[0-9]+(\\.[0-9])?$'),
    ]);
```

bodnara (changes requested):

apps/publications/src/app/dialogs/add-category-dialog/add-category-dialog.component.ts
HejdaJakub: If I try to create a typed FormControl with a null value, I get the ESLint error: `ESLint: Unsafe assignment of type FormControl<any> to a variable of type FormControl<number>.(@typescript-eslint/no-unsafe-assignment)`. For some reason, ESLint is OK with setting the null value but not when I create this typed object with the null value.

HejdaJakub: Sorry, my bad, I tried it without `<number>` part, so it works well [fixed].

apps/publications/src/assets/i18n/en.json
```diff
@@ -100,15 +98,17 @@
...
     "CREATE": "Create publication",
-    "RETURN": "Return",
+    "CANCEL": "Cancel",
     "SIMILAR_FOUND": "Similar publication already exists",
     "CHECK_BELOW": "Please check if your publication is listed below",
     "NOT_BELOW": "If NO, continue with Create button below.",
-    "IS_BELOW": "If YES, click on it to see details and add yourself as an author. Afterwards continue with the Return button. If desired publication is locked for changes, notify administrators about your request.",
+    "IS_BELOW": "If YES, click on it to see details and add yourself as an author. Afterwards continue with the Cancel button. If desired publication is locked for changes, notify administrators about your request.",
     "CHECK_PERFORM": "Checking for similar publications",
     "NO_SIMILAR": "Publication similarity check passed",
```
Johaney-s, line 107: I'd alter this text to something like `Check passed - no similar publications were found.`

Johaney-s (commented):
Love it, here's what I think could be improved:
- The similarity check states to 'see details and add yourself as an author' in case similar publication is found, but the cursor on the table is suggesting it is not clickable. After clicking it to see detail, this error pops up: 
image
- When listing all publications, after clicking next page in the table, my position on the page changes

bodnara (approved):

HejdaJakub:
> * The similarity check states to 'see details and add yourself as an author' in case similar publication is found, but the cursor on the table is suggesting it is not clickable. After clicking it to see detail, this error pops up:

Rights have been changed on the backend (https://github.com/CESNET/perun/pull/3761)

> * When listing all publications, after clicking next page in the table, my position on the page changes

The position change after clicking on next/prev page is common in all GUI.

Johaney-s (approved):

PR #1170, opened 2022-10-25
feat(pwd-reset): configurable labels

bodnara:
* added password labels to configuration
* changed password help to map
* allowed service access for pwd reset

### DIFF
--- a/apps/password-reset/src/app/pages/password-reset-page/password-reset-page.component.ts
+++ b/apps/password-reset/src/app/pages/password-reset-page/password-reset-page.component.ts
@@ -1,6 +1,8 @@
 import { Component, Input, OnInit } from '@angular/core';
 import { DomSanitizer, SafeHtml } from '@angular/platform-browser';
 import { StoreService } from '@perun-web-apps/perun/services';
+import { PasswordAction } from '@perun-web-apps/perun/models';
+import { TranslateService } from '@ngx-translate/core';
 
 @Component({
   selector: 'perun-web-apps-password-reset-page',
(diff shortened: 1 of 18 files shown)


mattjoke (commented):
There are some misspellings, probably from earlier version, otherwise I think it looks good!

bodnara:
Translations were fixed. Also query parameters are now merged correctly after login.

mattjoke (approved):

bodnara:
Marked this  PR as breaking change since it changed configuration.

apps/password-reset/src/app/pages/password-reset-page/password-reset-page.component.ts
```diff
@@ -1,38 +1,48 @@
 import { Component, Input, OnInit } from '@angular/core';
 import { DomSanitizer, SafeHtml } from '@angular/platform-browser';
 import { StoreService } from '@perun-web-apps/perun/services';
+import { PasswordAction, PasswordLabels } from '@perun-web-apps/perun/models';
```
HejdaJakub, line 4: 'PasswordLabels' is defined but never used
```suggestion
import { PasswordAction } from '@perun-web-apps/perun/models';
```

HejdaJakub (changes requested):
- There is one use case when we want to correctly handle also errors in response from the backend - not only check when I am typing but also catch errors after sending a request for reset/activation - there is an infinitive loader. NOT only for `PasswordStrengthException`, but also for `LoginNotExistsException`.

image
image

- I realized one more thing. There isn't a really good solution for how we parse mode from the URL - we parse any occurrence of a string in the whole URL - `this.mode = queryParams.includes('activation') ? 'activation' : 'reset';` - I know that the namespace like this probably doesn't exist (I hope), but we should parse it in some better way, something like this: `this.mode = parseQueryParams('mode', queryParams) === 'activation' ? 'activation' : 'reset';`
image

- I know that this is not a part of your task, but if I log in by `service-access` with queryParams, everything looks good. But if I refresh the page, the queryParams are lost and I am not even able to type them back manually.

HejdaJakub (changes requested):
- It looks like there is still a problem with query params when I am logged in with service-access.
- I am not sure if we have some specific error on the backend for the use case when the account is already activated (IMHO we have), but I get this message also if the backend error is `PasswordStrengthFailedException` so in this case we probably do NOT inform the user correctly. I am not sure if the solution is just to modify the instance config to present a universal message for errors during the account activation or if we want to somehow deal with these errors separately. 
image

HejdaJakub (commented):
If we decide to fix `/service-access` and the `queryParams` in another PR, I will have just one last suggestoin / question (in the comment below).

PR #1171, opened 2022-10-27
feat(profile): support securityText attribute

HejdaJakub:
* In Settings->Authentication section we already have support for security images. We wanted to add support also for security texts.

### DIFF
--- a/apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
+++ b/apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
@@ -11,6 +11,8 @@ import {
 import { RemoveStringValueDialogComponent } from '@perun-web-apps/perun/dialogs';
 import { MatSlideToggle } from '@angular/material/slide-toggle';
 import { AddAuthTextDialogComponent } from '../../../components/dialogs/add-auth-text-dialog/add-auth-text-dialog.component';
+import { ComponentType } from '@angular/cdk/overlay';
+import { Observable, Subject } from 'rxjs';
 
 @Component({
   selector: 'perun-web-apps-settings-authentication',
@@ -21,8 +23,13 @@ export class SettingsAuthenticationComponent implements OnInit {
   @ViewChild('toggle') toggle: MatSlideToggle;
 
   imgAtt: Attribute;
-  textAtt: Attribute;
+  imgAttrName: string;
   imageSrc = '';
+  textAtt: Attribute;
+  textAttrName: string;
+  componentMapper: {
+    [key: string]: ComponentType<AddAuthImgDialogComponent | AddAuthTextDialogComponent>;
+  };
   mfaUrl = '';
   displayImageBlock: boolean;
   displayTextBlock: boolean;
@@ -38,6 +45,13 @@ export class SettingsAuthenticationComponent implements OnInit {
 
   ngOnInit(): void {
     const mfa = this.store.getProperty('mfa');
+    this.imgAttrName = this.store.getProperty('mfa').security_image_attribute;
+    this.textAttrName = this.store.getProperty('mfa').security_text_attribute;
+
+    this.componentMapper = {
+      AddAuthImgDialogComponent: AddAuthImgDialogComponent,
+      AddAuthTextDialogComponent: AddAuthTextDialogComponent,
+    };
 
     this.translate.onLangChange.subscribe(() => {
       this.mfaUrl = this.translate.currentLang === 'en' ? mfa.url_en : mfa.url_cs;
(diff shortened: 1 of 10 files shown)


apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
```diff
@@ -18,33 +23,33 @@ export class SettingsAuthenticationComponent implements OnInit {
 
   removeDialogTitle: string;
```
xflord, line 24: You can delete this, since you use constants instead further on

apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
```diff
@@ -18,33 +23,33 @@ export class SettingsAuthenticationComponent implements OnInit {
...
   imgAtt: Attribute;
+  textAtt: Attribute;
   imageSrc = '';
   removeDialogDescription: string;
```
xflord, line 28: this too

apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
```diff
@@ -18,33 +23,33 @@ export class SettingsAuthenticationComponent implements OnInit {
...
-    translate
-      .get('AUTHENTICATION.DELETE_IMG_DIALOG_DESC')
-      .subscribe((res: string) => (this.removeDialogDescription = res));
-    translate
-      .get('AUTHENTICATION.SAVE_IMG_SUCCESS')
-      .subscribe((res: string) => (this.saveImgSuccess = res));
-    translate
-      .get('AUTHENTICATION.REMOVE_IMG_SUCCESS')
-      .subscribe((res: string) => (this.removeImgSuccess = res));
+    // translate
```
xflord, line 41: none if this is used anymore I believe

apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
```diff
@@ -63,68 +68,106 @@ export class SettingsAuthenticationComponent implements OnInit {
...
+        this.notificatorService.showSuccess(
+          this.translate.instant(`AUTHENTICATION.SAVE_${mode === 'image' ? 'IMG' : 'TEXT'}_SUCCESS`)
+        );
+        this.loadSecurityAttribute(mode);
       }
     });
   }
 
-  onDeleteImg(): void {
+  onAddText(): void {
```
xflord, line 138: forgot to delete this?

apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
```diff
@@ -63,68 +68,106 @@ export class SettingsAuthenticationComponent implements OnInit {
...
 
-  onAddImg(): void {
+  onAddAttribute(mode: 'image' | 'text'): void {
     const config = getDefaultDialogConfig();
     config.width = '500px';
-    config.data = { theme: 'user-theme', attribute: this.imgAtt };
+    config.data = { theme: 'user-theme', attribute: mode === 'image' ? this.imgAtt : this.textAtt };
 
-    const dialogRef = this.dialog.open(AddAuthImgDialogComponent, config);
+    let dialogRef: MatDialogRef<AddAuthImgDialogComponent | AddAuthTextDialogComponent>;
```
xflord, line 121: a ternary operator might be a bit prettier here, however up to you.

xflord (commented):
Few cosmetic details, functionally seems to work well (once the attribute is created).

apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
HejdaJakub: Did you mean this one `removeDialogDescription`?

HejdaJakub:
Sorry, I totally forgot to clean up this component after the refactoring... ( ❛︡ ₒ ❛︠ )

apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
```diff
@@ -16,115 +20,126 @@ import { MatSlideToggle } from '@angular/material/slide-toggle';
...
+    this.displayTextBlock = this.store.getProperty('mfa').enable_security_text;
+    if (this.displayTextBlock) {
+      this.loadSecurityAttribute('text');
     }
   }
 
-  loadImage(): void {
-    this.loadingImg = true;
-    const imgAttributeName = this.store.getProperty('mfa').security_image_attribute;
+  loadSecurityAttribute(mode: 'image' | 'text'): void {
```
bodnara, line 57: Using this "mode" argument throughout the component is in my opinion not the best solution. It seems like every other line is a `if else` statment. If at some point we would decine to allow another "mode" we need to add a lot of new branches so it is not easily expandable.

If possible pass all required information as an argument as it is known by the "caller".

xflord: I see your point, but IMO some if statements would still need to stay (eg. assigning the attribute to imgAttr or textAttr) and overall the mode argument makes the methods easier to use, no? Maybe a switch, which would set the variables based on mode at the beginning of the method would make it easier expandable, should a new mode be added?

xflord (approved):
fine by me

apps/user-profile/src/app/pages/settings-page/settings-authorization/settings-authentication.component.ts
bodnara: Yes, perhaps some if statements are neccessery which is fine but right now it feels like each assigment is a if else. It only looks somewhat clean because of the `? :` operator. For the example of assigning attributes I just imagined something like `imgAttr = load(attrName);`.
I think the mode argument passes all logic handling to the function itself and that I think is a bad "design". Imagine we have 10 modes instead of 2 or 3 now each method starts with a 10 case switch that just sets the rights values. Instead I would like to pass the correct values as arguments since they are known by the caller. You can still get some suggestions/autocomplete from IDE as to what is a valid input by giving it a nice type not just e.g. string. Now instead of adding new case to multiple methods you just need to expand the input type.

HejdaJakub: I am not sure if it looks better now since the `loadSecurityAttribute(...)` method needs to be asynchronous, but you can consider it on your own. Maybe there would be helpful some rxjs magic, but I wasn't able to find the better solution in rxjs.

bodnara: I think it is better and the amount of rxjs magic is just right. To touch on the number of modes we will probalby support I don't necesserily excpet we will add more modes or that some other ones would even make sense. I just wanted to point out that in my opinion this usage of "mode" is just digging a hole for ourselfs. The functions handle a lot of unnecessary logic which makes them harder to read and expand. It also leads to more usage of the mode as it leaks to other parts of the code and its a whole cycle then.

bodnara (approved):

PR #1174, opened 2022-11-02
fix(admin): expiration change dialog on the vo members list can no lo…

xflord:
…nger be opened if user doesn't have rights to change member expiration

### DIFF
--- a/libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.ts
+++ b/libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.ts
@@ -19,6 +19,7 @@ import {
   DynamicPaginatingService,
   GuiAuthResolver,
   TableCheckbox,
+  EntityStorageService,
 } from '@perun-web-apps/perun/services';
 import {
   Member,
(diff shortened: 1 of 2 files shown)


libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.ts
```diff
@@ -173,6 +173,9 @@ export class MembersDynamicListComponent implements AfterViewInit, OnInit, OnCha
   changeStatus(event: Event, member: RichMember, groupId?: number): void {
     event.stopPropagation();
 
+    // Do not open status dialog, if caller isn't VOADMIN/PERUNADMIN
+    if (!this.authResolver.isThisVoAdmin(this.voId)) return;
```
bodnara, line 177: If possible please do this check with the `isAuthorized` method from `GuiAuthResolver`.

bodnara (changes requested):

bodnara (changes requested):
Right now the dialog will open even if user does not have rights to change status, because the async call is made but you don't wait for the result.
Anyway I believe you can get the vo/group from the EntityStoreService and just check for the rights when the members list is created. Also if user does not have rights to open this dilaog change the cursor on hover to default.

libs/perun/components/src/lib/members-dynamic-list/members-dynamic-list.component.ts
bodnara: We have to make sure the correct policy is used for vo/group based on which page is the table used.

bodnara (approved):

HejdaJakub (approved):

PR #1176, opened 2022-11-03
fix(admin): move default filter option in applications up

Johaney-s:
* 'pending' applications are preselected in the filer
* the order of the selection makes the dropdown sidebar slide downwards, which is confusing for the users as they don't see the top options
* prioritized the pending applications field to keep sidebar at top when filter menu is opened


bodnara (commented):
It makes sense at least to me that the "all" option would be first and then "pending" as second. What do you think ?

bodnara (approved):

PR #1179, opened 2022-11-04
refactor: unify side menu config properties, add missing

Johaney-s:
* unified configuration properties for sidemenu and subsidemenu
* added new hover_text_color property


Johaney-s:
when this solution is approved, i'll update ansible templates and documentation of apps

bodnara (approved):

PR #1182, opened 2022-11-10
feat(admin): group synchronization dialog now displays the interval b…

xflord:
…etween synchronizations

### DIFF
--- a/libs/perun/dialogs/src/lib/group-sync-detail-dialog/group-sync-detail-dialog.component.ts
+++ b/libs/perun/dialogs/src/lib/group-sync-detail-dialog/group-sync-detail-dialog.component.ts
@@ -4,6 +4,7 @@ import { GroupsManagerService, RichGroup } from '@perun-web-apps/perun/openapi';
 import { Urns } from '@perun-web-apps/perun/urns';
 import { getAttribute } from '@perun-web-apps/perun/utils';
 import { NotificatorService } from '@perun-web-apps/perun/services';
+import { formatDate } from '@angular/common';
 
 export type SyncType = 'BASIC' | 'STRUCTURED';
 
@@ -21,13 +22,11 @@ export class GroupSyncDetailDialogComponent implements OnInit {
   theme: string;
   loading = true;
   group: RichGroup;
-  private syncEnabled: string;
-  private lastSyncState: string;
-  private lastSyncTime: string;
-  private structSyncEnabled: boolean;
-  private lastStructSyncState: string;
-  private lastStructSyncTime: string;
-  private type: SyncType;
+  syncInterval = '';
+  syncState = '';
+  syncTime = '';
+  syncType = '';
+  type: SyncType;
 
   constructor(
     public dialogRef: MatDialogRef<GroupSyncDetailDialogComponent>,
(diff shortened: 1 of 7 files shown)


bodnara (changes requested):
The interval is displayed properly, well done on that. However this dialog seems to be quite old and looks inconsistent with the reset of the app. I would like to ask you if you would be so kind and implement following changes:
image

- Change **all** the rows to be a input mat-form-field as shown in the screenshot (readonly), also remove ":" from the placeholder
- Change the order and color of buttons as per screenshot
- Remove redundant methods/elements that are in this component

libs/perun/dialogs/src/lib/group-sync-detail-dialog/group-sync-detail-dialog.component.ts
```diff
@@ -96,26 +83,6 @@ export class GroupSyncDetailDialogComponent implements OnInit {
     return this.type === 'STRUCTURED';
   }
```
mattjoke, line 85: I think `isBasic` and `isStructured` methods are not used. I think you should remove them

mattjoke (commented):
Otherwise I think it looks good.

apps/admin-gui/src/app/vos/pages/group-detail-page/group-detail-page.component.html
```diff
@@ -43,16 +43,12 @@
...
             , {{'GROUP_DETAIL.SYNCHRONIZATION' | translate}}:
             <span
               [matTooltipPosition]="'above'"
               [matTooltip]="group | groupSyncToolTip | translate">
-              <button
-                (click)="onSyncDetail()"
-                *ngIf="syncAuth"
-                [disabled]="(group | groupSyncIcon) === 'sync_disabled'"
-                mat-icon-button>
+              <button (click)="onSyncDetail()" mat-icon-button>
```
bodnara, line 51: Here we lost the information if user is authorized to force group synchornization. We can move this check to the dialog itself so we can still open it but cannot perform the action.

bodnara (changes requested):
There are couple of improvements:
image

- The synchronization field is not translated
- The dialog width is not set up, please set some reasonable width (450px maybe) when opening the dialog
- Right now the logic of sync and syncStructured is duplicated (2 buttons, 2 methods), I guess just merge it to one
- Last thing, I know the last sync time has the same format as before, but maybe we could use something more readable (this point more of a reference thing, if you think the format is fine than leave it as is)

xflord:
> Last thing, I know the last sync time has the same format as before, but maybe we could use something more readable (this point is more of a preference thing, if you think the format is fine than leave it as is)

For now, I just removed the milliseconds and we can discuss it further tomorrow.
Fixed the rest

libs/perun/dialogs/src/lib/group-sync-detail-dialog/group-sync-detail-dialog.component.ts
```diff
@@ -41,33 +40,21 @@ export class GroupSyncDetailDialogComponent implements OnInit {
...
-  }
-
   onCancel(): void {
     this.dialogRef.close(null);
   }
 
   onForce(): void {
     this.loading = true;
-    if (this.isBasic()) {
+    if (this.type === 'BASIC') {
```
HejdaJakub, line 49: Again please replace the deprecated subscription by the recommended observer argument `{next: (), error: ()}`

HejdaJakub (commented):
Just one suggestion, it looks good otherwise.

mattjoke (approved):
Looks good :+1:

bodnara (approved):

HejdaJakub (approved):

PR #1183, opened 2022-11-10
feat(lib): universal header component

sarkapalkovicova:
Replaces headers in all apps with new universal header component.


libs/perun/components/src/lib/perun-header/perun-header.component.ts
```diff
@@ -0,0 +1,123 @@
...
+  selector: 'perun-web-apps-header',
+  templateUrl: './perun-header.component.html',
+  styleUrls: ['./perun-header.component.scss'],
+})
+export class PerunHeaderComponent implements OnInit {
+  @Output() sidenavToggle = new EventEmitter();
+  @Input() hideToggle = false;
+  @Input() hideLanguageMenu = false;
+  @Input() hideHeaderMenu = false;
+  @Input() hideNotifications = true;
```
bodnara, line 31: Perhaps it would be better if these served as `showXYZ`. The reason being in case we add some component that is specific for one app we don't need to hide it on all others just show for one.

apps/password-reset/src/app/app.component.html
```diff
@@ -1,5 +1,9 @@
+<script src="../../../libs/perun/components/src/lib/perun-header/perun-header.component.ts"></script>
```
bodnara, line 1:
```suggestion
```

bodnara (changes requested):

bodnara (approved):

libs/perun/components/src/lib/perun-header/perun-header.component.html
```diff
@@ -0,0 +1,67 @@
...
+    <p [ngStyle]="{'color': textColor}" class="mt-auto mb-auto hide-label page-subtitle">
+      {{label}}
+    </p>
+  </div>
+
+  <div *ngIf="isDevel" id="nav-devel-info">
+    <span class="text-warning"> ~ 🗲&nbsp; </span>
+    <span class="text-danger"> DEVEL </span>
+    <span class="text-warning"> 🗲 ~ </span>
+  </div>
```
HejdaJakub, line 29: I am not 100 % sure, but since we don't have this label on our devel, this div is maybe no longer necessary. If it is true, then you can delete also the property in ts file, the property is_devel in ConfigProperties.ts, and also in all default configs.
```suggestion
```

libs/perun/components/src/lib/perun-header/perun-header.component.ts
```diff
@@ -0,0 +1,123 @@
...
+import { MatDialog } from '@angular/material/dialog';
+import { ShowNotificationHistoryDialogComponent } from '@perun-web-apps/perun/dialogs';
+
+export enum AppType {
+  Admin = 'admin',
+  Profile = 'profile',
+  PwdReset = 'pwdReset',
+  Consolidator = 'consolidator',
+  Linker = 'linker',
+}
```
HejdaJakub, line 19: It would be also fine if you created this enum/type in `libs/perun/models/src/lib/` and reuse it here and also in the otherApplicationService. Now the method `getUrlForOtherApplication()` gets as a parameter `appType`, but the type of this parameter is defined again - there can be used the same type imported from `models`.

PR #1184, opened 2022-11-11
feat(admin) add group assignment to create sponsored member dialog

xflord:
* user can now assign groups while creating sponsored member
* refactored the group assignment step of the dialog

### DIFF
--- a/apps/admin-gui/src/app/shared/components/assign-groups-sponsored-members/assign-groups-sponsored-members-component.ts
+++ b/apps/admin-gui/src/app/shared/components/assign-groups-sponsored-members/assign-groups-sponsored-members-component.ts
@@ -1,0 +1,188 @@
+import { Component, EventEmitter, Input, OnDestroy, OnInit, Output } from '@angular/core';
+import { SelectionModel } from '@angular/cdk/collections';
+import {
+  AttributesManagerService,
+  Group,
+  GroupsManagerService,
+  RichGroup,
+} from '@perun-web-apps/perun/openapi';
+import { GuiAuthResolver } from '@perun-web-apps/perun/services';
+import { TABLE_VO_GROUPS } from '@perun-web-apps/config/table-config';
+import { Observable, Subscription } from 'rxjs';
+import { Urns } from '@perun-web-apps/perun/urns';
+import { hasBooleanAttributeEnabled, isGroupSynchronized } from '@perun-web-apps/perun/utils';
+
+@Component({
+  selector: 'app-assign-groups-sponsored-members-component',
+  templateUrl: 'assign-groups-sponsored-members-component.html',
+  styleUrls: ['assign-groups-sponsored-members-component.scss'],
+})
+export class AssignGroupsSponsoredMembersComponent implements OnInit, OnDestroy {
+  @Input() submit: Observable<void>;
+
+  @Input() voId: number;
+  @Output() groupsToAdd = new EventEmitter<number[]>();
+  @Output() submitAllowed = new EventEmitter<boolean>();
+
+  loading = false;
+  createGroupAuth: boolean;
+  assignableGroups: Group[] = [];
+  allVoGroups: Group[] = [];
+  manualMemberAddingBlocked = false;
+  groupAssignment = 'none';
+
+  selection = new SelectionModel<Group>(true, []);
+  name = '';
+  description = '';
+  asSubgroup = false;
+  parentGroup: Group = null;
+  filterValue = '';
+  tableId = TABLE_VO_GROUPS;
+  groupToCreate: Group = {
+    id: 0,
+    beanName: 'Group',
+  };
+  groupIds: number[] = [];
+  onSubmit: Subscription;
+
+  private groupAttrNames = [Urns.GROUP_SYNC_ENABLED, Urns.GROUP_BLOCK_MANUAL_MEMBER_ADDING];
+
+  constructor(
+    private groupsService: GroupsManagerService,
+    private guiAuthResolver: GuiAuthResolver,
+    private attributesService: AttributesManagerService
+  ) {}
+
+  ngOnInit(): void {
+    this.onSubmit = this.submit.subscribe(() => this.handleGroups());
+    this.selection.changed.subscribe(() => this.isSubmitAllowed());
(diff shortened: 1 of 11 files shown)


apps/admin-gui/src/app/shared/components/dialogs/generate-sponsored-members-dialog/generate-sponsored-members-dialog.component.ts
```diff
@@ -333,7 +224,7 @@ export class GenerateSponsoredMembersDialogComponent implements OnInit {
     return output;
   }
 
-  private onGenerate(): void {
+  private onGenerate(groupIds: number[]): void {
```
bodnara, line 227:
```suggestion
  onGenerate(groupIds: number[]): void {
```

bodnara (changes requested):
When generating sponsored members from csv the members are not created for some reason.

bodnara (approved):
Nice job. You were right the generation works just fine, I selected wrong namespace.

I have one note here for future. Sometimes we (at least I did couple of times) need a variable from child (in this case the `submitAllowed`) in the parent component.  So we emit each time the variable changes in child and have a "duplicate" variable for it, in parent. But it would maybe make more sense to just use the `@ViewChild` in parent and that way we can easily access child.

apps/admin-gui/src/app/shared/components/assign-groups-sponsored-members/assign-groups-sponsored-members-component.ts
```diff
@@ -0,0 +1,181 @@
...
+            this.groupToCreate.name,
+            this.groupToCreate.description
+          )
+          .subscribe(
+            (group) => {
+              this.groupIds.push(group.id);
+              this.groupsToAdd.emit(this.groupIds);
+            },
+            () => (this.loading = false)
+          );
```
HejdaJakub, line 142: Do not use the deprecated subscription variant in the new code, please (more places).
```suggestion
          .subscribe({
            next: (group) => {
              this.groupIds.push(group.id);
              this.groupsToAdd.emit(this.groupIds);
            },
            error: () => (this.loading = false)
          });
```

apps/admin-gui/src/app/shared/components/assign-groups-sponsored-members/assign-groups-sponsored-members-component.ts
```diff
@@ -0,0 +1,181 @@
...
+        this.groupAssignment === null ||
+        (this.groupAssignment === 'existing' && this.selection.selected.length === 0) ||
+        (this.groupAssignment === 'new' &&
+          (this.groupToCreate.name.length === 0 || this.groupToCreate.description.length === 0)) ||
+        (this.asSubgroup && this.groupToCreate.parentGroupId === null)
+      )
+    );
+  }
+
+  handleGroups(): void {
```
HejdaJakub, line 126: I am not sure where exactly, but probably somewhere in this method you should clear the groups (group Ids) when you submit the last step (or on an error) because when for example user type an already existing login, the last step will fail, but then if I try it again to add to another group, for example, the previous group is also presented (tested on the use case when I tried to create new group, failed on login, try again with correct login BUT without group -> failed that group already exists)

HejdaJakub (changes requested):
There is a bug when I list some groups in the dialog and click on some row, the routing (in the background doesn't stop). I know that it is not directly related to this PR, but I'll be happy if you fix it. It took me a while since I found the problem, but in the end, it is just a triviality. In the group-list component, there are two similar properties `disableRouting` (correct one) and `disabledRouting`. We want to use in the HTML only the first property (`disableRouting`) and the second we can remove.

Please make the dialog wider. When you added the fourth step, the labels in the stepper are to long to for current wide.
image

apps/admin-gui/src/app/shared/components/assign-groups-sponsored-members/assign-groups-sponsored-members-component.ts
xflord: When an error occurred and the stepper reloaded, a new instance of the groups component got initialized, hence duplicating the subscription to the submit button (and the groups). Added onDelete method that should deal with it

xflord:
> We want to use in the HTML only the first property (disableRouting) and the second we can remove

We can't remove the second property, since it serves to stop routing when opening dialogs through the list (the expiration dialog In member -> groups for example). The problem was that the second property needed to be initialized to the same value as the first one.

apps/admin-gui/src/app/shared/components/assign-groups-sponsored-members/assign-groups-sponsored-members-component.ts
```diff
@@ -0,0 +1,188 @@
...
+  private groupAttrNames = [Urns.GROUP_SYNC_ENABLED, Urns.GROUP_BLOCK_MANUAL_MEMBER_ADDING];
+
+  constructor(
+    private groupsService: GroupsManagerService,
+    private guiAuthResolver: GuiAuthResolver,
+    private attributesService: AttributesManagerService
+  ) {}
+
+  ngOnInit(): void {
+    this.submit.subscribe(() => this.handleGroups());
```
HejdaJakub, line 57: You probably forgot to assign the subscription to the property you are trying to unsubscribe in the `ngOnDestroy` block. I tested it and after this assignment, it looks better.
```suggestion
    this.onSubmit = this.submit.subscribe(() => this.handleGroups());
```

HejdaJakub (changes requested):

libs/perun/components/src/lib/groups-list/groups-list.component.ts
```diff
@@ -129,6 +137,10 @@ export class GroupsListComponent {
     this.displayButtons = window.innerWidth > 800;
   }
 
+  ngOnInit(): void {
+    this.disabledRouting = this.disableRouting;
```
HejdaJakub, line 141: I realized that this could also cause some problems. For example, in vo-groups there is routeAuth which is not executed until some async call is done. Now, the routing is disabled on this page because the property `[disableRouting]="!routeAuth"` is undefined.

HejdaJakub (approved):

bodnara (approved):

PR #1186, opened 2022-11-14
fix(admin): show styles correctly on preview for HTML form items

HejdaJakub:
* In the preview column in the application form list and also on the preview page there are now the HTML items displayed correctly with CSS styles.

### DIFF
--- a/apps/admin-gui/src/app/vos/components/application-form-list/application-form-list.component.html
+++ b/apps/admin-gui/src/app/vos/components/application-form-list/application-form-list.component.html
@@ -96,10 +96,11 @@
                   value="{{'VO_DETAIL.SETTINGS.APPLICATION_FORM.CLEAR_SELECTION' | translate}}" />
               </form>
             </div>
-            <div
-              *ngIf="applicationFormItem.type ==='HEADING' ||
-                      applicationFormItem.type === 'HTML_COMMENT'">
+            <div *ngIf="applicationFormItem.type ==='HEADING'">
               <span [innerHTML]="getLocalizedLabel(applicationFormItem)"></span>
+            </div>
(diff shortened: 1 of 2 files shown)


mattjoke (approved):

apps/admin-gui/src/app/vos/components/application-form-list/application-form-list.component.html
```diff
@@ -96,11 +96,12 @@
...
               </form>
             </div>
-            <div
-              *ngIf="applicationFormItem.type ==='HEADING' ||
-                      applicationFormItem.type === 'HTML_COMMENT'">
+            <div *ngIf="applicationFormItem.type ==='HEADING'">
               <span [innerHTML]="getLocalizedLabel(applicationFormItem)"></span>
             </div>
+            <div *ngIf="applicationFormItem.type === 'HTML_COMMENT'">
+              <span [innerHTML]="getLocalizedLabel(applicationFormItem) | sanitizeHtml"></span>
```
xflord, line 103: From what I understand, `HTML_COMMENT` is any html the user inputs, is it safe to call `bypassSecurityTrustHtml` here?

HejdaJakub: Yes, I already discussed it with the Product Owner and the main problem is a few steps before displaying it in GUI when we allow saving potentially unsafe HTML to the database. We will figure it out in the next Sprint(s) and we should solve it on the front end and also on the back end. But the PO said that we can display it at this moment when we already have the HTML in the database.

xflord: Ok, seems good then

xflord (approved):

PR #1188, opened 2022-11-15
fix(admin): open the correct new window after the mouse middle click

HejdaJakub:
* There was the main problem when a user was logged in by service-access and tried to open the new window with the mouse middle click.
* The problem when the new window was redirected to the home page was fixed and also was fixed the problem when the new window didn't reflect the current type of the authentication.
* Some other small bugs related to the routing were fixed and also some refactoring was accomplished.

### DIFF
--- a/libs/perun/pipes/src/lib/select-application-link.pipe.ts
+++ b/libs/perun/pipes/src/lib/select-application-link.pipe.ts
@@ -1,0 +1,45 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { Application, Group, Member } from '@perun-web-apps/perun/openapi';
+
+@Pipe({
+  name: 'selectApplicationLink',
+})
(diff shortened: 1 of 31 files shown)


libs/perun/pipes/src/lib/select-application-link.pipe.ts
```diff
@@ -0,0 +1,52 @@
...
+   * @param group group if it os a group application
+   * @param member member if it should be a link for a member application
+   */
+  transform(
+    application: Application,
+    disableRouting: boolean,
+    group: Group,
+    member: Member
+  ): string[] {
+    if (!disableRouting) {
```
bodnara, line 22: Please return early if routing is disabled so there is no nesting and hopefully help readability.

bodnara (changes requested):

bodnara (approved):

PR #1191, opened 2022-11-21
fixup! fix(admin): open the correct new window after the mouse middle click

HejdaJakub:
* Fixed incorrect if condition.
* Added class for pointer in application-dynamic-list component.


bodnara (approved):
Good catch!

PR #1192, opened 2022-11-22
fix: preserve query parameters during the routing

HejdaJakub:
* Programmatic navigation in all ts files was fixed by merging or preserving query parameters.
* Correct navigation with all query params by <a> tag and <tr> tag via [routerLink] is managed by the new directive.
* Special use cases were handled like navigation between the same entities (e.g. hierarchical organizations) and do NOT preserve query params by redirecting from the preview page, where the query parameters are used for the definition of application items.

### DIFF
--- a/apps/admin-gui/src/app/admin/pages/admin-page/admin-services/service-detail-page/service-detail-page.component.html
+++ b/apps/admin-gui/src/app/admin/pages/admin-page/admin-services/service-detail-page/service-detail-page.component.html
@@ -17,8 +17,7 @@
               [perunWebAppsMiddleClickRouterLink]="['/admin/services', service.id.toString()]"
               (auxclick)="$event.preventDefault()"
               [routerLink]="['/admin/services', service.id]"
-              class="service-link"
-              queryParamsHandling="merge">
+              class="service-link">
               {{service.name}}
             </a>
             <span class="text-muted"> &nbsp; #{{service.id}} </span>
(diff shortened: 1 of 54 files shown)


apps/admin-gui/src/app/shared/side-menu/side-menu-router.service.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  navigate(url: string[]): void {
+    this.route.queryParams.pipe(take(1)).subscribe((params) => {
+      let paramsHandlingMethod: QueryParamsHandling = 'merge';
+      const queryParams = Object.assign({}, params);
+
+      if (location.pathname.endsWith('applicationForm/preview')) {
+        paramsHandlingMethod = '';
+        delete queryParams.applicationFormItems;
+      }
```
bodnara, line 19: The problem here is that we can navigate await from the preview page by different means than just side menu e.g. links on vo detail. And different forms preserve the application form items params.

bodnara (changes requested):
There is another issue with deleting the query params and then navigating back with the back button.

apps/admin-gui/src/app/shared/side-menu/side-menu-router.service.ts
HejdaJakub: I hope this should fix the problem with another type of navigation from the preview page (fixed for home button, on vo application preview for the vo link and the link for member/parent vo, on the group application preview for the group link and vo link).

Johaney-s (approved):

bodnara (approved):
Seems to work correctly. We might revisit this topic once we stop using `[perunWebAppsForceRouterLink]`.

PR #1193, opened 2022-11-22
feat(admin): filter audit messages by event type

sarkapalkovicova:
Added input for search string according to which messages will be filtered.


bodnara:
Has to be merged **AFTER** https://github.com/CESNET/perun/pull/3777

Johaney-s (commented):
Looks good, not sure the changes in angular.json should be part of this pr. Could we have a "remove filter" button next to the selection? It might be easier to delete the selection that way than to scroll the list and search for the selected events while searching for another. What do you think?

apps/admin-gui/src/app/admin/pages/admin-page/admin-audit-log/admin-audit-log.component.html
```diff
@@ -1,3 +1,19 @@
...
+    [entities]="eventOptions"
+    [searchFunction]="searchFunction"
+    (entitySelected)="toggleEvent($event)"
+    [mainTextFunction]="mainTextFunction"
+    [secondaryTextFunction]="secondaryTextFunction"
+    [multiple]="true"
+    [disableAutoSelect]="true"
+    [fullWidth]="false"
+    [moveSelectedUp]="true"
+    [selectPlaceholder]="'ADMIN.AUDIT_LOG.SEARCH_DESCRIPTION'|translate"></perun-web-apps-entity-search-select>
```
bodnara, line 14: Entity search select is the base class for search select components. We don't use it directly but rather create a specification e.g. vo search select. I would like to ask if you could please create such a specification also for audit events.

apps/admin-gui/src/app/admin/pages/admin-page/admin-audit-log/admin-audit-log.component.ts
```diff
@@ -1,16 +1,49 @@
...
+
   refreshTable(): void {
     this.refresh = !this.refresh;
   }
+
+  searchFunction = (entity: string): string => entity;
+
+  mainTextFunction = (entity: string): string => entity;
+
+  secondaryTextFunction = (_: string): string => '';
```
bodnara, line 36: As I mentioned, define these in the search select specification. There are also examples of methods that don't need the parameter.

apps/admin-gui/src/app/admin/pages/admin-page/admin-audit-log/admin-audit-log.component.ts
```diff
@@ -1,16 +1,49 @@
...
+    if (this.selectedEvents.includes(event)) {
+      const index = this.selectedEvents.indexOf(event);
+      this.selectedEvents.splice(index, 1);
+    } else {
+      this.selectedEvents.push(event);
+    }
+
+    // change this array to force refresh of component and calling endpoint
+    this.selectedEvents = [...this.selectedEvents];
+  }
```
bodnara, line 48: Only propagate the changes after user closes the selection. That way we only refresh the table once users select all events they need. You can take a look of how it is done for selecting member status on vo member page.

libs/perun/components/src/lib/entity-search-select/entity-search-select.component.html
```diff
@@ -1,4 +1,4 @@
-<mat-form-field class="w-100">
+<mat-form-field class="{{ fullWidth ? 'w-100' : '' }}">
```
bodnara, line 1: I understand why we would like this change but we can achieve this in little bit more elegant way. Keep the class here as before, but in the parent component put the the search select inside a container of fixed width (e.g. 450px). The search select will take up 100% of it's parent (here by parent I mean element like div). This way we don't need another input parameter.

libs/perun/components/src/lib/entity-search-select/entity-search-select.component.ts
```diff
@@ -58,10 +62,6 @@ export class EntitySearchSelectComponent<T extends PerunBean>
   ) => entity.status;
 
   ngOnInit(): void {
-    this.entitiesCtrl.valueChanges
-      .pipe(distinctUntilChanged())
-      .subscribe((entity: T) => this.entitySelected.emit(entity));
```
bodnara, line 63: This code prevents emitting of the same value multiple times. This is quite important and can cause errors in some places. If you insist on removing it, make sure you handle it the onChange method.

bodnara (changes requested):

Johaney-s (approved):

bodnara (approved):
Really good job.
I have one suggestion/question we have this notion of advaced filtering and there we have this button to clear the filters. Would it make sense to use the same one here ? If yes could you please change the icon and or the text here ? Otherwise this can be merged.
image

PR #1199, opened 2022-11-29
feat(admin): Bans on entities

bodnara:
* it is now possible to manage (create, delete, update) bans on Vo, Resource and Facility
* created page to manage bans for Vo, Resource and Facility
* created page to overview bans for Member and User


Johaney-s (commented):
- update vo ban dialog doesn't work
- valid to (Jan 1, 2999) looks a bit weird if no date is set
- cannot filter bans by name, description
- cannot filter user's bans by description, object's name
- cannot unset once set validity in the setting/updating ban dialog
- it is not very clear there are some fields (description, validity) left under the member's selection
- shouldn't different kind of checkbox be used if only one member is permitted to be selected?
- getting `'banOnXY.getValidityTo' is null ` error if didn't set validity for resource/vo ban. probably backend bug though.

xflord:
Screenshot from 2022-11-29 10-29-23
When adding a ban do Facility/Resource WITHOUT setting the expiry date, an exception is thrown (unlike in Vos, where the expiration is set automatically). Please either disable the add button or also set the expiration automatically here.

libs/perun/components/src/lib/ban-on-entity-list/ban-on-entity-list.component.ts
```diff
@@ -0,0 +1,170 @@
...
+    switch (column) {
+      case 'banId':
+        return String(data.ban.id);
+      case 'subjectName': {
+        if (this.isFacilityBan(data)) return data.facility.name;
+        else if (this.isResourceBan(data)) return data.resource.name;
+        else return data.vo.name;
+      }
+      default: {
+        return String(data[column]);
```
xflord, line 95: This is why filtering/sorting of the list fails IMO, column names don't match the names of the attributes

xflord (changes requested):

xflord:
>     * valid to (Jan 1, 2999) looks a bit weird if no date is set
> 
>     * cannot unset once set validity in the setting/updating ban dialog

I think it would be nice to be able to set the expiration as 'never', similarly to how membership works. However, this would require further discussion and changes to backend, I think we can omit it from this task. Though I would at least either force the user to set a date or set it to something sensible (like a year maybe?), because 2999 really looks odd.


>     * it is not very clear there are some fields (description, validity) left under the member's selection

Strongly agree with this.

xflord:
Unbanning multiple members/users at once should be made possible as well.

bodnara:
- update Vo dialog was restored (it got lost in the depths of git)
- validity set to (1. Jan 2999) is displayed and treated as "never"
- filtering in ban table now uses correct properties for each column
- specification for the ban (validity, description) was moved to a second step for more clarity

Using the add ban dialog to create ban for User/Member if one already exists will inform the user that ban already exists but will update the ban when dialog is submitted instead of throwing an error. Various labels were changed to further emphasize that update was performed in this use case.

> Unbanning multiple members/users at once should be made possible as well.

All actions (create, update, remove) can take only a single ban, so this is not possible right now. We also don't to do chain call so it is consistent for step-up authentication.

xflord:
- Filtering/sorting by Member name still doesn't work
- When choosing the specific date option, without actually setting a date and then pressing `Add`sets it to 'never'. Not sure whether this was intended, seems quite unintuitive to me. (Maybe disable the button until a date is chosen?)
Screenshot from 2022-11-30 11-35-39

-  The name of the column is 'Validity to' and the value is 'never'. 'Validity to never' doesn't make much sense.

Johaney-s (approved):

PR #1204, opened 2022-12-07
feat(admin): add option to copy members from one group to other group…

xflord:
…s in the same vo

* added dialog to copy members
* updated openapi

### DIFF
--- a/apps/admin-gui/src/assets/i18n/en.json
+++ b/apps/admin-gui/src/assets/i18n/en.json
@@ -664,6 +664,7 @@
       "ADD_MEMBER": "Add",
       "ADD_MEMBER_DISABLED": "Manual adding of members is blocked for this group.",
       "REMOVE_MEMBERS": "Remove",
+      "COPY_MEMBERS": "Copy",
       "SEARCH_DESCRIPTION": "Search by name, login, email or UUID",
       "LIST_ALL": "List all members",
       "INVITE": "Invite",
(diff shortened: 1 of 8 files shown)


xflord:
Don't merge until relevant changes to BE are merged.

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -1175,6 +1176,20 @@
       "NAME_COLUMN": "Name",
       "ERROR_COLUMN": "Error"
     },
+    "COPY_MEMBERS": {
+      "TITLE": "Copy members",
+      "DISABLED_COPY_SELECTION": "No members selected",
+      "COPY_SELECTION": "Copy selected members of this group",
+      "COPY_ALL": "Copy all members of this groups",
```
Johaney-s, line 1183:
```suggestion
      "COPY_ALL": "Copy all direct members of this groups",
```

Johaney-s (commented):
looks nice, maybe this tooltip could be extended to "... cannot be removed or copied"
Screenshot from 2022-12-12 08-48-14(1).

Johaney-s (approved):

mattjoke (approved):

bodnara (commented):
Other than these small comments I think the use-case can be a little confusing at first. We could discuss the UX aspects of  the dialog. My general idea is to maybe separate the steps to select members (all, or manual selection) and selection of group. Maybe it would be usefull to allow selection of members in the dialog itself. By selecting the manual option a table would appear ? Because right now if you just open the dialog there is a disable option that say nothing is selected by there is no option to select.

PR #1205, opened 2022-12-09
feat(admin): allow expiration change through member status icon

HejdaJakub:
* On a member detail is membership section (for vo and group) which allows to change status or expiration. The change status dialog opened after clicking on status icon in the dynamic member tyble was replaced by new dialog with this membership functionality.


xflord (commented):
- The title of the dialog ("Group/Organization membership") looks weird to me, I would personally center it. 
- The 'back' button of the expiration dialog for group status closes the dialogs instead of going back 
- The list doesn't refresh after setting the Vo status, but not setting the expiration and closing the dialog using the back button.

Johaney-s (commented):
One thing that distracts me is the alignment of the dialog - both the status and expiration setting dialogues are aligned to the left, while this one is aligned to the center. I would align the general dialogue to the left too.

HejdaJakub:
- I left the title on the left side (other dialogs also have the title on the left), but I changed the width of dialog and also the title font size, so now the title dialog is the same as in other dialogs.
- Back button functionality fixed.
- Refreshing for this use case fixed.
- Alignment of the dialog content fixed.

xflord:
Works fine now but this happens on my browser 
Screenshot from 2022-12-13 13-28-35
Maybe make the dialog a tiny bit larger?

xflord (approved):

Johaney-s (approved):

PR #1206, opened 2022-12-13
feat(admin): add option to send request with non authentication to se…

xflord:
…rvice methods

* all methods from openapi are now generated with optional parameter 'useNon', which if set to true sends the request to the backend server as 'non' instead of the usual (oauth, krb...)
* this was achieved by modifying the default openapi-generator template, see https://openapi-generator.tech/docs/templating/ for more info
* used the new 'useNon' parameter in methods for password reset via token and password strength check, fixing the password reset page when using token.

### DIFF
--- a/.openapi-generator/templates/api.service.mustache
+++ b/.openapi-generator/templates/api.service.mustache
@@ -1,0 +1,396 @@
+{{>licenseInfo}}
+/* tslint:disable:no-unused-variable member-ordering */
+
+import { Inject, Injectable, Optional }                      from '@angular/core';
+import { HttpClient, HttpHeaders, HttpParams,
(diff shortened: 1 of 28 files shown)


bodnara (commented):
Really like this solution, very elegant nice job.

libs/perun/services/src/lib/dynamic-paginating.service.ts
```diff
@@ -102,6 +102,7 @@ export class DynamicPaginatingService {
         offset: pageSize * pageNumber,
         pageSize: pageSize,
         order: order,
+        selectedEvents: [],
```
HejdaJakub, line 105: Please rebase it. After rebase I think this should use one of parameters (selectedEvents: string[]):
```suggestion
        selectedEvents: selectedEvents,
```

HejdaJakub (commented):
Good job. I think it works really well.

PR #1208, opened 2022-12-19
fix(admin) disable copy members button when there aren't any members …

xflord:
…to copy

* check number of group direct members before enabling copy button
* updated openapi


xflord:
merge after https://github.com/CESNET/perun/pull/3812

HejdaJakub (commented):
Please re-run `openapi-gen` to fix build error. It looks good otherwise.

libs/perun/openapi/src/lib/api/databaseManager.service.ts
```diff
@@ -448,7 +445,6 @@ export class DatabaseManagerService {
       path[1] = 'non';
       helperUrl.pathname = path.join('/');
       requestUrl = helperUrl.toString();
-      console.log(requestUrl);
```
Kuliak, line 451: Is there any specific reason why there were all those console logs before and were removed?

xflord: The PR in which 'useNon' was implemented was probably merged with openapi generated by the test template. This isn't a big issue, since a new generation of openapi with the correct template (which was merged) will no longer have this log (hence these changes).

xflord:
Added new field isCopyDisabled which is changed upon adding/removing members and then disabled the button based on its value. Also added tooltip when disabled

apps/admin-gui/src/app/vos/pages/group-detail-page/group-members/group-members.component.ts
```diff
@@ -264,6 +268,18 @@ export class GroupMembersComponent implements OnInit {
...
+  isCopyMembersDisabled(): void {
+    this.groupService.getGroupDirectMembersCount(this.group.id).subscribe({
+      next: (count) => {
+        this.copyDisabled = count === 0;
+      },
+      error: () => {
+        this.copyDisabled = true;
+      },
+    });
+    this.copyDisabled = false;
```
HejdaJakub, line 280: NITPICK: I would move line `this.copyDisabled = false;` to the beginning of this method before the async call due to readability. It's probably clearer to set the variable to false and wait for the async call instead of running an async call (skip waiting), set the variable to `false`, and then rewrite the variable according to the async response.

HejdaJakub (approved):

libs/perun/pipes/src/lib/group-members-action-button-disabled-tooltip.pipe.ts
```diff
@@ -7,7 +7,10 @@ import { Urns } from '@perun-web-apps/perun/urns';
   name: 'groupMembersActionButtonDisabledTooltip',
 })
 export class GroupMembersActionButtonDisabledTooltipPipe implements PipeTransform {
-  transform(group: RichGroup, button?: 'add'): string {
+  transform(group: RichGroup, button?: 'add' | 'copy'): string {
+    if (button === 'copy') {
+      return 'GROUP_DETAIL.MEMBERS.COPY_MEMBERS_DISABLED';
+    }
```
bodnara, line 13: It seems to me that the copy action is not affected by the other cases. In that case we could just use a direct translation for the tooltip.

libs/perun/pipes/src/lib/group-members-action-button-disabled-tooltip.pipe.ts
```diff
@@ -7,7 +7,7 @@ import { Urns } from '@perun-web-apps/perun/urns';
   name: 'groupMembersActionButtonDisabledTooltip',
 })
 export class GroupMembersActionButtonDisabledTooltipPipe implements PipeTransform {
-  transform(group: RichGroup, button?: 'add'): string {
+  transform(group: RichGroup, button?: 'add' | 'copy'): string {
```
bodnara, line 10:
```suggestion
  transform(group: RichGroup, button?: 'add'): string {
```

bodnara (approved):

HejdaJakub (approved):

PR #1209, opened 2022-12-19
feat(admin): user/group can have (un)set roles as a bulk operation

Kuliak:
* pipes for transforming between EnrichedFacility and Facility created
* admin can select multiple objects for user/group to gain roles/lose roles in
* searching of objects supported

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/add-role-dialog/add-role-dialog.component.html
+++ b/apps/admin-gui/src/app/shared/components/dialogs/add-role-dialog/add-role-dialog.component.html
@@ -1,35 +1,49 @@
 <div class="{{theme}}">
-  <h1 mat-dialog-title>{{'DIALOGS.ADD_ROLE.TITLE' | translate}}</h1>
+  <div mat-dialog-title>
+    <h1 mat-dialog-title>{{'DIALOGS.ADD_ROLE.TITLE' | translate}}</h1>
+  </div>
   <mat-spinner *ngIf="loading" class="mr-auto ml-auto"></mat-spinner>
   <div *ngIf="!loading" mat-dialog-content class="dialog-container">
-    <perun-web-apps-role-search-select
-      [roles]="rules"
-      (roleSelected)="selectedRule = $event; selectedEntity = null">
+    <mat-spinner *ngIf="loading" class="ml-auto mr-auto"></mat-spinner>
+    <perun-web-apps-role-search-select [roles]="rules" (roleSelected)="resetSelection($event)">
     </perun-web-apps-role-search-select>
-    <perun-web-apps-vo-search-select
-      *ngIf="selectedRule?.primaryObject === 'Vo'"
+    <perun-web-apps-immediate-filter
+      *ngIf="selectedRule?.primaryObject"
+      (filter)="filterValue = $event"
+      [placeholder]="'SHARED_LIB.PERUN.COMPONENTS.UNIVERSAL_OBJECTS_LIST.FILTER'"
+      class="font-size-1rem"></perun-web-apps-immediate-filter>
+    <perun-web-apps-vos-list
+      [filterValue]="filterValue"
+      *ngIf="!loading && selectedRule?.primaryObject === 'Vo'"
       [vos]="vos | manageableEntities: selectedRule | unassignedRole: roles : selectedRule"
-      [disableAutoSelect]="true"
-      (voSelected)="selectedEntity = $event">
-    </perun-web-apps-vo-search-select>
-    <perun-web-apps-group-search-select
-      *ngIf="selectedRule?.primaryObject === 'Group'"
+      [displayedColumns]="['checkbox', 'id', 'shortName', 'name']"
+      [selection]="selected"
+      [disableRouting]="true">
+    </perun-web-apps-vos-list>
+    <perun-web-apps-groups-list
+      [filterValue]="filterValue"
+      *ngIf="!loading && selectedRule?.primaryObject === 'Group'"
       [groups]="groups | manageableEntities: selectedRule | unassignedRole: roles : selectedRule"
-      [disableAutoSelect]="true"
-      (groupSelected)="selectedEntity = $event">
-    </perun-web-apps-group-search-select>
-    <perun-web-apps-facility-search-select
-      *ngIf="selectedRule?.primaryObject === 'Facility'"
-      [facilities]="facilities | manageableEntities: selectedRule| unassignedRole: roles : selectedRule"
-      [disableAutoSelect]="true"
-      (facilitySelected)="selectedEntity = $event">
-    </perun-web-apps-facility-search-select>
-    <perun-web-apps-resource-search-select
-      *ngIf="selectedRule?.primaryObject === 'Resource'"
+      [displayedColumns]="['select', 'id', 'vo', 'name', 'description']"
+      [selection]="selected"
+      [disableRouting]="true">
+    </perun-web-apps-groups-list>
+    <perun-web-apps-facilities-list
+      [filterValue]="filterValue"
+      *ngIf="!loading && selectedRule?.primaryObject === 'Facility'"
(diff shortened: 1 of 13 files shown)


apps/admin-gui/src/app/shared/components/dialogs/add-role-dialog/add-role-dialog.component.html
```diff
@@ -1,47 +1,66 @@
 <div class="{{theme}}">
-  <h1 mat-dialog-title>{{'DIALOGS.ADD_ROLE.TITLE' | translate}}</h1>
+  <div mat-dialog-title>
+    <h1 mat-dialog-title>{{'DIALOGS.ADD_ROLE.TITLE' | translate}}</h1>
+    <perun-web-apps-immediate-filter
+      *ngIf="selectedRule?.primaryObject"
+      (filter)="applyFilter($event)"
```
HejdaJakub, line 6: I know that we use it in many places in GUI like this, but you can simplify it (so you don't need a special method in the ts file anymore).
```suggestion
      (filter)="filterValue = $event"
```

apps/admin-gui/src/app/shared/components/dialogs/add-role-dialog/add-role-dialog.component.ts
```diff
@@ -54,9 +58,15 @@ export class AddRoleDialogComponent implements OnInit {
...
-        .getAllFacilities()
-        .subscribe({ next: (facilities) => (this.facilities = facilities) });
+      this.facilityService.getAllFacilities().subscribe({
+        next: (facilities) =>
+          (this.facilities = facilities.map((fac) => {
+            const ef: EnrichedFacility = {
+              facility: fac,
+            };
+            return ef;
+          })),
```
HejdaJakub, line 68: You are able to reuse existing pipe also in ts file.
```suggestion
          (this.facilities = new ToEnrichedFacilityPipe().transform(facilities)),
```

apps/admin-gui/src/app/shared/components/dialogs/add-role-dialog/add-role-dialog.component.ts
```diff
@@ -37,9 +38,12 @@ export class AddRoleDialogComponent implements OnInit {
 
   selectedRule: RoleManagementRules;
   selectedEntity: PerunBean = null;
+  selected = new SelectionModel<PerunBean>(true, []);
```
HejdaJakub, line 41: The property `selectedEntity` is NOT necessary anymore. If you agree, you can delete it also from the HTML file.
```suggestion
  selected = new SelectionModel<PerunBean>(true, []);
```

apps/admin-gui/src/app/users/pages/user-detail-page/user-dashboard/dashboard-card/dashboard-card.component.html
```diff
@@ -13,15 +13,17 @@
         [pageSizeOptions]="null"
         [displayedColumns]="['id', 'recent', 'shortName', 'name']"
         [recentIds]="this.recentIds"
-        [vos]="this.objects">
+        [vos]="this.objects"
+        [selection]="selected">
```
HejdaJakub, line 17: Since there isn't displayed column with checkboxes on the dashboard page, you don't have to put selection as a parameter here -> the same for all other lists here ... IMHO you can restore changes in the whole file and also in the ts file for the dashboard and it should still work well.
```suggestion
        [vos]="this.objects">
```

HejdaJakub (changes requested):
There is some bug in group-list:
image

apps/admin-gui/src/app/shared/components/dialogs/add-role-dialog/add-role-dialog.component.html
```diff
@@ -1,47 +1,63 @@
...
+      [placeholder]="'SHARED_LIB.PERUN.COMPONENTS.UNIVERSAL_OBJECTS_LIST.FILTER'"
+      class="font-size-1rem"></perun-web-apps-immediate-filter>
+  </div>
   <mat-spinner *ngIf="loading" class="mr-auto ml-auto"></mat-spinner>
   <div *ngIf="!loading" mat-dialog-content class="dialog-container">
-    <perun-web-apps-role-search-select
-      [roles]="rules"
-      (roleSelected)="selectedRule = $event; selectedEntity = null">
+    <mat-spinner *ngIf="loading" class="ml-auto mr-auto"></mat-spinner>
+    <perun-web-apps-role-search-select [roles]="rules" (roleSelected)="selectedRule = $event">
```
xflord, line 13: Did you consider having this search select above the object filter? I feel like the filter should be the first thing above the list, also this way the search select wouldn't disappear when scrolling the list (see screenshot)
Screenshot from 2022-12-23 13-04-56

apps/admin-gui/src/app/shared/components/dialogs/add-role-dialog/add-role-dialog.component.html
```diff
@@ -1,47 +1,63 @@
 <div class="{{theme}}">
-  <h1 mat-dialog-title>{{'DIALOGS.ADD_ROLE.TITLE' | translate}}</h1>
+  <div mat-dialog-title>
+    <h1 mat-dialog-title>{{'DIALOGS.ADD_ROLE.TITLE' | translate}}</h1>
+    <perun-web-apps-immediate-filter
+      *ngIf="selectedRule?.primaryObject"
+      (filter)="filterValue = $event"
```
xflord, line 6: When a filter value is typed in, but then the user swaps to a role that does not have the filter window and back to a role that does, the filter is empty, but the previous value is still used to filter the objects

apps/admin-gui/src/app/shared/components/dialogs/add-role-dialog/add-role-dialog.component.ts
```diff
@@ -36,10 +38,12 @@ export class AddRoleDialogComponent implements OnInit {
   @Output() submitForm = new EventEmitter<AddRoleForm>();
 
   selectedRule: RoleManagementRules;
-  selectedEntity: PerunBean = null;
+  selected = new SelectionModel<PerunBean>(true, []);
```
xflord, line 41: Since the selection does not clear itself on role change, if user chooses for example some groups under group admin, then swaps to resource admin instead and tries to set that role, an error is thrown

xflord (changes requested):
Also the routing in the add dialog should be disabled, but we discussed this at review so you are already aware of that.

HejdaJakub (changes requested):
There is still some inconsistency when you change roles and some filter for objects is applied. I found this problem for the group list and group roles. If I type some filter (e.g. "test") for the group list and then change the role to another role with an object filter (e.g. FACILITY ADMIN) and change the role immediately back to the previous group role, there is a different count of groups.
First filter count (244):
image
Count after the change (215):
image

xflord (approved):

apps/admin-gui/src/app/shared/components/dialogs/add-role-dialog/add-role-dialog.component.ts
```diff
@@ -74,6 +88,22 @@ export class AddRoleDialogComponent implements OnInit {
...
+    }
+  }
+
+  resetSelection(selectedRule: RoleManagementRules): void {
+    this.selectedRule = selectedRule;
+    this.selected.clear();
+    this.selectedFacilities.clear();
+    this.filterValue = '';
+    this.loadObjects();
+    this.filterComponent.formControl = new UntypedFormControl('');
```
HejdaJakub, line 107: 1) There is an error when you try to set the value to a non-existing element (the assignment is called before the element is created) -> so you want to change the value just for the existing element.
image
2) You don't have to replace form control by new form control. You can just change the value like this:
```suggestion
    if (this.filterComponent) {
      this.filterComponent.formControl.setValue('');
    }
```

HejdaJakub (changes requested):
Just a tiny last suggestion.

HejdaJakub (approved):

PR #1212, opened 2022-12-27
fix(admin): sort attributes list by display name

xflord:
* attributes and attribute definition lists are now sorted by display name by default
* added display name column to admin attribute definitions list
* the mentioned lists now display 25 entities on one page by default
* fixed the personal preferred table page size config (changes didn't take effect correctly before)

### DIFF
--- a/libs/config/table-config/src/lib/table-config.service.ts
+++ b/libs/config/table-config/src/lib/table-config.service.ts
@@ -5,7 +5,11 @@ import { GUIConfigService, LS_TABLE_PREFIX, PREF_PAGE_SIZE } from './guiconfig.s
   providedIn: 'root',
 })
 export class TableConfigService {
-  constructor(private guiConfigService: GUIConfigService) {}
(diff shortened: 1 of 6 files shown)


xflord:
Do not merge, we should discuss whether we really want to add the column.

libs/config/table-config/src/lib/table-config.service.ts
```diff
@@ -17,7 +21,7 @@ export class TableConfigService {
       return pref;
     }
 
-    return 10;
+    return this.defaultTableSizes.get(tableId) || 10;
```
bodnara, line 24: Personally I would prefer this over the `||` operator just to prevent potential confusion. Not sure if it's some javascript special but I assumed it would return a boolean and that it's a mistake.
```suggestion
    return this.defaultTableSizes.get(tableId) ?? 10;
```

bodnara (approved):

mattjoke (approved):

PR #1217, opened 2023-01-04
fix: don't display login page after user has been logged in

xkostka2:
* fixed bug where if user tried to log in by manually typing /service-access in the url, the login page would be displayed even after authentication (instead of homepage of the app)


xflord (approved):

bodnara (commented):
This works, but maybe we could set up a `canActivate` route guard to prevent the navigation if user is logged in.

PR #1223, opened 2023-01-10
ci: update semantic release version

Johaney-s:
* will resolve current release action warnings

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,9 +11,9 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -11,9 +11,9 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v3
```
xflord, line 14: Is there a reason not to update these also in the 'Build' workflow?

Johaney-s: thought you're updating it in the caching PR. will add

xflord (approved):

PR #1226, opened 2023-01-11
feat(core): HTML sanitizing and escaping

mattjoke:
* HTML inputs for notifications and application forms are sanitized and (optionally can be escaped)


apps/admin-gui/src/app/shared/components/dialogs/edit-application-form-item-dialog/edit-application-form-item-dialog.component.ts
```diff
@@ -9,11 +9,11 @@ import {
   Group,
   Type,
 } from '@perun-web-apps/perun/openapi';
-import { createNewApplicationFormItem } from '@perun-web-apps/perun/utils';
+import { createNewApplicationFormItem, escapeDangerousHtml } from '@perun-web-apps/perun/utils';
 import DisabledEnum = ApplicationFormItem.DisabledEnum;
 import HiddenEnum = ApplicationFormItem.HiddenEnum;
 import { ItemType, NO_FORM_ITEM, SelectionItem } from '@perun-web-apps/perun/components';
-import { StoreService } from '@perun-web-apps/perun/services';
+import { AuthService, StoreService } from '@perun-web-apps/perun/services';
```
xflord, line 16: unused import

xflord (commented):
Is there a reason not to use an already existing sanitizer library? For example [DOMPurify](https://github.com/cure53/DOMPurify) as provided in https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html ( @HejdaJakub hinted in the task description). If our requirements for the HTML sanitization aren't too specific, I believe using an existing library is safer since it gets updated when new vulnerabilities are found and so on (better not to reinvent the wheel).

I didn't really dig too deep into this, just making sure you considered this.

apps/admin-gui/src/app/shared/components/dialogs/edit-application-form-item-dialog/edit-application-form-item-dialog.component.ts
mattjoke: Good point; I must have overlooked it. The more I think about it, I should combine the DOMPurify and the specific property filter (coded in this PR).

libs/perun/utils/src/lib/perun-utils.ts
```diff
@@ -864,3 +865,33 @@ export function isMemberIndirectString(member: RichMember): string {
...
   return 'UNKNOWN';
 }
+
+export function escapeDangerousHtml(html: string): string {
+  const config = {
+    FORBID_TAGS: ['script', 'link', 'iframe', 'object', 'embed', 'frame', 'frameset', 'style'],
+  };
+  const sanitized = sanitize(html, config);
+
+  // Sanitize style properties (this is not supported by DOMPurify)
```
xflord, line 875: This doesn't seem right. Skimmed through the DOMPurify docs and it seemed like there is a way to sanitize attribute values using the 'uponSanitizeAttribute' hook. (See [here](https://github.com/cure53/DOMPurify/tree/main/demos#hook-to-allow-and-sand-box-all-javascript-link) for inspiration). However if you tested your method thoroughly and it works, fine by me.

libs/perun/utils/src/lib/perun-utils.ts
```diff
@@ -864,3 +865,33 @@ export function isMemberIndirectString(member: RichMember): string {
   }
   return 'UNKNOWN';
 }
+
+export function escapeDangerousHtml(html: string): string {
+  const config = {
+    FORBID_TAGS: ['script', 'link', 'iframe', 'object', 'embed', 'frame', 'frameset', 'style'],
```
xflord, line 871: Use whitelists instead of blacklists https://stackoverflow.com/questions/2475784/why-use-a-whitelist-for-html-sanitizing

libs/perun/utils/src/lib/perun-utils.ts
```diff
@@ -864,3 +865,33 @@ export function isMemberIndirectString(member: RichMember): string {
...
+  };
+  const sanitized = sanitize(html, config);
+
+  // Sanitize style properties (this is not supported by DOMPurify)
+  const bannedAttributes = ['calc', 'expression', 'url', 'javascript', 'vbscript', 'data'];
```
xflord, line 876: whitelist here as well

libs/perun/utils/src/lib/perun-utils.ts
mattjoke: I guess i did not see it there :+1: Probably will use this hook instead of my method.

libs/perun/utils/src/lib/perun-utils.ts
```diff
@@ -864,3 +865,120 @@ export function isMemberIndirectString(member: RichMember): string {
   }
   return 'UNKNOWN';
 }
+
+export function escapeDangerousHtml(html: string): string {
```
bodnara, line 869: I think this DOM sanitizing would be better off as a service instead of a util function. Otherwise it is nicely done.

bodnara (approved):

xflord (approved):

apps/admin-gui/src/app/shared/components/dialogs/edit-email-footer-dialog/edit-email-footer-dialog.component.ts
```diff
@@ -50,6 +51,12 @@ export class EditEmailFooterDialogComponent implements OnInit {
     this.loading = true;
     this.mailAttribute.value = this.mailFooter;
     this.htmlMailAttribute.value = this.htmlMailFooter;
+
+    // Remove unsafe tags
+    this.htmlMailAttribute.value = this.escapeInput.escapeDangerousHtml(
+      this.htmlMailAttribute.value as string
```
HejdaJakub, line 57:
```suggestion
      String(this.htmlMailAttribute.value)
```

libs/perun/services/src/lib/html-escape.service.ts
```diff
@@ -0,0 +1,124 @@
...
+        'h2',
+        'h3',
+        'h4',
+        'h5',
+        'h6',
+        'header',
+        'i',
+        'img',
+        'kbd',
+        'label',
```
HejdaJakub, line 37: I just noticed that if we have allowed tags `ul/ol`, then we should also allow `li` tag - without this tag list items cannot be added to the un/ordered list.
```suggestion
        'label',
        'li'
```

libs/perun/services/src/lib/html-escape.service.ts
```diff
@@ -0,0 +1,124 @@
...
+        'left',
+        'right',
+        'overflow',
+        'float',
+        'clear',
+        'z-index',
+      ],
+    };
+
+    // Sanitize CSS properties (DOMPurify does not support sanitiizng of style properties)
```
HejdaJakub, line 97:
```suggestion
    // Sanitize CSS properties (DOMPurify does not support sanitizing of style properties)
```

HejdaJakub (changes requested):

HejdaJakub (approved):

HejdaJakub:
We are waiting for the implementation of the backend part just due to consistency (if some critical changes occur that will be important also for this frontend part).

apps/admin-gui/src/app/shared/components/dialogs/edit-application-form-item-dialog/edit-application-form-item-dialog.component.html
```diff
@@ -175,7 +175,11 @@ <h1 mat-dialog-title>
                 <textarea
                   [cdkTextareaAutosize]="true"
                   [(ngModel)]="applicationFormItem.i18n[lang].label"
+                  (input)="checkHtmlContent()"
                   matInput></textarea>
+                <span *ngIf="htmlIsInvalid" class="error">
```
xflord, line 180: Use the 'mat-error' component for the error (see password-form component for inspo)

xflord (changes requested):

xflord (changes requested):
add the error to the notification dialog as well. Also the validator and its function could be extracted to the HTML service to avoid boilerplate code

apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.ts
```diff
@@ -110,6 +138,18 @@ export class AddEditNotificationDialogComponent implements OnInit {
 
   save(): void {
```
xflord, line 139: changes to plain text notifications do not take effect. Probably an issue with the form controls?

apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.html
```diff
@@ -215,7 +220,7 @@ <h1 mat-dialog-title>{{'DIALOGS.NOTIFICATIONS_ADD_EDIT_MAIL.TITLE_EDIT' | transl
           *ngIf="!data.createMailNotification"
           class="ml-2"
           color="accent"
-          [disabled]="loading || !editAuth"
+          [disabled]="loading || !editAuth || inputFormGroup.invalid"
```
xflord, line 223: Please add a tooltip as with the other components (it's very useful when a user on the plaintext tab and the button is disabled because of something on the HTML tab)

apps/admin-gui/src/app/shared/components/dialogs/edit-application-form-item-dialog/edit-application-form-item-dialog.component.html
```diff
@@ -273,18 +278,23 @@ <h1 mat-dialog-title>
...
-            applicationFormItem.perunSourceAttribute === '' &&
-            applicationFormItem.federationAttribute === '' &&
-            (applicationFormItem.disabled === 'ALWAYS' || applicationFormItem.hidden === 'ALWAYS'))"
-          (click)="submit()">
-          {{'DIALOGS.APPLICATION_FORM_EDIT_ITEM.SUBMIT_BUTTON' | translate}}
-        </button>
+        <div
+          [matTooltipDisabled]="!htmlInput.invalid"
+          [matTooltipPosition]="'above'"
+          matTooltip="{{'DIALOGS.NOTIFICATIONS_EDIT_FOOTER.INVALID_TAGS' | translate}}">
```
xflord, line 284: Please edit the tooltip text to fit this component

xflord (changes requested):

apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.ts
mattjoke: Shouldn't the plain text input be considered as text input and therefore we don't need to validate it?

mattjoke: Oh yeah I see the problem. My bad

xflord (approved):
good job!

xflord:
Please fix the commit message and scope ( we do not use core in the new GUI commits)

bodnara (approved):
Great job with this task!

bodnara (changes requested):
There is still some overlap when error message is too long:
image
image
I believe the support from MDC is that error is expected to be just one line so it is not easy to change the margin dynamically. Maybe in the footer case we could put the comment first and than the input to prevent the overlap.


Also if some element spans multiple lines error message is empty. It seems unlikely that this would naturally happen.

apps/admin-gui/src/app/shared/components/dialogs/edit-email-footer-dialog/edit-email-footer-dialog.component.html
```diff
@@ -15,12 +15,21 @@ <h1 mat-dialog-title>{{'DIALOGS.NOTIFICATIONS_EDIT_FOOTER.TITLE' | translate}}</
...
+            <mat-error *ngIf="htmlInput.invalid && format==='html'">
+              {{htmlInput.errors.invalidHtmlContent}}
+            </mat-error>
           </mat-form-field>
+          <div class="mt-2 font-italic text-muted" *ngIf="format === 'plain_text'">
+            {{'DIALOGS.NOTIFICATIONS_EDIT_FOOTER.DESCRIPTION' | translate}}
+          </div>
+          <div class="mt-2 font-italic text-muted" *ngIf="format === 'html'">
+            {{'DIALOGS.NOTIFICATIONS_EDIT_FOOTER.DESCRIPTION_HTML' | translate}}
+          </div>
```
HejdaJakub, line 32: As you added these lines with dynamic text, you should also delete an old div with static text:
```
<div class="mt-2 font-italic text-muted">
    {{'DIALOGS.NOTIFICATIONS_EDIT_FOOTER.DESCRIPTION' | translate}}
</div>
```
image

libs/perun/services/src/lib/html-escape.service.ts
```diff
@@ -0,0 +1,250 @@
...
+    return (control: AbstractControl): ValidationErrors | null => {
+      const { isHtmlInvalid, invalidMessage } = this.checkHtmlContent(control);
+      return isHtmlInvalid ? { invalidHtmlContent: invalidMessage } : null;
+    };
+  }
+
+  private checkHtmlContent(control: AbstractControl): HtmlEscapedMessage {
+    // Remove unsafe tags
+    const { escapedHtml, removedTags, removedAttrs, removedStyleProps } = this.escapeDangerousHtml(
+      control.value as string
```
HejdaJakub, line 235:
```suggestion
      String(control.value)
```

HejdaJakub (changes requested):
I think it looks good otherwise :clap:

HejdaJakub (approved):

PR #1230, opened 2023-01-18
fix(admin): entityless attributes key-value pairs correctly displayed

Kuliak:
* fixed problem when adding new key-value pair for attribute definition, table was not rendered (loaded) properly - missing newly prepared row
* checkboxes fixed
* mat-sort-header keyword removed from table headers (for now)


xflord (commented):
Fix prettier issues, looks good otherwise

bodnara (approved):

xflord (approved):

PR #1231, opened 2023-01-20
feat: allow inputting multi-line string values to string attributes

xflord:
* when changing/adding string attribute value, an edit icon is now present that opens a dialog with a text area that supports multi-line input
* when displaying a multi-line attribute value, show an indicator that it contains multiple lines (whole value is displayed upon clicking or opening the edit dialog)

### DIFF
--- a/libs/perun/components/src/lib/attributes-list/attribute-value/attribute-value-string/attribute-value-string.component.html
+++ b/libs/perun/components/src/lib/attributes-list/attribute-value/attribute-value-string/attribute-value-string.component.html
@@ -1,12 +1,17 @@
 <mat-form-field appearance="{{readonly ? 'none' : 'standard'}}" class="w-100">
-  <input
-    matInput
-    (click)="readonly && value !== undefined && showValue(value, attribute.displayName)"
-    [class.cursor-pointer]="readonly && value !== undefined"
-    [readonly]="readonly"
-    class="overflow-ellipsis"
-    [(ngModel)]="attribute.value"
-    type="text"
-    attr.data-cy="{{attribute.displayName | multiWordDataCy}}-value"
-    (keydown)="_sendEventToParent()" />
+  <div class="inline">
+    <input
+      matInput
+      matTooltip="{{'SHARED_LIB.PERUN.COMPONENTS.ATTRIBUTE_VALUE_STRING.TOOLTIP' | translate}}"
+      [matTooltipDisabled]="readonly || !multiline"
+      (click)="(readonly || multiline) && value !== undefined && showValue(attribute.value, attribute.displayName)"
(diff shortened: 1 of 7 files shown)


sarkapalkovicova (approved):

libs/perun/components/src/lib/attributes-list/attribute-value/attribute-value-string/attribute-value-string.component.html
```diff
@@ -1,12 +1,17 @@
...
-    [readonly]="readonly"
-    class="overflow-ellipsis"
-    [(ngModel)]="attribute.value"
-    type="text"
-    attr.data-cy="{{attribute.displayName | multiWordDataCy}}-value"
-    (keydown)="_sendEventToParent()" />
+  <div class="inline">
+    <input
+      matInput
+      matTooltip="Please use the edit button to edit multi-line string attributes"
```
bodnara, line 5: Missing translation.

libs/perun/components/src/lib/attributes-list/attribute-value/attribute-value-string/attribute-value-string.component.ts
```diff
@@ -15,19 +18,34 @@ export class AttributeValueStringComponent implements OnInit {
...
 
+  checkMultilineValue(): void {
+    this.multiline = false;
+    if (this.attribute.value && (this.attribute.value as string).includes('\n')) {
+      this.multiline = true;
+      const values = (this.attribute.value as string).split('\n');
+      this.value =
+        `${values[0]} ... (${values.length - 1} more line` + (values.length > 2 ? 's)' : ')');
+    }
+    this.editable = !this.readonly && !this.multiline;
```
bodnara, line 46: Seems like `editable` is not used for anything and can be removed.

libs/perun/components/src/lib/attributes-list/attribute-value/attribute-value-string/attribute-value-string.component.ts
```diff
@@ -15,19 +18,34 @@ export class AttributeValueStringComponent implements OnInit {
...
     this.sendEventToParent.emit();
   }
 
+  checkMultilineValue(): void {
+    this.multiline = false;
+    if (this.attribute.value && (this.attribute.value as string).includes('\n')) {
+      this.multiline = true;
+      const values = (this.attribute.value as string).split('\n');
+      this.value =
+        `${values[0]} ... (${values.length - 1} more line` + (values.length > 2 ? 's)' : ')');
```
bodnara, line 44: More lines indicator could also be localized.

bodnara (changes requested):

bodnara (approved):

balcirakpeter:
We are waiting for feedback from other people.

PR #1235, opened 2023-01-24
feat(apps): remove redundant comments and TODO(v15) comments

Kuliak:
* as for TODO(V15) comments, only includes were automatically added


mattjoke (approved):
:+1:

bodnara (changes requested):
Missing user-profile styles.

PR #1239, opened 2023-01-26
feat(app): Added script for color migration

mattjoke:
* color-migration.js will use _styles.css in apps and generate static files with colors instead of var(--variable)
* added color-migration.js as a postinstall script
* replaces all style.css with _style.css (which are used as a template by the script)


color-migration.js
```diff
@@ -0,0 +1,132 @@
...
+    try {
+      const file = await fs.readFile(`./apps/${dirent.name}/src/assets/config/defaultConfig.json`, 'utf-8');
+      const obj = JSON.parse(file);
+      config.colors = obj["theme"];
+      // Update configs with instanceConfig.json
+      const file2 = await fs.readFile(`./apps/${dirent.name}/src/assets/config/instanceConfig.json`, 'utf-8');
+      const obj2 = JSON.parse(file2);
+      // Update missing colors
+      for (let color in config.colors) {
+        if (config.colors[color] === null) {
```
xflord, line 44: feel like this doesn't work as intended

apps/admin-gui/src/_styles.scss
```diff
@@ -439,8 +410,7 @@ $perun-theme: mat.define-light-theme(
     200: rgba(var(--admin-theme-primary-200), 1),
     300: rgba(var(--admin-theme-primary-300), 1),
     400: rgba(var(--admin-theme-primary-400), 1),
-    //FIXME(theme-error): we want ot load color from the config
-    // 500: rgba(var(--admin-theme-primary-500), 1),
+    500: rgba(var(--admin-theme-primary-500), 1),
     500: rgba(#c62828, 1),
```
xflord, line 414: forgot to delete this one i suppose

xflord (commented):
Otherwise very nice, works well

color-migration.js
mattjoke: Yes, I suppose it does not work as is should

mattjoke: After tinkering with the script, I found out that the colours from instanceConfig.json are not that necessary to be used by this script. The AppConfigService will update the instance with the colours from the instanceConfig.json. Nevertheless, I updated the script to work properly

HejdaJakub (commented):
Why do you have so many deletions in your package-lock file? Do you have the same amount of deletions if you try to run `npm install --legacy-peer-deps`?

mattjoke:
> Why do you have so many deletions in your package-lock file? Do you have the same amount of deletions if you try to run `npm install --legacy-peer-deps`?

Yes, I ran `npm install --legacy-peer-deps`, and it seems that the number of changes went down to ~250.

xflord (commented):
Last few notes:

Screenshot from 2023-02-06 10-07-15
You can disable these warnings.

Also, just a thought: maybe add a comment at the top of the styles files that any changes should be made to the _styles template and not the styles file itself? This could confuse new developers that are used to making changes to styles.scss

xflord:
also edit the CI to work with these changes

apps/admin-gui/src/_styles.scss
```diff
@@ -40,21 +45,22 @@ $perun-accent: mat.define-palette(mat.$green-palette, 600);
...
     ),
+    typography: mat.define-typography-config(),
+    density: 0,
   )
 );
 
 // Include theme styles for core and each component used in your app.
 // Alternatively, you can import and @include the theme mixins for each component
 // that you are using.
+$mat-theme-ignore-duplication-warnings: true;
```
xflord, line 63:
```suggestion
mat.$theme-ignore-duplication-warnings: true;
```
it works when I change it to this.

mattjoke: Fixed.

xflord (approved):

HejdaJakub (approved):

PR #1248, opened 2023-01-31
feat(admin): user roles are split by source

Kuliak:
* user-detail page splits his roles by source (directly assigned or obtained as a member of authorized group)
* tabs added to change what source we want to see
* authorized-group-based roles cannot be managed, only viewed

### DIFF
--- a/apps/admin-gui/src/app/shared/components/roles-list/roles-page.component.html
+++ b/apps/admin-gui/src/app/shared/components/roles-list/roles-page.component.html
@@ -1,9 +1,8 @@
-<h1 class="page-subtitle d-flex">{{'ROLES.TITLE'| translate}}</h1>
 <button
-  *ngIf="assignableRules.length !== 0"
+  *ngIf="assignableRules.length !== 0 && editable"
   mat-flat-button
   color="accent"
-  class="mb-3 mr-2"
+  class="mb-3 mr-2 mt-2"
   (click)="addRole()">
   {{'ROLES.ADD'| translate}}
 </button>
@@ -22,7 +21,7 @@
     </mat-expansion-panel-header>
     <ng-template matExpansionPanelContent>
       <span
-        *ngIf="role.roleName !== 'SELF' && role.roleName !== 'MEMBERSHIP' && role.roleName !== 'SPONSORSHIP'"
+        *ngIf="editable && role.roleName !== 'SELF' && role.roleName !== 'MEMBERSHIP' && role.roleName !== 'SPONSORSHIP'"
         matTooltip="{{'ROLES.REMOVE_DISABLED_TOOLTIP'| translate}}"
         matTooltipPosition="left"
         [matTooltipDisabled]="(selection.selected.length === 0 && selectedFacilities.selected.length === 0) || !disableRemove">
@@ -43,9 +42,11 @@
               {{('ROLES.' + role.roleName + '_VOS_' + entityType) | translate}}
             </div>
             <perun-web-apps-vos-list
-              [displayedColumns]="role.roleName === 'MEMBERSHIP' ? ['id', 'name', 'shortName'] : ['checkbox', 'id', 'name', 'shortName']"
+              [displayedColumns]="role.roleName === 'MEMBERSHIP' ? ['id', 'name', 'shortName'] : (editable ? ['checkbox', 'id', 'name', 'shortName'] : ['id', 'name', 'authzGroup', 'shortName'])"
               [vos]="vos | async"
-              [selection]="selection">
+              [selection]="selection"
+              [authzVoNames]="voNames"
+              [voWithAuthzGroupPairs]="_complementaryObjectsWithAuthzGroups?.get(role.roleName)?.get('vo')">
             </perun-web-apps-vos-list>
           </div>
           <div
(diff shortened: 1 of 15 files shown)


Johaney-s (commented):
It seems myProfile -> roles view is not working properly yet:
image

and just a thought, but it might be too much work for such little thing, showing which authoritative group is giving us the role could be useful.

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -2511,7 +2511,9 @@
       "SHOW_RECENTLY_VIEWED": "Show Recently viewed",
       "SHOW_ROLES": "Show roles",
       "MAIL_CHANGE_SUCCESS": "Preferred mail has been changed",
-      "ANONYMIZED": "User anonymized"
+      "ANONYMIZED": "User anonymized",
+      "DIRECT_ROLES": "Direct",
+      "AUTHORIZED_GROUP_BASED_ROLES": "Authorized-group-based"
```
xflord, line 2516: Would prefer it without the dashes

xflord (approved):
Remove the now unused imports, otherwise looks good, good job

libs/perun/services/src/lib/role.service.ts
```diff
@@ -21,4 +22,28 @@ export class RoleService {
     });
     return preparedRoles;
   }
+
+  prepareComplementaryObjects(
```
xflord, line 26: please add javadocs

apps/admin-gui/src/app/shared/components/roles-list/roles-page.component.html
```diff
@@ -43,9 +43,11 @@ <h1 class="page-subtitle d-flex">{{'ROLES.TITLE'| translate}}</h1>
               {{('ROLES.' + role.roleName + '_VOS_' + entityType) | translate}}
             </div>
             <perun-web-apps-vos-list
-              [displayedColumns]="role.roleName === 'MEMBERSHIP' ? ['id', 'name', 'shortName'] : ['checkbox', 'id', 'name', 'shortName']"
+              [displayedColumns]="role.roleName === 'MEMBERSHIP' || !editable? ['id', 'name', 'authzGroup', 'shortName'] : ['checkbox', 'id', 'name', 'authzGroup', 'shortName']"
               [vos]="vos | async"
-              [selection]="selection">
+              [selection]="selection"
+              [authzVoNames]="voNames"
+              [voWithAuthzGroupPairs]="_complementaryObjectsWithAuthzGroups.get(role.roleName).get('vo')">
```
xflord, line 50: This causes exceptions in the user based tab.

xflord (changes requested):

apps/admin-gui/src/app/shared/components/roles-list/roles-page.component.html
```diff
@@ -27,6 +26,7 @@ <h1 class="page-subtitle d-flex">{{'ROLES.TITLE'| translate}}</h1>
         matTooltipPosition="left"
         [matTooltipDisabled]="(selection.selected.length === 0 && selectedFacilities.selected.length === 0) || !disableRemove">
         <button
+          *ngIf="editable"
```
bodnara, line 29: Nitpick but the proper target should be the parent of the button (span tag). Right now the span is present so the tooltip could be triggered.

apps/admin-gui/src/app/users/pages/user-detail-page/user-settings/user-roles/user-roles.component.ts
```diff
@@ -43,22 +43,38 @@ export class UserRolesComponent implements OnInit {
   }
 
   getData(): void {
+    let loadingTypes = 3;
     this.outerLoading = true;
+
```
bodnara, line 48: While this works a better solution (without rxjs magic) would be to nest these calls. Also stop the loading if some of the requests fail.

libs/perun/components/src/lib/facilities-list/facilities-list.component.html
```diff
@@ -51,6 +51,16 @@
...
       </ng-container>
+      <ng-container matColumnDef="authzGroup">
+        <th mat-header-cell *matHeaderCellDef mat-sort-header>
+          {{'SHARED.COMPONENTS.FACILITIES_LIST.AUTHZ_GROUP' | translate}}
+        </th>
+        <td mat-cell *matCellDef="let facility">
+          {{this._facilityWithAuthzGroupPairs.has(facility.facility.id)
+              ? _authzVoNames.get(this._facilityWithAuthzGroupPairs.get(facility.facility.id)[0].voId) + ":" + this._facilityWithAuthzGroupPairs.get(facility.facility.id)[0].name
+              : "NO" }}
+        </td>
```
bodnara, line 62: I prefer the empty string for the else clause.

libs/perun/components/src/lib/facilities-list/facilities-list.component.ts
```diff
@@ -41,11 +41,21 @@ export class FacilitiesListComponent implements OnChanges {
...
 
   constructor(private authResolver: GuiAuthResolver) {}
 
+  @Input() set facilityWithAuthzGroupPairs(pairs: Map<number, Group[]>) {
+    this._facilityWithAuthzGroupPairs = pairs;
+  }
+
+  @Input() set authzVoNames(authzVoNames: Map<number, string>) {
+    this._authzVoNames = authzVoNames;
+  }
```
bodnara, line 57: It is not technically wrong to use input setter like this, it is more useful when there is a peace of code that needs to run every time the input changes. However to stay consistent with the rest of the codebase could you please make these a simple `@Input`. This goes for other list components.

bodnara (changes requested):

apps/admin-gui/src/app/users/pages/user-detail-page/user-settings/user-roles/user-roles.component.ts
```diff
@@ -44,21 +44,51 @@ export class UserRolesComponent implements OnInit {
...
+                    // prepare map
+                    this.rolesComplementaryObjectsWithAuthzGroups =
+                      this.roleService.prepareComplementaryObjects(
+                        Object.keys(obtainedComplementaryObjects),
+                        obtainedComplementaryObjects
+                      );
+                    this.outerLoading = false;
+                  },
+                  () => (this.outerLoading = false)
+                );
```
xflord, line 86:
```suggestion
                .subscribe({
                  next: (obtainedComplementaryObjects) => {
                    // prepare map
                    this.rolesComplementaryObjectsWithAuthzGroups =
                      this.roleService.prepareComplementaryObjects(
                        Object.keys(obtainedComplementaryObjects),
                        obtainedComplementaryObjects
                      );
                    this.outerLoading = false;
                  },
                  error: () => (this.outerLoading = false)
                });
```
use the non-deprecated syntax with all the subscribe methods please (like this I believe)

xflord (commented):
fix the scope to feat(admin) since the roles page is only in the admin gui

xflord (approved):

bodnara (commented):
There is some uninitialized value in the group-list on user based tab.
image

bodnara (approved):

PR #1249, opened 2023-02-01
feat: allow inputting multi-line string values to string attributes

xflord:


libs/perun/components/src/lib/attributes-list/attribute-value/attribute-value-string/attribute-value-string.component.scss
```diff
@@ -2,6 +2,19 @@
...
 
+.hint-color {
+  color: darkorange;
+}
+
 .overflow-ellipsis {
   text-overflow: ellipsis;
 }
+//hide hint when field not in focus
+.mat-form-field:not(.mat-focused) .mat-hint {
```
HejdaJakub, line 13: I will have to change this component (fix the UI on global level) in my PR on migration to MDC, but at least this comment for this local scss file will be useful.
NOTE: I used `mat-hint` without the dot (`.`) on purpose, but if we want to use class definition, then we can use this one in Angular 15: `.mat-mdc-form-field-hint`.
```suggestion
// FIXME: After the migration to MDC and Angular 15 have to be replaced the line below
// .mat-mdc-form-field:not(.mat-focused) mat-hint {
.mat-form-field:not(.mat-focused) .mat-hint {
```

HejdaJakub (approved):

HejdaJakub (changes requested):
1) We have had ellipsis for long text/smaller screens since there was input, but for the textarea ellipsis doesn't work and it doesn't look very well (especially for single line non-editable attributes).
image

2) I would also fix CSS for vertical mover - it is probably caused by our custom global CSS for attributes.
image

It is weird when the mover is exactly on the bottom line and I believe it should look like e.g. in notifications:
image

xflord:
TODO: check whether multine attribute values get correctly propagated to LDAP

bodnara (approved):

mattjoke (approved):

HejdaJakub:
@xflord tested the propagation to LDAP and it works

PR #1254, opened 2023-02-07
fix(admin): fixed a bug that caused an incorrect member to initially …

xflord:
…load on the user accounts page

* the bug was caused by vos arrays being sorted differently in the user-accounts component and the vo select component

### DIFF
--- a/apps/admin-gui/src/app/users/pages/user-detail-page/user-accounts/user-accounts.component.ts
+++ b/apps/admin-gui/src/app/users/pages/user-detail-page/user-accounts/user-accounts.component.ts
@@ -9,6 +9,7 @@ import {
 } from '@perun-web-apps/perun/openapi';
 import { ActivatedRoute } from '@angular/router';
 import { Urns } from '@perun-web-apps/perun/urns';
+import { compareFnName } from '@perun-web-apps/perun/utils';
 
 @Component({
   selector: 'app-perun-web-apps-user-accounts',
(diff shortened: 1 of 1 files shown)


apps/admin-gui/src/app/users/pages/user-detail-page/user-accounts/user-accounts.component.html
```diff
@@ -5,6 +5,7 @@ <h1 class="page-title">{{'USER_DETAIL.ACCOUNTS.TITLE' | translate}}</h1>
     *ngIf="!(initLoading || vos.length === 0)"
     class="w-50"
     [vos]="vos"
+    [vo]="selectedVo"
```
HejdaJakub, line 8: I think this is not necessary since you sort it in ts file correctly now.
```suggestion
```

apps/admin-gui/src/app/users/pages/user-detail-page/user-accounts/user-accounts.component.ts
```diff
@@ -39,7 +40,7 @@ export class UserAccountsComponent implements OnInit {
         (vos) => {
           this.vos = vos;
           if (this.vos.length) {
-            this.loadMember(this.vos[0]);
+            this.loadMember(this.vos.sort(compareFnName)[0]);
```
HejdaJakub, line 43: Could you please also fix the subscribe methods (three occurrences) to a not deprecated version?

HejdaJakub (approved):

PR #1255, opened 2023-02-07
fix(admin): replace mat-selects by our custom search selects

xkostka2:
* fixed AddMemberToResource dialog where we were still using plain mat-selects instead of our custom search selects

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/add-member-to-resource-dialog/add-member-to-resource-dialog.component.html
+++ b/apps/admin-gui/src/app/shared/components/dialogs/add-member-to-resource-dialog/add-member-to-resource-dialog.component.html
@@ -9,37 +9,27 @@
         >
         <mat-spinner *ngIf="loading" class="mr-auto ml-auto"></mat-spinner>
         <div *ngIf="!loading">
-          <div>
-            <mat-form-field class="input-style">
-              <input
-                [formControl]="facilityCtrl"
-                [matAutocomplete]="autoFacility"
-                aria-label="Number"
-                matInput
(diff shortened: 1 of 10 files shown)


apps/admin-gui/src/app/shared/components/dialogs/add-member-to-resource-dialog/add-member-to-resource-dialog.component.html
```diff
@@ -9,37 +9,24 @@ <h1 mat-dialog-title>{{'DIALOGS.ADD_MEMBER_TO_RESOURCE.TITLE' | translate}}</h1>
...
-                  [value]="resource">
-                  {{resource.name}}
-                </mat-option>
-              </mat-select>
-            </mat-form-field>
+          <div class="input-style">
+            <perun-web-apps-resource-search-select
+              (resourceSelected)="setResource($event);
+                         stepper.selected.completed = true"
+              *ngIf="!processing"
```
xflord, line 24:
```suggestion
```
think this is unnessacary (or is this meant to reset the search value on select? if so, could this be done without the search-selects disappearing (this seems quite ugly to me)? (I'm assuming this would require changes to the search-select components)

HejdaJakub (changes requested):
If I select a facility, some resource is automatically selected (at least it looks like the resource is selected), but the resource is not really selected, so the `Continue` button is NOT disabled, but if you try to click on this button, it does not move you to the next step.

xflord (approved):

HejdaJakub (approved):

PR #1257, opened 2023-02-07
refactor(admin): use date-range component from Angular material

xkostka2:
* picking date range using 2 date pickers has been replaced by single date-range component from Angular material

### DIFF
--- a/libs/perun/components/src/lib/date-range/date-range.component.ts
+++ b/libs/perun/components/src/lib/date-range/date-range.component.ts
@@ -1,5 +1,5 @@
-import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
-import { UntypedFormControl } from '@angular/forms';
+import { Component, Input, OnInit } from '@angular/core';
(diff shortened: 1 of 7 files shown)


libs/perun/components/src/lib/date-range/date-range.component.ts
```diff
@@ -11,20 +11,14 @@ export class DateRangeComponent implements OnInit {
   @Input() endDate: UntypedFormControl;
```
bodnara, line 11: Please, change the Form Controls to the typed variant.

bodnara (changes requested):
The date range could also be used in the `YearRangeComponent` or maybe these two could be merged to one component.

xflord (commented):
Screenshot from 2023-02-09 09-15-53
When choosing only the start date and clicking away, the table refreshes and shows this, which to me looks like results from the start date forward should be displayed (but that's not the case, nothing is displayed)

xflord (approved):

bodnara (approved):

PR #1260, opened 2023-02-11
feat(admin): consents for assigned members

sarkapalkovicova:
Show consents on the page with members assigned to the resource.


xkostka2 (approved):

libs/perun/components/src/lib/members-list/members-list.component.ts
```diff
@@ -270,4 +275,17 @@ export class MembersListComponent implements OnChanges, AfterViewInit {
 
     this.dialog.open(MemberTreeViewDialogComponent, config);
   }
+
+  ngOnInit(): void {
+    if (this.includeConsents && this.facilityId) {
```
bodnara, line 280: I would argue that this should not be the responsibility of the table but rather the caller. In this case the `ResourceAssignedMembersComponent`.

libs/perun/components/src/lib/members-list/members-list.component.ts
```diff
@@ -270,4 +278,27 @@ export class MembersListComponent implements OnChanges, AfterViewInit {
...
+    return members;
+  }
+
+  transformConsentForMember(member: RichMember, consents: Consent[]): string {
+    if (!consents || consents.length === 0) return;
+    const membersConsents = consents.filter((consent) => consent.userId === member.userId);
+    // return latest consent (one with the highest id)
+    if (membersConsents.length === 0) return;
+    return membersConsents.reduce((a, b) => (a.id > b.id ? a : b)).status;
+  }
```
bodnara, line 303: The solution is nice but I have couple of suggestions. It would be nice to allow filtering/sorting by consents. However that would require to make the transformation every time the filter or sorting changes (also apply the `mat-sort-header` to the column) which is redundant. I would also still prefer if this was not handled by the table itself.

I suggest to pass the `MemberWithConsentStatus` from the parent since it has all the information.

It seems the `transformConsentForMember` is just a duplication of `ConsentStatusIconPipe`. You can use the pipe inside of `.ts` files as well. For this provide the pipe inside of  `@Component` decorator. In this case `providers: [ConsentStatusIconPipe]`. Than you can create an instance of the pipe like this `constructor(private somePipe: SomePipe)` and use it's transform method where needed `this.somePipe.transform(...)`.

xflord: I also agree with the need for filtering. The most optimal solution would IMO be to use the dynamic members list and do the filtering backend side, however, that would require changes to the paginated members get method (which I'm not quite sure is necessary, depending on how large the list of assigned members can get, up to debate maybe?).
For now, I would be satisfied with the filtering taking place in resource-assigned-members component, which would also allow us to pass MemberWithConsentStatus to the list component (as Adam suggested).

apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-assigned-members/resource-assigned-members.component.html
```diff
@@ -1,16 +1,33 @@
 <h1 class="page-subtitle">{{'RESOURCE_DETAIL.ASSIGNED_MEMBERS.TITLE' | translate}}</h1>
 <perun-web-apps-refresh-button (refresh)="refreshTable()"> </perun-web-apps-refresh-button>
+
+<mat-form-field class="mr-2 fix-select" *ngIf="includeConsents">
+  <mat-label>{{'RESOURCE_DETAIL.ASSIGNED_MEMBERS.FILTER_CONSENT_STATUS' | translate}}</mat-label>
+  <mat-select (closed)="changeConsentStatuses()" [formControl]="consentStatuses" multiple>
+    <mat-select-trigger>
+      {{displaySelectedConsentStatuses()}}
```
bodnara, line 8: For any calculations inside of a template `{{ ... }}` prefer to use pipe instead of a function call. The pipe is evaluated when needed (on relevant changes) while the function is executed on any change.

apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-assigned-members/resource-assigned-members.component.html
```diff
@@ -1,16 +1,33 @@
...
+    </mat-select-trigger>
+    <mat-option
+      class="resource-theme"
+      *ngFor="let consentStatus of consentStatusList"
+      [value]="consentStatus"
+      >{{consentStatus | memberStatus}}</mat-option
+    >
```
bodnara, line 15: Prettier struggles to format this nicely.
```suggestion
      [value]="consentStatus">
      {{consentStatus | memberStatus}}
    </mat-option>
```

apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-assigned-members/resource-assigned-members.component.ts
```diff
@@ -1,29 +1,56 @@
...
   filterValue = '';
   tableId = TABLE_RESOURCE_MEMBERS;
 
   resource: Resource;
-  members: RichMember[];
+  members: MemberWithConsentStatus[];
+  columns = ['id', 'fullName'];
+  includeConsents = false;
+  consents: Consent[] = [];
+  consentStatuses = new UntypedFormControl();
```
bodnara, line 38: Please use the typed version for Form controls.

apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-assigned-members/resource-assigned-members.component.ts
```diff
@@ -35,21 +62,74 @@ export class ResourceAssignedMembersComponent implements OnInit {
...
     this.filterValue = filterValue;
   }
+
+  getConsentsForMembers(members: RichMember[]): MemberWithConsentStatus[] {
+    const result: MemberWithConsentStatus[] = [];
+    members.forEach((member) => {
+      const mwc: MemberWithConsentStatus = member;
+      mwc.consent = this.translate.instant(
+        'CONSENTS.STATUS_' + this.consentStatusPipe.transform(member.userId, this.consents)
+      ) as string;
```
bodnara, line 106: I would suggest keeping the raw consent status inside of the object. Also for translation use the `PerunTranslateService`.

bodnara (changes requested):
Nice work with the filtering. There is one bug when opening the filter not selecting anything and closing it. Other than that its just small adjustments.

libs/perun/components/src/lib/members-list/members-list.component.html
```diff
@@ -134,6 +134,15 @@
         <th *matHeaderCellDef mat-header-cell>{{'MEMBERS_LIST.LOGINS' | translate}}</th>
         <td *matCellDef="let member" mat-cell>{{member | memberLogins}}</td>
       </ng-container>
+      <ng-container matColumnDef="consentStatus">
+        <th *matHeaderCellDef mat-header-cell mat-sort-header>
+          {{'CONSENTS.STATUS' | translate}}
+        </th>
+        <td mat-cell *matCellDef="let member">
+          <perun-web-apps-consent-status
+            [consentStatus]="member.userId | consentStatusIcon: consents"></perun-web-apps-consent-status>
```
xflord, line 143: There's no need to call the pipe here, since we already called it in the parent and assigned the consent to MemberWithConsent.

apps/admin-gui/src/app/facilities/pages/resource-detail-page/resource-assigned-members/resource-assigned-members.component.ts
xflord: With this change, the data is exported without the consent status translated (which is an inconsistency compared to users dynamic list where it's translated, should we not translate it there as well?)

bodnara: If we want the consents to be translated when exported (seems reasonable) we could can do that it the `getExportDataForColumn`.

sarkapalkovicova: Good point. I overlooked it. However, there is a problem with using injected translate service in the `getExportDataForColumn` static method. This method is used only once in `getExportDataForColumnFun` in the same component. Is it necessary for this method to be static?

xflord: Yep, that's what I was dealing with when implementing the consent column for the assigned users page. I believe there was a reason, cannot remember though. @bodnara ?

libs/perun/components/src/lib/members-list/members-list.component.ts
```diff
@@ -63,9 +69,10 @@ export class MembersListComponent implements OnChanges, AfterViewInit {
   @Input() tableId: string;
   @Input() disableRouting = false;
   @Input() filter = '';
+  @Input() consents: Consent[] = [];
```
bodnara, line 72: After the changes we do not need the consents anymore. 
Other than that the filtering and exporting is working properly so well done.

xflord (approved):
good job

bodnara (approved):
Please rebase and merge after https://github.com/CESNET/perun/pull/3883.

PR #1261, opened 2023-02-14
refactor(admin): move application state selection into separate component

xkostka2:
* logic for filtering applications to VOs and groups has been moved to separate component

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "perun-web-apps",
       "version": "0.0.0-development",
       "hasInstallScript": true,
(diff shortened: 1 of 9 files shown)


package-lock.json
```diff
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "perun-web-apps",
```
xflord, line 8: not sure what this is doing here

xkostka2: it appeared after I used "npm install". Since this file is generated, it shouldn't be an issue.

xflord (approved):

bodnara (approved):

PR #1264, opened 2023-02-20
feat: improve loading of dialogs and tables

HejdaJakub:
* Two new loader components were created and used across all apps (LoadingTableComponent and LoadingDialogComponent).
* The use case when some dialog includes also table was checked and it should work correctly (only one spinner should be visible).
* Another possible candidates for new loader components are "entity-detail-page" and "overview" components, but it is usually not possible to display some data (data are not fetched yet) during the loading behind the spinner for these components. Since it is not about UX but only about "consistency" or possible future easier changes, I did not implemented it in this PR.

### DIFF
--- a/apps/user-profile/src/app/pages/groups-page/groups-page.component.ts
+++ b/apps/user-profile/src/app/pages/groups-page/groups-page.component.ts
@@ -23,6 +23,7 @@ import { Membership } from '../../components/membership-list/membership-list.com
 })
 export class GroupsPageComponent implements OnInit {
   loading = false;
+  initialLoading = false;
   userId: number;
   vos: Vo[] = [];
   myControl = new UntypedFormControl();
@@ -33,6 +34,8 @@ export class GroupsPageComponent implements OnInit {
 
   userMemberships: Membership[] = [];
   adminMemberships: Membership[] = [];
+  userMembershipsTemp: Membership[] = [];
+  adminMembershipsTemp: Membership[] = [];
 
   constructor(
     private usersService: UsersManagerService,
@@ -45,6 +48,7 @@ export class GroupsPageComponent implements OnInit {
 
   ngOnInit(): void {
     this.loading = true;
+    this.initialLoading = true;
     this.userId = this.store.getPerunPrincipal().userId;
 
     this.usersService.getVosWhereUserIsMember(this.userId).subscribe((vos) => {
@@ -61,8 +65,8 @@ export class GroupsPageComponent implements OnInit {
     this.loading = true;
     let i = 0;
     let j = 0;
-    this.userMemberships = [];
-    this.adminMemberships = [];
+    this.userMembershipsTemp = [];
+    this.adminMembershipsTemp = [];
     const allMemberIds = this.store.getPerunPrincipal().roles['SELF']['Member'];
     if (!allMemberIds.length) {
       this.loading = false;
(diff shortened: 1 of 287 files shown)


xkostka2 (changes requested):
I found a minor consistency issue in the user profile app. On the Organisations page, there is a loader for each table separately, but on the Linked accounts page, there is just one loader for the whole page, even tho this page also contains multiple tables.

.eslintcache
```diff
@@ -0,0 +1 @@
...":0},{"filePath":"2777","messages":"2778","suppressedMessages":"2779","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"2780","messages":"2781","suppressedMessages":"2782","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"2783","messages":"2784","suppressedMessages":"2785","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"2786","messages":"2787","suppressedMessages":"2788","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"2789","messages":"2790","suppressedMessages":"2791","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"2792","messages":"2793","suppressedMessages":"2794","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"2795","messages":"2796","suppressedMessages":"2797","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},{"filePath":"2798","messages":"2799","suppresse
```
xflord, line 1: remove this please

xflord:
Screenshot from 2023-02-21 10-37-57
IMO the 'no entities found' warning should not be visible during loading

xflord:
Screenshot from 2023-02-21 11-15-30
Screenshot from 2023-02-21 11-12-42
I would probably also limit the 'greyed out area' of the dialogs to dialog content only, the greyed out buttons and title looks weird to me (and is inconsistent with how dialogs with tables act)

xflord:
Screenshot from 2023-02-21 11-47-04
The spinners in publication-detail component still replace the content

xflord:
That's all I found for now

HejdaJakub:
Discussion on Daily:
- for this moment we want to let a visible warning (empty table) during the initial loading
- the special use case for publication detail we want to skip for this moment
- we want to have a different spinner for a table in the dialog and for the whole dialog (already implemented), but the screenshot, where the dialog looks disabled just partly during the loading (without gray margins), was taken in firefox. This is due to firefox's default setting (for `:has()` CSS pseudo-class) which is criticized by the community and hopefully will be changed soon. So, we can use some workaround, which probably can be handled only with some javascript to add class dynamically, or we can hope that this default setting will be changed soon for firefox.
image
image

HejdaJakub:
Fixed pages `identities-page` and `groups-page` and deleted file `.eslintcache`.

apps/user-profile/src/app/pages/groups-page/groups-page.component.ts
```diff
@@ -11,7 +11,7 @@ import {
 import { StoreService } from '@perun-web-apps/perun/services';
 import { UntypedFormControl } from '@angular/forms';
 import { MatAutocompleteSelectedEvent } from '@angular/material/autocomplete';
-import { map, startWith } from 'rxjs/operators';
+import { finalize, map, startWith, tap } from 'rxjs/operators';
```
xflord, line 14: Unused imports

xflord (approved):
Fix the imports please, otherwise look good to me

mattjoke (approved):
Apart from changes mentioned above, I haven't found anything.

PR #1265, opened 2023-02-21
fix(profile): login page is correctly displayed after session expiration

xkostka2:
* login page is correctly displayed after session expiration


bodnara (commented):
After I logged in to profile it shows a bunch of errors regarding tokens. It was a "fresh" log in from private window.
image

xkostka2:
> After I logged in to profile it shows a bunch of errors regarding tokens. It was a "fresh" log in from private window. image

I did not get any of these errors when I tried to replicate what you did.

bodnara (approved):
Seems like the issue was on my end.

PR #1266, opened 2023-02-21
feat(admin): sponsored member create dialog input fields fix

Kuliak:
* login and password input fields are disabled by default (namespace must be selected first)
* internationalized message added


HejdaJakub (commented):
In the task's definition is mentioned also the use case when we want to disable inputs also in the case when the instance supports creating an account without login.

PR #1268, opened 2023-02-23
fix: UX improvements regarding transition to Angular 15

xkostka2:


apps/admin-gui/src/app/shared/components/resources-tags-list/resources-tags-list.component.html
```diff
@@ -55,8 +55,10 @@
           </div>
           <div *ngIf="isChanging.isSelected(tag)">
             <mat-form-field>
-              <mat-label>{{'VO_DETAIL.RESOURCES.TAGS.CHANGE_TAG_NAME' | translate}}</mat-label>
-              <input [(ngModel)]="tag.tagName" matInput />
+              <input
+                [(ngModel)]="tag.tagName"
+                matInput
+                placeholder="{{'VO_DETAIL.RESOURCES.TAGS.CHANGE_TAG_NAME' | translate}}" />
```
bodnara, line 61: I am not sure but to my understanding the `<mat-label>` should be used instead of placeholder value ?

xkostka2: if I used mat-label this would be the result
image

HejdaJakub: It's true, according to mdc-migration guide: 
`The removed "legacy" appearance promoted input placeholders to the floating label if the label was not specified. All newer appearance settings require explicitly specifying a <mat-label> if one was not provided before. This change addresses an accessibility best practice of not using labels and placeholders interchangeably.`
On the other hand, our accessibility probably is not very good across all GUI, so if this is the only place where we use floating label in a table, can we maybe skip it?

But at least I would change `<mat-form-field>` to `<mat-form-field subscriptSizing="dynamic">` for better UX.

HejdaJakub (changes requested):

HejdaJakub:
Looks good now. I move all these changes to the main PR (https://github.com/CESNET/perun-web-apps/pull/1275) that should be merged further into the master branch.

PR #1269, opened 2023-02-23
fix: minor visual bug fixes after angular 15 migration

xflord:
* added user style to defaultConfig.json in apps it was missing, since it now won't load without it
* fixed mailing lists checkbox misalignment
* added the visibility icon to the form field in settings-samba-password.component
* fixed a bug that made password form jump when spinner was visible (global spinner margin was removed, it was added just for a slight visual change, I believe we can do without it)

### DIFF
--- a/apps/password-reset/src/_styles.scss
+++ b/apps/password-reset/src/_styles.scss
@@ -78,9 +78,16 @@ $password-reset-theme: mat.define-light-theme(
   $user-primary: mat.define-palette($user-dynamic-colors, 500);
   $user-accent: mat.define-palette(mat.$green-palette, 600);
 
-  $user-theme: mat.define-light-theme($user-primary, $user-accent);
+  $user-theme: mat.define-light-theme(
+    (
+      color: (
+        primary: $user-primary,
+        accent: $user-accent,
+      ),
+    )
+  );
 
-  @include mat.all-component-themes($user-theme);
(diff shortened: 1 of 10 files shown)


bodnara (approved):

HejdaJakub (commented):
I also found and realized that `_styles.scss` for password-reset should be changed for consistency with other sccs global files and also to avoid warnings:
`$user-theme: mat.define-light-theme($user-primary, $user-accent);`
should be replaced by this:
```
$user-theme: mat.define-light-theme(
    (
      color: (
        primary: $user-primary,
        accent: $user-accent,
      ),
    )
  );
```
and mat.all-component-**themes**($user-theme); should be replaced by mat.all-component-**colors**($user-theme);

apps/password-reset/src/_styles.scss
```diff
@@ -78,7 +78,14 @@ $password-reset-theme: mat.define-light-theme(
...
+    (
+      color: (
+        primary: $user-primary,
+        accent: $user-accent,
+      ),
+    )
+  );
 
   @include mat.all-component-themes($user-theme);
 }
```
HejdaJakub, line 91: As I said before, also change **themes** to **colors**, please.
```suggestion
  );

  @include mat.all-component-colors($user-theme);
```
image

HejdaJakub (approved):

PR #1270, opened 2023-02-23
fix: bug fixes after angular 15 migration

xflord:
* fixed a bug where user could skip to second step of the add member org dialog without actually picking a VO
* removed "No filter" dummy objects from resource assigned users page filters
* fixed a bug where the bans table for the member wouldn't load if the user had no bans, also added Vo object to the EnrichedBanForVo object to avoid exceptions in ban on entity list
* fixed a bug where user could try to add a group inclusion without a VO selected, resulting in an empty dialog

### DIFF
--- a/apps/admin-gui/src/app/facilities/pages/facility-detail-page/facility-allowed-users/facility-allowed-users.component.ts
+++ b/apps/admin-gui/src/app/facilities/pages/facility-detail-page/facility-allowed-users/facility-allowed-users.component.ts
@@ -38,19 +38,14 @@ export class FacilityAllowedUsersComponent implements OnInit {
 
   allowed = true;
 
-  emptyResource: Resource = { id: -1, beanName: 'Resource', name: 'No filter' };
-  resources: Resource[] = [this.emptyResource];
-  filteredResources: Resource[] = [this.emptyResource];
-  selectedResource: Resource = this.emptyResource;
-
-  emptyVo: Vo = { id: -1, beanName: 'Vo', name: 'No filter' };
-  vos: Vo[] = [this.emptyVo];
-  selectedVo: Vo = this.emptyVo;
-
-  emptyService: Service = { id: -1, beanName: 'Service', name: 'No filter' };
-  services: Service[] = [this.emptyService];
-  filteredServices: Service[] = [this.emptyService];
-  selectedService: Service = this.emptyService;
+  resources: Resource[] = [];
+  filteredResources: Resource[] = [];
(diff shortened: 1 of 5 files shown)


apps/admin-gui/src/app/vos/pages/member-detail-page/member-bans/member-bans.component.ts
```diff
@@ -34,7 +39,14 @@ export class MemberBansComponent implements OnInit {
...
       next: (ban) => {
-        this.bans = [{ ban: ban, member: null, vo: null }];
+        if (ban) {
+          this.voService.getVoById(this.member.voId).subscribe({
+            next: (vo) => {
+              this.bans = [{ ban: ban, member: this.member as RichMember, vo: vo }];
+            },
+            error: () => (this.loading = false),
+          });
+        }
```
bodnara, line 49: This should not be needed. The information about member is already displayed since this is the member detail. Same goes for organization.

apps/admin-gui/src/app/facilities/pages/facility-detail-page/facility-allowed-users/facility-allowed-users.component.ts
```diff
@@ -156,47 +154,45 @@ export class FacilityAllowedUsersComponent implements OnInit {
 
   voSelected(vo: Vo): void {
     // prevents "fake" triggers
-    if (this.selectedVo.id === vo.id) {
+    if (this.selectedVo !== undefined && this.selectedVo.id === vo.id) {
       return;
     }
```
bodnara, line 159: I believe we fixed the select component to not emit if the same object is selected. Could you investigate if this check is still needed (same for resources) ?

apps/admin-gui/src/app/vos/pages/member-detail-page/member-bans/member-bans.component.ts
bodnara: I was under the impression that the entity storage had all relevant objects stored which is not the case. Turns out that if related object is required for authorization it is declared like (for this case) `vo: {id: member.voId, beanName: 'Vo'}`. The advantage being that authorization for actions is not blocked but for this case of edit inside of table does not change much.

xflord: ok, will fix like this, thanks

bodnara (approved):

HejdaJakub (approved):

PR #1271, opened 2023-02-23
fix: bug fixes after angular 15 migration

xflord:
* added translations for table exporting in user-profile app
* apps config is now loaded correctly in publications app to avoid exceptions

### DIFF
--- a/apps/publications/src/app/services/publications-config.service.ts
+++ b/apps/publications/src/app/services/publications-config.service.ts
@@ -79,6 +79,7 @@ export class PublicationsConfigService {
           return this.initAuthService
             .loadPrincipal()
             .then(() => this.loadPolicies())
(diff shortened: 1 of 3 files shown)


bodnara (approved):

apps/user-profile/src/assets/config/defaultConfig.json
```diff
@@ -8,7 +8,8 @@
   "instance_favicon": false,
   "auto_auth_redirect": true,
   "supported_languages": [
-    "en"
+    "en",
+    "cs"
```
HejdaJakub, line 12: It looks good otherwise.
```suggestion
    "en"
```

xflord: oops, leftover from testing, fixed

HejdaJakub (approved):

PR #1273, opened 2023-03-02
feat(admin): extended the authorized group column in roles page list

xflord:
* the column now contains ALL the groups the role is acquired through
* changed the header to 'Role acquired through' and made it the last column in the table to make the table easier to understand

### DIFF
--- a/libs/perun/components/src/lib/authorized-groups-cell/authorized-groups-cell.component.html
+++ b/libs/perun/components/src/lib/authorized-groups-cell/authorized-groups-cell.component.html
@@ -1,0 +1,20 @@
+<div class="mt-1">
+  <div class="mb-1" *ngFor="let group of groups.slice(0, itemsShown)">
+    <a
+      class="group-link"
+      [routerLink]="disableRouting ? null : ['/organizations', group.voId, 'groups', group.id]"
+      [perunWebAppsMiddleClickRouterLink]="disableRouting ? null : ['/organizations', group.voId.toString(), 'groups', group.id.toString()]"
+      (auxclick)="$event.preventDefault()"
+      [class.cursor-pointer]="!disableRouting"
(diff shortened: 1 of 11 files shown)


libs/perun/components/src/lib/authorized-groups-cell/authorized-groups-cell.component.html
```diff
@@ -0,0 +1,18 @@
+<div class="mt-1">
+  <div
+    class="mb-1"
+    *ngFor="let group of groups.slice(0, itemsShown)"
+    [routerLink]="disableRouting ? null : ['/organizations', group.voId, 'groups', group.id]"
+    [perunWebAppsMiddleClickRouterLink]="disableRouting ? null : ['/organizations', group.voId.toString(), 'groups', group.id.toString()]"
+    [class.cursor-pointer]="!disableRouting">
+    {{authzVoNames.get(group.voId) + " : " + group.name}}
```
HejdaJakub, line 8: I would like something like that for that link on hover on the chosen group: (we can discus it on daily)
```suggestion
    <span class="active-link">{{authzVoNames.get(group.voId) + " : " + group.name}}</span>
```

libs/perun/components/src/lib/authorized-groups-cell/authorized-groups-cell.component.scss
```diff
@@ -0,0 +1,3 @@
+.cursor-pointer {
+  cursor: pointer;
+}
```
HejdaJakub, line 3:
```suggestion
}

.active-link:hover {
  color: var(--group-color);
  text-decoration: underline;
}
```
Screenshot from 2023-03-02 16-12-56

HejdaJakub (commented):
It looks good. I have just one suggestion.

bodnara (approved):

HejdaJakub (approved):

PR #1275, opened 2023-03-07
update to Angular 15

HejdaJakub:
This PR includes:
* update to Angular 15
* migration to MDC (Material Design Component) standard
* update to bootstrap 5
* fix for the wrong initialization of color themes through CSS variables (fixed by script)
* fixes for several bugs caused by this update and migration


mattjoke (approved):
Good job 👍

PR #1276, opened 2023-03-08
fix(admin): fixed a bug where consent status column would not display…

xflord:
… on facility assigned users page


bodnara:
This way we go back on our decision that we would not hide content during loading.

HejdaJakub (approved):

bodnara (approved):

PR #1277, opened 2023-03-08
fix: fixed displaying filters in search-select

xkostka2:
* fixed bug where filters were not displayed in the search-select field


libs/perun/components/src/lib/entity-search-select/entity-search-select.component.html
```diff
@@ -36,7 +36,7 @@
         [style.height.px]="getViewportHeight()"
         #scrollViewport
         [minBufferPx]="5 * 48"
-        [maxBufferPx]="10 * 48">
+        [maxBufferPx]="(entitiesLen > 5 ? entitiesLen : 5) * 48">
```
bodnara, line 39: If I understand it correctly the max buffer sets the max size of virtual scroll view port. So if we set it to the length of all elements we negate the purpose of having virtual scroll, which is to prevent thousands of elements being present in the DOM.

xkostka2: This bug occurred in the first place because selected elements were not present in the DOM. Another possible fix could be moving them all to the start of the list if it works for you.

libs/perun/components/src/lib/entity-search-select/entity-search-select.component.html
```diff
@@ -36,7 +36,7 @@
         [style.height.px]="getViewportHeight()"
         #scrollViewport
         [minBufferPx]="5 * 48"
-        [maxBufferPx]="10 * 48">
+        [maxBufferPx]="(selectedEntities.length > 10 ? selectedEntities.length : 10) * 48">
```
bodnara, line 39: This should not be needed anymore.

bodnara (approved):

PR #1283, opened 2023-03-14
feat(profile): remove the option to add admin ssh key

xflord:
* the logic is no longer used


mattjoke (approved):

HejdaJakub:
You forgot to remove it from `admin-gui/en.json`, so remove it also there, please.

PR #1284, opened 2023-03-15
fix(admin):color migration luminance fix

mattjoke:
* fixed error with luminance, removed usage of color.isLight() when computing contrast
* Using contrast and luminance equations to pick the correct contrast value.

### DIFF
--- a/color-migration.js
+++ b/color-migration.js
@@ -66,6 +66,43 @@ const getConfigs = async () => {
   return configs;
 }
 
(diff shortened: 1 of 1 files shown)


mattjoke:
* on production builds, luminance is not computed correctly

HejdaJakub (approved):
We could simplify some reused calculations but whilst this should be just a temporary script (I hope), you can probably skip it as this implementation works correctly.

xflord (approved):

PR #1287, opened 2023-03-16
fix(admin): repair all e2e tests after angular 15 upgrade

HejdaJakub:
* All e2e test has been reviewed and fixed.
* Cypress has been upgraded to the latest version.
* One bug fix for the task-result-list is also included in this PR.


HejdaJakub:
I am not sure about the consistency of these tests cause some of them sometimes failed for me - but only sometimes and also usually only during the execution from the command line, not from the UI (it looks more consistent from UI - maybe because it is slower(?)). But the last three attempts after some changes look good and the whole scenario finished successfully for me (for all three last attempts from the command line), so there is the current state of these tests.

Screenshot from 2023-03-16 18-43-16

xflord (approved):
great job

PR #1289, opened 2023-03-22
feat(profile): visual rework of the MFA settings

xflord:
* MFA settings are now part of an expandable accordion


bodnara (commented):
My only suggestion would be to put the label on the left side of the toggle. Since the label also acts an "activate" it can cause accidental change when user is trying to expand the panel.

xflord:
Replaced slide toggles with checkboxes (with indeterminate states) and refactored the component a bit

apps/user-profile/src/app/pages/settings-page/settings-authorization/mfa-settings/mfa-settings.component.html
```diff
@@ -1,25 +1,48 @@
...
+            class="toggle-ellipsis mt-2 me-5"
+            [disabled]="!mfaAvailable"
+            (click)="$event.stopPropagation()"
+            [(ngModel)]="enforceMfa"
+            (change)="toggleEnableMfa()"
+            [indeterminate]="enforceMfa && (!allRpsSelected || categorySelection.selected.length !== categoriesList.length)"
+            >{{'AUTHENTICATION.MFA_TOGGLE' | customTranslate | translate}}
+          </mat-checkbox>
+        </mat-expansion-panel-header>
+        <mat-spinner class="ms-auto me-auto mt-2" *ngIf="loadingCategories"></mat-spinner>
```
HejdaJakub, line 21: IMHO this spinner is not necessary since the first spinner also depends on the `loadingCategories`.
```suggestion
```

apps/user-profile/src/app/pages/settings-page/settings-authorization/mfa-settings/mfa-settings.component.html
```diff
@@ -1,25 +1,48 @@
...
+        <mat-accordion multi *ngIf="!loadingCategories">
+          <mat-expansion-panel *ngFor="let category of categories | keyvalue">
+            <mat-expansion-panel-header>
+              <mat-checkbox
+                class="toggle-ellipsis mt-2"
```
HejdaJakub, line 26: I would add also `me-2` class to avoid bad UX near the arrow due to the long category name.
image
```suggestion
                class="toggle-ellipsis mt-2 me-2"
```

HejdaJakub (commented):
These checkboxes look better than slide toggles, good job. I have just a few small suggestions.

bodnara (commented):
Some UI changes:
- put the save button above the accordion, right now on small screens the button is placed under
- make the manage MFA button stroked, it is not the primary action
- when user does not have available MFA show info alert instead of showing a disabled accordion with a tooltip
- filter out categories with no service ?

image

I appreciate the refactor, however the  `MfaSettingsComponent` is still very convoluted. Most of the logic handled here should be the responsibility of the `MfaApiService`. It would make sense if the service fetches the data, parses it and returns a nice to work with objects to the caller. Than recieves the objects with some changes and again formats it so the mfa endpoint can handle it.

HejdaJakub:
Do not forget - when we will rework this by moving the button above and showing an info alert, it would be also good to not display `Save settings` button when a user doesn't have any token.
> * when user does not have available MFA show info alert instead of showing a disabled accordion with a tooltip
> 
> image
>

HejdaJakub (changes requested):

apps/user-profile/src/app/pages/settings-page/settings-authorization/mfa-settings/mfa-settings.component.ts
```diff
@@ -49,11 +55,19 @@ export class MfaSettingsComponent implements OnInit {
...
+    });
+
+    this.mfaUrl = this.translate.currentLang === 'en' ? mfa.url_en : mfa.url_cs;
+    this.categorySelection = new SelectionModel<string>(true, []);
     this.enableDetailSettings = this.store.getProperty('mfa').enable_detail_settings;
     this.mfaApiService.isMfaAvailable().subscribe({
       next: (isAvailable) => {
         this.mfaAvailable = isAvailable;
-        if (isAvailable) {
+        if (this.mfaAvailable) {
```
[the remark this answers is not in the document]
xflord: Also removed the line from 'loadSettings()' since it was redundant there

bodnara (commented):
When trying to save the settings I get an error resulting in an infinite load. Some property is undefined.

apps/user-profile/src/app/pages/settings-page/settings-authorization/mfa-settings/mfa-settings.component.ts
HejdaJakub, line 1: I also think that in `loadMfa()` method in all error cases should be added `this.loadingCategories = false;`

HejdaJakub (commented):
It would be nice to test it also with real data (which is in progress), but from the side of code review and testing on old data, it looks good now (just some small fixes of turning off the spinner).

HejdaJakub:
When  I wanted to save settings, I got this error:
image

HejdaJakub (approved):

bodnara (commented):
Good job with the refactor and everything. My last nitpick would be to set default cursor for categories that are not expandable. After that we can merge.

PR #1291, opened 2023-03-27
feat: use cookies in requests

Johaney-s:
* without sending cookies, sessions for service-access (basic auth) are renewed on backend for each request
* that causes extsource loa to be updated which then updates relevant attribute (2 audit messages for EACH request!)
* this was not a problem for oidc access, because session was not renewed in case of missing cookies
* withCredentials header will solve problem for renewing sessions, but requests will fail if apache configuration does not allow domain for this behavior
* localhost is not supported origin, therefore we only add withCredentials header to calls from domain not coming from localhost

### DIFF
--- a/libs/perun/utils/src/lib/perun-utils.ts
+++ b/libs/perun/utils/src/lib/perun-utils.ts
@@ -868,3 +868,10 @@ export function isMemberIndirectString(member: RichMember): string {
 export function containsExactlyInAnyOrder(array1: string[], array2: string[]): boolean {
   return array1.length === array2.length && array1.every((v) => array2.includes(v));
 }
+
+/**
+ * Return true, if GUI is run locally, false otherwise.
+ */
+export function isRunningLocally(): boolean {
+  return location.hostname === 'localhost' || location.hostname === '127.0.0.1';
+}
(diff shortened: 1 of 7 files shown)


libs/perun/utils/src/lib/perun-utils.ts
```diff
@@ -864,3 +864,10 @@ export function isMemberIndirectString(member: RichMember): string {
   }
   return 'UNKNOWN';
 }
+
+/**
+ * Return true, if GUI is run locally, false otherwise.
+ */
+export function isRunningLocally(): boolean {
+  return location.hostname === 'localhost' || location.hostname === '127.0.0.1';
```
xflord, line 872: The hostname could differ when running local GUI and [rerouting localhost](https://perunaai.atlassian.net/wiki/spaces/STRIBOG/pages/66879489/Https+localhost) to enable https. Can this be an issue?

HejdaJakub (approved):

libs/perun/utils/src/lib/perun-utils.ts
HejdaJakub: In this case, we want to pretend non-localhost access and the GUI should behave like it is not the localhost, so this shouldn't be a problem. :no_good_man:

HejdaJakub:
Tested from localhost (oidc and basic auth) and from https localhost (again oidc and basic auth) and it works :tada:

PR #1293, opened 2023-03-28
feat(admin): allow deletion of users according to config property

HejdaJakub:
* There is new config property 'user_deletion_forced' which defines if users on instance can be deleted or anonymized (this is the default option).
* This task is related to new backend implementation where deletion of user is now connected with blocking of user's logins.
* The new configuration property 'user_deletion_forced' (default false) was added. When this property is not set, still the default anonymization will be applied.

BREAKING-CHANGE: new configuration property 'user_deletion_forced'

### DIFF
--- a/apps/admin-gui/src/app/admin/pages/admin-user-detail-page/admin-user-detail-page.component.html
+++ b/apps/admin-gui/src/app/admin/pages/admin-user-detail-page/admin-user-detail-page.component.html
@@ -22,8 +22,14 @@
             data-cy="edit-user-button">
             <mat-icon>edit</mat-icon>
           </button>
-          <button *ngIf="!anonymized" (click)="anonymizeUser()" mat-icon-button>
+          <button
+            *ngIf="!anonymized && !userDeletionForced"
+            (click)="anonymizeUser()"
+            mat-icon-button>
             <mat-icon>no_accounts</mat-icon>
+          </button>
+          <button *ngIf="userDeletionForced" (click)="deleteUser()" mat-icon-button>
+            <mat-icon>delete_forever</mat-icon>
           </button>
           <mat-icon
             *ngIf="anonymized"
--- a/apps/admin-gui/src/assets/i18n/en.json
+++ b/apps/admin-gui/src/assets/i18n/en.json
@@ -2192,6 +2192,11 @@
       "GROUP_RELATION": "User's membership in all groups",
       "SUCCESS_NOTIFICATION": "User was successfully anonymized"
     },
+    "DELETE_USER": {
+      "VO_RELATION": "User's membership in all organizations",
+      "GROUP_RELATION": "User's membership in all groups",
+      "SUCCESS_NOTIFICATION": "User was successfully deleted"
+    },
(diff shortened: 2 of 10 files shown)


apps/admin-gui/src/app/admin/pages/admin-user-detail-page/admin-user-detail-page.component.html
```diff
@@ -22,9 +22,15 @@
...
             <mat-icon>edit</mat-icon>
           </button>
-          <button *ngIf="!anonymized" (click)="anonymizeUser()" mat-icon-button>
+          <button
+            *ngIf="!anonymized && !userDeletionForced"
+            (click)="anonymizeUser()"
+            mat-icon-button>
             <mat-icon>no_accounts</mat-icon>
           </button>
+          <button *ngIf="!anonymized && userDeletionForced" (click)="deleteUser()" mat-icon-button>
```
bodnara, line 31: It should be possible to delete anonymized users. In case the instance switches to forced deletion after some users were already anonymized.
```suggestion
          <button *ngIf="userDeletionForced" (click)="deleteUser()" mat-icon-button>
```

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -2192,6 +2192,11 @@
       "GROUP_RELATION": "User's membership in all groups",
       "SUCCESS_NOTIFICATION": "User was successfully anonymized"
     },
+    "DELETE_USER": {
+      "VO_RELATION": "User's membership in all organizations",
+      "GROUP_RELATION": "User's membership in all groups",
```
bodnara, line 2197: Since this action is quite destructive are there other relations worth mentioning ? I am not sure but we can discuss.

libs/perun/dialogs/src/lib/delete-user-dialog/delete-user-dialog.component.ts
```diff
@@ -0,0 +1,67 @@
...
+@Component({
+  selector: 'perun-web-apps-delete-user-dialog',
+  templateUrl: './delete-user-dialog.component.html',
+  styleUrls: ['./delete-user-dialog.component.scss'],
+})
+export class DeleteUserDialogComponent implements OnInit {
+  theme: string;
+  force = false;
+  loading: boolean;
+  displayedColumns: string[] = ['name'];
```
bodnara, line 22:
```suggestion
```

libs/perun/models/src/lib/ConfigProperties.ts
```diff
@@ -152,6 +152,7 @@ export interface PerunConfig {
   // Optional
   auto_auth_redirect?: boolean;
   supported_languages?: string[];
+  user_deletion_forced?: boolean;
```
bodnara, line 155: This is a admin gui specific property, also it should not be optional imho.

bodnara (changes requested):

bodnara:
Merge after https://github.com/CESNET/perun/pull/3908.

bodnara (approved):

PR #1296, opened 2023-04-14
feat: make table headers sticky and tables scrollable

xflord:
* all tables now have a maximum height set and their scrollbar
* headers are sticky
* removed redundant extra div in attr def list, which caused some issues

### DIFF
--- a/apps/admin-gui/src/_styles.scss
+++ b/apps/admin-gui/src/_styles.scss
@@ -1679,3 +1679,10 @@ table .mdc-text-field--outlined {
 .mdc-data-table__cell {
   border-bottom: 1px solid rgba(0, 0, 0, 0.12) !important;
 }
+
+// sticky table headers
+.mat-mdc-header-cell {
+  position: sticky !important;
+  top: 0;
+  z-index: 100;
+}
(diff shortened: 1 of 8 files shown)


bodnara:
The header column should be sticky as well so there is no confusion about what column is what. This was one of the reasons for the change as well.

bodnara:
Since the pagination will be sticky there is no need for the bottom pagination and  perhaps even the `scrollIntoView` can be removed. I created a separate task for that but it can be done in this PR if that would make more sense.

HejdaJakub:
> Since the pagination will be sticky there is no need for the bottom pagination and perhaps even the `scrollIntoView` can be removed. I created a separate task for that but it can be done in this PR if that would make more sense.

I agree that the bottom pagination can be removed, but I am not sure if we want to remove also scrolling to the top. I think that it is the most common thing when you are clicking on the next page that you want to continue to the top of the new page. If you have 25 or 100 rows, it can be maybe more confusing if you will be able to click on the next page also somewhere in the middle of 100 rows and appear again somewhere in the middle of the next page. Or do you think that this is correct and expected behavior?

xflord:
> The header column should be sticky as well so there is no confusion about what column is what. This was one of the reasons for the change as well.

It is. Make sure to load the changes to 'styles.scss' files by running the color migration/postinstall script.

xflord:
Made the top nav bar scroll to the top, left bottom nav bar untouched for now (we can discuss if we want to remove it tomorrow at daily). Also made the table smaller to make it less likely for user to scroll past the top part of the table (in case we decide to remove bottom nav bar), let me know what you think.

bodnara (approved):

xkostka2 (approved):

PR #1297, opened 2023-04-14
fix(admin): services status bulk operations

sarkapalkovicova:
Used bulk operations for forcing propagation, blocking, and unblocking services.


bodnara:
Merge after https://github.com/CESNET/perun/pull/3932

bodnara (commented):
We could also remove the translations and their variables as this is now handled by the notification service.

bodnara (approved):

HejdaJakub (approved):

PR #1301, opened 2023-04-18
fix(admin): load ext sources from config file when opening ext source…

xflord:
…s admin page

### DIFF
--- a/apps/admin-gui/src/app/admin/pages/admin-page/admin-ext-sources/admin-ext-sources.component.html
+++ b/apps/admin-gui/src/app/admin/pages/admin-page/admin-ext-sources/admin-ext-sources.component.html
@@ -1,7 +1,9 @@
 <div>
   <h1 class="page-subtitle">{{'ADMIN.EXT_SOURCES.TITLE' | translate}}</h1>
-  <perun-web-apps-refresh-button (click)="onLoad()"></perun-web-apps-refresh-button>
-
+  <perun-web-apps-refresh-button (click)="refreshTable()"></perun-web-apps-refresh-button>
+  <button (click)="loadConfigExtSources()" class="me-2" mat-stroked-button>
+    {{'ADMIN.EXT_SOURCES.LOAD' | translate}}
+  </button>
   <perun-web-apps-immediate-filter
     (filter)="applyFilter($event)"
     [placeholder]="'ADMIN.EXT_SOURCES.SEARCH'"></perun-web-apps-immediate-filter>
--- a/apps/admin-gui/src/app/admin/pages/admin-page/admin-ext-sources/admin-ext-sources.component.ts
+++ b/apps/admin-gui/src/app/admin/pages/admin-page/admin-ext-sources/admin-ext-sources.component.ts
@@ -1,6 +1,5 @@
 import { Component, OnInit } from '@angular/core';
-import { GuiAuthResolver, NotificatorService } from '@perun-web-apps/perun/services';
-import { TranslateService } from '@ngx-translate/core';
+import { GuiAuthResolver } from '@perun-web-apps/perun/services';
 import { ExtSource, ExtSourcesManagerService } from '@perun-web-apps/perun/openapi';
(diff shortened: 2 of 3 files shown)


apps/admin-gui/src/app/admin/pages/admin-page/admin-ext-sources/admin-ext-sources.component.ts
```diff
@@ -37,15 +34,19 @@ export class AdminExtSourcesComponent implements OnInit {
...
-      this.refreshTable();
+  refreshTable(): void {
+    this.loading = true;
+    this.extSourceService.loadExtSourcesDefinitions().subscribe({
+      next: () => {
+        this.getExtSources();
+      },
+      error: () => {
+        this.getExtSources();
+      },
```
HejdaJakub, line 45: According to the task definition, is `loadExtSourcesDefinitions()` necessary if there is no further action with a returned value? 
```suggestion
    this.getExtSources();
```

apps/admin-gui/src/app/admin/pages/admin-page/admin-ext-sources/admin-ext-sources.component.ts
HejdaJakub, line 1: Also remove unused imports, please.

HejdaJakub (changes requested):

apps/admin-gui/src/app/admin/pages/admin-page/admin-ext-sources/admin-ext-sources.component.ts
bodnara: My question would be if this is the desired way of loading the ext sources from config. Shouldn't the database be automatically filled with the contents of the config ? Or can it be changed during run time ? It not there is no reason to load the config on every refresh.

apps/admin-gui/src/app/admin/pages/admin-page/admin-ext-sources/admin-ext-sources.component.html
```diff
@@ -1,7 +1,13 @@
...
-  <perun-web-apps-refresh-button (click)="onLoad()"></perun-web-apps-refresh-button>
-
+  <perun-web-apps-refresh-button (click)="refreshTable()"></perun-web-apps-refresh-button>
+  <button
+    (click)="loadConfigExtSources()"
+    color="accent"
+    class="action-button me-2"
+    mat-flat-button>
+    {{'ADMIN.EXT_SOURCES.LOAD' | translate}}
+  </button>
```
bodnara, line 10:
```suggestion
  <button
    (click)="loadConfigExtSources()"
    class="action-button me-2"
    mat-stroked-button>
    {{'ADMIN.EXT_SOURCES.LOAD' | translate}}
  </button>
```

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -2498,7 +2498,7 @@
     "EXT_SOURCES": {
       "TITLE": "External sources",
       "SEARCH": "Search by Id, Name or Type",
-      "LOAD_SUCCESS": "External sources successfully loaded"
+      "LOAD": "Load external sources from config file"
```
bodnara, line 2501: Button labels should be kept short at least imho.
```suggestion
      "LOAD": "Load configuration"
```

apps/admin-gui/src/app/admin/pages/admin-page/admin-ext-sources/admin-ext-sources.component.html
xflord: Was meaning to do this as well, but decided against it since I thought you would be mad it's not consistent with the rest of the buttons :D

bodnara: Well the way it was before it was inconsistent so I guess you were right.

bodnara (approved):

HejdaJakub (approved):

PR #1302, opened 2023-04-19
fix(admin): minor fixes to facility hosts page

xflord:
* removed double load of the facility entity
* added pointer cursor to the hosts list
* used one-entity-attribute-page.component on the hosts detail page

### DIFF
--- a/apps/admin-gui/src/app/shared/components/one-entity-attribute-page/one-entity-attribute-page.component.ts
+++ b/apps/admin-gui/src/app/shared/components/one-entity-attribute-page/one-entity-attribute-page.component.ts
@@ -85,6 +85,12 @@ export class OneEntityAttributePageComponent implements OnInit {
           this.loading = false;
         });
         break;
+      case 'host':
+        this.attributesManagerService.getHostAttributes(this.entityId).subscribe((attributes) => {
+          this.attributes = attributes;
+          this.selection.clear();
+          this.loading = false;
+        });
     }
   }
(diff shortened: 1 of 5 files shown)


apps/admin-gui/src/app/shared/components/one-entity-attribute-page/one-entity-attribute-page.component.ts
```diff
@@ -85,6 +85,12 @@ export class OneEntityAttributePageComponent implements OnInit {
           this.loading = false;
         });
         break;
+      case 'host':
+        this.attributesManagerService.getHostAttributes(this.entityId).subscribe((attributes) => {
+          this.attributes = attributes;
```
HejdaJakub, line 90: I think that there should be also this according to the old behavior.
```suggestion
          this.attributes = filterCoreAttributes(attributes);
```

HejdaJakub (changes requested):

apps/admin-gui/src/app/shared/components/one-entity-attribute-page/one-entity-attribute-page.component.ts
HejdaJakub: EDIT: `filterCoreAttributes` is called in attribute-list, so there it would be redundant

mattjoke (approved):

bodnara (approved):

PR #1304, opened 2023-04-19
refactor(admin): minor refactoring in the admin pages

xflord:
* refactored calling of methods in searcher
* fixed double load on services page init


HejdaJakub (approved):

bodnara (commented):
The task was to remove the switch not the methods, since it does not serve any purpose only adds unnecessary logic of passing around `entity`.

mattjoke (approved):

PR #1307, opened 2023-04-24
feat(admin): e2e login command

sarkapalkovicova:
Added and used new custom command for login.

### DIFF
--- a/apps/admin-gui-e2e/src/support/commands.ts
+++ b/apps/admin-gui-e2e/src/support/commands.ts
@@ -23,3 +23,13 @@
 //
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+
+Cypress.Commands.add('login', (envRole: string, baPrincipalName: string) => {
+  envRole = envRole ? '_' + envRole : envRole;
+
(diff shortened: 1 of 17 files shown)


mattjoke (changes requested):
I'm unsure about that line I mentioned in the comment before. Otherwise, the code looks good.

apps/admin-gui-e2e/src/support/commands.ts
```diff
@@ -23,3 +23,13 @@
 //
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+
+Cypress.Commands.add('login', (envRole: string, baPrincipalName: string) => {
+  envRole = envRole ? '_' + envRole : envRole;
+
+  sessionStorage.setItem('baPrincipal', `{"name": "${baPrincipalName}"`);
+  sessionStorage.setItem('basicUsername', Cypress.env('BA_USERNAME' + envRole) as string);
```
HejdaJakub, line 31: It should be achievable like this (the same also for the password).
```suggestion
  sessionStorage.setItem('basicUsername', String(Cypress.env('BA_USERNAME' + envRole)));
```

mattjoke (approved):
Looks good 👍

xflord (approved):
well done

PR #1308, opened 2023-04-25
fix(admin): fix misaligned ellipsis in sponsored member creation drop…

michalberky:
…down menu

Added inline-block display property to each action-button on the page. 
Changed margin-left of action-button::after to properly align it.

### DIFF
--- a/apps/admin-gui/src/_styles.scss
+++ b/apps/admin-gui/src/_styles.scss
@@ -1581,8 +1581,13 @@ td.mat-mdc-cell {
   contain: size !important;
 }
 
+.action-button {
+  display: inline-block !important;
+}
+
 .action-button::after {
   content: '\2026';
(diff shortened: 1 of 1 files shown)


HejdaJakub (changes requested):
I know, that the task was defined just for the page `sponsored-members`, but I would prefer to move this fix to a global level for this reasons:
1) We will have this solution in one place, so we will avoid this bug also in the future (we will not have to think about it in the future and reuse this solution in more places)
2) We have at least two other pages now, where this bug also occurs (`vo-members`, `group-members`)
image

So you can add `.action-button` class to `_styles.scss` with `display: inline-block !important;` and also add the left margin to the existing `.action-button::after` in `_styles.scss` .

HejdaJakub (approved):

PR #1309, opened 2023-04-27
test(admin): refactor structure of e2e tests

HejdaJakub:
* Structure of e2e tests has been changed - they are now structured to GUI sections (vos, groups, facilities, resources, members, admin, my-profile) + there is one folder for shared logic (e.g. login).
* There is also initial implementatoin of "test code structure" with describe and context.
* I also changed commented tests to it.skip / context.skip to follow some cypress conventions.
* According to new structure I also extracted one test to the new file vos/perun-admin.cy.js to follow this new stucture.


HejdaJakub:
Merge after this MR (data for delete VO test): https://gitlab.ics.muni.cz/perun/e2e-perun/-/merge_requests/24

HejdaJakub:
NOTE: I created a new structure for the e2e test and I also prepared "code structure" using `describe` and `context`. My idea is to have one `describe` block and then structure the test into several contexts. Now I have a different attitude to structuring each test because it didn't make sense to me to create a new context for each sub-section if we have e.g. just one test for it. According to this, I created contexts like the `Advanced settings`, but there are now also contexts like `Services`/`Attribute definitions` or `Advanced settings - Managers`. I know that it's not enough consistent, but I don't have any bulletproof solution now if we don't even know which tests we will want to add in the future. So I think, that we can create new/rework old contexts according to new tests - if there will be a more meaningful structure. There are some examples of the current code structure.
Screenshot from 2023-04-27 16-40-09
Screenshot from 2023-04-27 16-41-47

xflord (approved):
love it, gj

PR #1310, opened 2023-04-29
feat(admin): prototype of e2e tests checking for console errors

xflord:


apps/admin-gui-e2e/src/e2e/crawler.cy.js
```diff
@@ -0,0 +1,67 @@
...
+  });
+  // watch the video "Crawl Local Pages Using Cypress"
+// https://youtu.be/FDNeiwKWdb4
+  it('crawls all local pages', () => {
+    // let's keep track of all URLs we've visited
+    // NOTE: these URLs do not disambiguate between pages
+    // so if /about.html and /about are the same page,
+    // we will visit it twice if there are links to it
+    const visited = new Set()
+    const toVisit = ['/' , '/admin', '/myProfile']
```
Kuliak, line 22: Shouldn't we check '/facilities' and '/organizations' as well?

xflord: I omitted those because this implementation wouldn't find any new links on them anyway. 
This crawler is not intended for merge, just for people to see how it could be implemented. The problem is that our Angular app mostly uses onClick events to route, whereas this only looks for anchor elements. I couldn't figure out how to reliably get all the onClick event elements and retrieve the links from them.

HejdaJakub (changes requested):
Sorry for this late review, but I think it would be nice to fix (as a part of this PR) also the tests that are failing due to this new check - several tests are now failing due to `ExpressionChangedAfterItHasBeenCheckedError` in the console (and maybe due to some another errors).
image

HejdaJakub (commented):
It looks good now, I just think that you should include also changes in `package-lock.json` in this PR.

HejdaJakub (approved):

PR #1312, opened 2023-05-03
fix(admin): immediate replaced with debounce filter

sarkapalkovicova:
All occurrences of immediate filter were replaced with debounce filter.


mattjoke (approved):
Looks good 👍

PR #1314, opened 2023-05-04
feat: handle proxy/local logout and related post logout redirect

HejdaJakub:
* Logout process now differs when the new property for proxy_logout is set to true/false and if the oauth_post_logout_redirect_uri exists or not.
* If the oauth_post_logout_redirect_uri does NOT exist, user is redirected to the login page with info about logout.
* If uri exists and the redirection should be handled by proxy, user is temporarily redirected to the logout page where is displayed the loader (then user should be redirected by proxy).
* If uri exists and the redirection should be handled locally, user is directly redirected to this uri.

BREAKING CHANGE: new mandarory property proxy_logout: boolean (added to admin-gui, user-profile, publications and consolidator), by default is true


Kuliak (commented):
It looks good, however, I had some troubles with testing. It would be fine to test it before merging.

libs/perun/services/src/lib/auth.service.ts
```diff
@@ -46,14 +47,35 @@ export class AuthService {
...
+      const postLogoutUrl = this.store.getProperty('oidc_client').oauth_post_logout_redirect_uri;
+      if (!postLogoutUrl) {
+        // redirect to the login page if there is no postLogoutUrl
+        void this.router.navigate(['/login'], { queryParamsHandling: 'preserve' });
+      } else if (this.store.getProperty('proxy_logout')) {
+        // redirect to the logout loading page if postLogoutUrl exist and logout should be handled by proxy
+        void this.router.navigate(['/logout'], { queryParamsHandling: 'preserve' });
+      } else {
+        // directly redirect if postLogoutUrl exist and logout should be handled locally (not by proxy)
+        void this.router.navigate([postLogoutUrl], { queryParamsHandling: 'preserve' });
```
bodnara, line 62: I assumed the post logout url is somewhere outside of the web app. But the directly redirect locally case navigates to `instanceUrl/postLogoutUrl`. But maybe this is what was expected ?

apps/admin-gui/src/assets/config/defaultConfig.json
```diff
@@ -22,6 +22,7 @@
     "oauth_response_type": "code",
     "oauth_offline_access_consent_prompt": true
   },
+  "proxy_logout": true,
```
bodnara, line 25: Should we also add new config property to this repo?
 https://github.com/CESNET/ansible-role-perun-docker-server

HejdaJakub:
Fixed `LOGOUT_INFO` to "logged out". Fixed the navigation for local logout to enable any post-logout url (not only redirection inside the app).

I also realized that there are missing routes in app-routing in other apps (user-profile, consolidator, and publications). Linker and password reset don't have enabled the logout functionality, so there it's not needed (so I also removed `LOGOUT_INFO` from password-reset en/cs.json).

I tested it manually for all 4 use cases and it works (if the `proxy_logout` property is set correctly).

apps/admin-gui/src/assets/config/defaultConfig.json
HejdaJakub: Done: https://github.com/CESNET/ansible-role-perun-docker-server/pull/69

Kuliak (approved):

PR #1316, opened 2023-05-11
fix(admin): disable deselect button

sarkapalkovicova:
Deselect button was disabled in places where it was not needed.

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/application-form-copy-items-dialog/application-form-copy-items-dialog.component.ts
+++ b/apps/admin-gui/src/app/shared/components/dialogs/application-form-copy-items-dialog/application-form-copy-items-dialog.component.ts
@@ -1,6 +1,5 @@
 import { ChangeDetectorRef, Component, Inject, OnInit } from '@angular/core';
 import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
-import { TranslateService } from '@ngx-translate/core';
 import {
   Group,
   GroupsManagerService,
@@ -8,8 +7,13 @@ import {
   Vo,
   VosManagerService,
 } from '@perun-web-apps/perun/openapi';
-import { ApiRequestConfigurationService, NotificatorService } from '@perun-web-apps/perun/services';
+import {
+  ApiRequestConfigurationService,
+  NotificatorService,
+  PerunTranslateService,
+} from '@perun-web-apps/perun/services';
 import { RPCError } from '@perun-web-apps/perun/models';
+import { compareFnName } from '@perun-web-apps/perun/utils';
 
 export interface ApplicationFormCopyItemsDialogData {
   voId: number;
@@ -27,6 +31,7 @@ export class ApplicationFormCopyItemsDialogComponent implements OnInit {
   groups: Group[] = [];
   selectedVo: Vo;
   selectedGroup: Group = null;
+  fakeGroup: Group;
   theme: string;
   loading = false;
   private successMessage: string;
@@ -37,7 +42,7 @@ export class ApplicationFormCopyItemsDialogComponent implements OnInit {
     private dialogRef: MatDialogRef<ApplicationFormCopyItemsDialogComponent>,
     private voService: VosManagerService,
     private groupService: GroupsManagerService,
-    private translateService: TranslateService,
+    private translateService: PerunTranslateService,
     private registrarManager: RegistrarManagerService,
     private notificatorService: NotificatorService,
     private apiRequest: ApiRequestConfigurationService,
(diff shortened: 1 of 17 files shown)


HejdaJakub:
I am not sure, but I think that there are much more places, where we don't want to use this clear button (e.g. I tried several places with `vo-search-select` - searcher, copy-notifications, user-accounts - and there we don't want this button). IMHO it would be much easier and more straightforward for future work with this if we by default hide this button (and ideally only at one place, if it is possible, like at `entity-search-select`) and enable it only in places, where it makes sense.

libs/perun/components/src/lib/user-search-select/user-search-select.component.ts
```diff
@@ -10,6 +10,7 @@ import { compareFnUser, parseFullName } from '@perun-web-apps/perun/utils';
 export class UserSearchSelectComponent implements OnInit {
   @Input() users: User[];
   @Input() disableAutoSelect = false;
+  @Input() disableDeselectButton = true;
```
bodnara, line 13: Please also pass the value to the underlying entity select.

libs/perun/components/src/lib/string-search-select/string-search-select.component.ts
```diff
@@ -10,7 +10,7 @@ export class StringSearchSelectComponent {
   @Input() preselectedValues: string[];
   @Input() selectPlaceholder: string;
   @Input() mainTextFunction: (string) => string;
-  @Input() disableDeselectButton = false;
+  @Input() disableDeselectButton = true;
```
bodnara, line 13: Please also pass the value to the underlying entity select.

libs/perun/components/src/lib/entity-search-select/entity-search-select.component.ts
```diff
@@ -143,6 +143,7 @@ export class EntitySearchSelectComponent<T extends PerunBean>
         this.entitySelected.emit(value);
       }
     }
+    // this.visibleDeselectButton = !this.disableDeselectButton && (this.selectedEntities.length !== 0 || this.entitySelected != null);
     this.visibleDeselectButton = !this.disableDeselectButton && this.selectedEntities.length !== 0;
```
bodnara, line 147: I would prefer for the button to be always visible if it is enabled. Maybe it can be in disabled state if nothing is selected.  But removing it completely seems distracting.
I am also strongly considering we should remove the text from the button. It is a simple action and right now it takes way too much space, especially if we have multiple selections on the screen.

apps/admin-gui/src/app/shared/components/dialogs/application-form-copy-items-dialog/application-form-copy-items-dialog.component.ts
```diff
@@ -61,11 +63,25 @@ export class ApplicationFormCopyItemsDialogComponent implements OnInit {
     this.loading = true;
     this.theme = this.data.theme;
     this.translateService.get('DIALOGS.APPLICATION_FORM_COPY_ITEMS.NO_GROUP_SELECTED').subscribe(
-      () => {
+      (text: string) => {
+        this.fakeGroup = {
+          id: -1,
```
bodnara, line 68: Could you please, replace the deprecated `subscribe` as well as old `TranslateService` in this file.

bodnara (changes requested):

bodnara (approved):
Looks good, also thanks for fixing deprecated parts.

libs/perun/components/src/lib/user-search-select/user-search-select.component.html
```diff
@@ -1,6 +1,7 @@
 <perun-web-apps-entity-search-select
   [entities]="users"
   [disableAutoSelect]="disableAutoSelect"
+  [disableDeselectButton]="disableDeselectButton"
```
HejdaJakub, line 4:
```suggestion
```

libs/perun/components/src/lib/user-search-select/user-search-select.component.ts
HejdaJakub, line 13:
```suggestion
```

libs/perun/components/src/lib/vo-search-select/vo-search-select.component.html
```diff
@@ -6,6 +6,7 @@
   [searchFunction]="searchFunction"
   [mainTextFunction]="nameFunction"
   [secondaryTextFunction]="shortNameFunction"
+  [disableDeselectButton]="disableDeselectButton"
```
HejdaJakub, line 9:
```suggestion
```

libs/perun/components/src/lib/vo-search-select/vo-search-select.component.ts
```diff
@@ -11,6 +11,7 @@ export class VoSearchSelectComponent implements OnChanges {
   @Input() vo: Vo;
   @Input() vos: Vo[];
   @Input() disableAutoSelect = false;
+  @Input() disableDeselectButton = true;
```
HejdaJakub, line 14:
```suggestion
```

HejdaJakub (changes requested):
Many `disableDeselectButton` = true can be removed. I checked and marked all redundant occurrences.

HejdaJakub (changes requested):
Many `disableDeselectButton` = true can be removed. I checked and marked all redundant occurrences.

libs/perun/components/src/lib/audit-log-search-select/audit-log-search-select.component.html
```diff
@@ -8,6 +8,7 @@
     [secondaryTextFunction]="secondaryTextFunction"
     [multiple]="true"
     [disableAutoSelect]="true"
+    [disableDeselectButton]="disableDeselectButton"
```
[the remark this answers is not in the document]
HejdaJakub: Sorry, you're right!

HejdaJakub (approved):

PR #1318, opened 2023-05-16
fix(lib): do not display bottom pagination in tables

michalberky:
Removed div containging the bottom pagination


mattjoke (approved):
Looks good :+1:

HejdaJakub (changes requested):
I think that we can also remove the function `pageChangedBottom()` in the ts file which was related only to this bottom paginator.

HejdaJakub (approved):

PR #1319, opened 2023-05-17
fix(lib): fix group list icons alignment and margin

michalberky:
* Removed horiz_right class and its only application
* Added flex to buttons-container class and removed nowrap and width that I found no longer needed

### DIFF
--- a/libs/perun/components/src/lib/group-menu/group-menu.component.html
+++ b/libs/perun/components/src/lib/group-menu/group-menu.component.html
@@ -1,8 +1,8 @@
-<div *ngIf="displayButtons" class="buttons-container">
+<div *ngIf="displayButtons" class="d-flex justify-content-end">
   <button
     [cdkCopyToClipboard]="group.name"
     [matTooltipPosition]="'above'"
-    class="ms-auto horiz_right"
+    class="ms-auto"
     mat-icon-button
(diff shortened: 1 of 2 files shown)


HejdaJakub (commented):
Good job. :heavy_check_mark: 

You can try even more simplified version of this by using Bootstrap classes. I know, that we probably don't use them consistently across the whole app, but I think that it is good practice to learn.

So you can instead of this `class="buttons-container"` use something like this `class="d-flex justify-content-end"` (https://getbootstrap.com/docs/4.0/utilities/flex/) and it should do the same. So if you prefer this, then you will be able to remove all CSS from the .scss file.

michalberky:
> Good job. heavy_check_mark
> 
> You can try even more simplified version of this by using Bootstrap classes. I know, that we probably don't use them consistently across the whole app, but I think that it is good practice to learn.
> 
> So you can instead of this `class="buttons-container"` use something like this `class="d-flex justify-content-end"` (https://getbootstrap.com/docs/4.0/utilities/flex/) and it should do the same. So if you prefer this, then you will be able to remove all CSS from the .scss file.

Thank you for the tip. Replaced buttons-container class with bootstrap classes according to the suggestion.

HejdaJakub (approved):

PR #1320, opened 2023-05-23
fix(admin): application form items i18n correctly displayed

Kuliak:
* custom form control variables created for each field in each language used
* fixed console error when undefined value was added into sourceAttributes/destinationAttributes array when corresponding attribute had not been found
* inputs for adding new options separated from the rest of form group

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/edit-application-form-item-dialog/edit-application-form-item-dialog.component.ts
+++ b/apps/admin-gui/src/app/shared/components/dialogs/edit-application-form-item-dialog/edit-application-form-item-dialog.component.ts
@@ -14,7 +14,7 @@ import DisabledEnum = ApplicationFormItem.DisabledEnum;
 import HiddenEnum = ApplicationFormItem.HiddenEnum;
 import { ItemType, NO_FORM_ITEM, SelectionItem } from '@perun-web-apps/perun/components';
 import { HtmlEscapeService, StoreService } from '@perun-web-apps/perun/services';
-import { FormControl } from '@angular/forms';
+import { FormControl, FormGroup } from '@angular/forms';
 
 export interface EditApplicationFormItemDialogComponentData {
   theme: string;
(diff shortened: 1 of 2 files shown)


apps/admin-gui/src/app/shared/components/dialogs/edit-application-form-item-dialog/edit-application-form-item-dialog.component.ts
```diff
@@ -41,7 +41,7 @@ export class EditApplicationFormItemDialogComponent implements OnInit {
   hiddenValues: HiddenEnum[] = ['NEVER', 'ALWAYS', 'IF_EMPTY', 'IF_PREFILLED'];
   disabledValues: DisabledEnum[] = ['NEVER', 'ALWAYS', 'IF_EMPTY', 'IF_PREFILLED'];
   possibleDependencyItems: ApplicationFormItem[] = [];
-  htmlInput = new FormControl('', [this.escapeService.htmlContentValidator()]);
+  inputFormGroup: FormGroup = null;
```
bodnara, line 44: This is a typed form group so we should also define the type to get the benefits of type checking and safety. But since the number and name of controls is generated dynamically based on the form item type and supported languages we can't use exact type. The `Record<string, FormControl<string>>` says it will be some structure with key of type `string` and value of type `FormControl<string>`. With this change you can remove any type casting mainly in the `submit()` method.
```suggestion
  inputFormGroup: FormGroup<Record<string, FormControl<string>>> = null;
```

apps/admin-gui/src/app/shared/components/dialogs/edit-application-form-item-dialog/edit-application-form-item-dialog.component.ts
```diff
@@ -249,6 +257,42 @@ export class EditApplicationFormItemDialogComponent implements OnInit {
...
+      // Plain
+      formGroupFields[`${lang}-plain-error-message`] = new FormControl(
+        this.applicationFormItem.i18n[lang].errorMessage,
+        []
+      );
+      formGroupFields[`${lang}-plain-help`] = new FormControl(
+        this.applicationFormItem.i18n[lang].help,
+        []
+      );
+
```
bodnara, line 272: Shouldn't these be only added for non HTML fields? It seems the HTML forms need just `lang-html-label` FormControl, while the plain text forms need `lang-plain-label`, `lang-plain-error` and `lang-plain-help`. Based on the content that is shown on withing the tab. Unless I am missing something.

bodnara (changes requested):
For non HTML field there is some error about undefined attributes, please check this out as well.

bodnara (changes requested):
There are still some undefined properties that cause an error in the dialog.
One other thing is that when a form item that includes the "option" section is used it causes some errors as well. I suspect that it is because the `formGroup` is applied to the whole tab and the option part contains form-fields that are not connected. Try changing the scope of the form group just to the relevant form fields.

bodnara (approved):
Otherwise it works properly.

HejdaJakub (approved):
From my side, it looks good. So after fixing empty `.concat()` it can be merged IMHO.

HejdaJakub (approved):

PR #1323, opened 2023-05-24
fix(admin): disable buttons until the sponsors are not loaded

HejdaJakub:
* Just a quick fix according to the behavior of drop-down menu button.
* All buttons are now disabled during the loading so there is no longer the broken behavior of related dialogs - e.g. dialog for copy members didn't display any sponsors, because they were not fully loaded in the parent component.


bodnara (approved):
Did you consider adding tooltip to these buttons as well ?

HejdaJakub:
> Did you consider adding tooltip to these buttons as well ?

For this moment we decided to let these buttons without tooltips.

PR #1324, opened 2023-05-24
feat: added response parameter to oauth configuration

xflord:


xflord:
After further discussion, this will not be needed ('scopes' parameter should be enough information from the client)

PR #1325, opened 2023-05-26
feat(admin): allow managing embedded groups also in a group form settings

HejdaJakub:
* Managing of embedded groups for embedded form items in group form is now available.
* The backend part was chaged, so new methods for get/add/remove are used.
* Managing in vo form was also refactored to avoid code duplication.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -19,7 +19,7 @@
             "outputPath": "dist/apps/admin-gui",
             "index": "apps/admin-gui/src/index.html",
             "main": "apps/admin-gui/src/main.ts",
-            "polyfills": [ "zone.js" ],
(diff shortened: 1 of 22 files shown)


HejdaJakub:
IMPORTANT: Do not merge before backend implementation https://github.com/CESNET/perun/pull/3964

NOTE: `@nestjs/common` was added to fix openapi-generator-cli behavior after update of this library

bodnara (approved):
**Should be retested once BE is finalized.**

When it comes to add rights:
- Vo - as VO ADMIN it is not necessary to check rights for any groups
- Group - as GROUP ADMIN it is not necessary to check rights for any subgroup

PR #1326, opened 2023-05-26
fix(lib): remove white spaces from the full name pipe

HejdaJakub:
* All parts of full name are joined with space, but if there is some part of the name missing (e.g. middle name), there are more white spaces in the full name.

### DIFF
--- a/libs/perun/pipes/src/lib/user-full-name.pipe.ts
+++ b/libs/perun/pipes/src/lib/user-full-name.pipe.ts
@@ -14,6 +14,6 @@ export class UserFullNamePipe implements PipeTransform {
       value.titleAfter,
     ];
 
-    return nameParts.join(' ');
+    return nameParts.filter((part) => !!part).join(' ');
   }
 }


HejdaJakub:
It looks in the browser view that the full name is joined correctly (according to some browser default behavior), but if you look at the plain HTML, there is visible the incorrect joining of single parts of this name (more spaces instead of middle name and after the name):
Screenshot from 2023-05-26 20-05-14
Screenshot from 2023-05-26 20-04-52
It cause some trouble for the functional tests of our customers.

libs/perun/pipes/src/lib/user-full-name.pipe.ts
```diff
@@ -14,6 +14,6 @@ export class UserFullNamePipe implements PipeTransform {
       value.titleAfter,
     ];
 
-    return nameParts.join(' ');
+    return nameParts.join(' ').trim().replace(/\s+/g, ' ');
```
bodnara, line 17: Would prefer to filter out empty name parts instead of replacing some regex.

bodnara (approved):

PR #1327, opened 2023-05-28
fix(admin): fix login screen service access component

mattjoke:
* service-access login component no longer shows sidebar and header after logout
* added logout notification so it is consistent with regular login

### DIFF
--- a/libs/perun/services/src/lib/init-auth.service.ts
+++ b/libs/perun/services/src/lib/init-auth.service.ts
@@ -10,7 +10,7 @@ import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
 import { Params, Router } from '@angular/router';
 import { OAuthInfoEvent, OAuthService } from 'angular-oauth2-oidc';
 import { filter } from 'rxjs/operators';
-import { firstValueFrom } from 'rxjs';
+import { firstValueFrom, timer } from 'rxjs';
 import { MfaHandlerService } from './mfa-handler.service';
 
 @Injectable({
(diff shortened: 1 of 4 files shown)


libs/perun/services/src/lib/init-auth.service.ts
```diff
@@ -190,6 +190,18 @@ export class InitAuthService {
...
+  invalidateServiceAcess(): void {
+    // Has to be promise, bc. of NG0100.
+    Promise.resolve()
+      .then(() => {
+        this.serviceAccess = false;
+        this.setLoginScreen(true);
+      })
+      .catch(() => {
+        // This will never happen
+      });
```
HejdaJakub, line 202: Probably not the best solution, but it works, so at least I think that this can lead to better readability.
```suggestion
    // Has to be async, bc. of ExpressionChangedAfterItHasBeenCheckedError
    timer(0).subscribe(() => {
      this.serviceAccess = false;
      this.setLoginScreen(true);
    });
```

libs/perun/services/src/lib/init-auth.service.ts
```diff
@@ -190,6 +190,18 @@ export class InitAuthService {
     }
   }
 
+  invalidateServiceAcess(): void {
```
HejdaJakub, line 193:
```suggestion
  invalidateServiceAccess(): void {
```

HejdaJakub (changes requested):
Dunno why, but when I log out and try to log in again, it is not possible to log in without refresh. :thinking:

Kuliak (commented):
Code looks fine to me, however, it needs to be tested.

mattjoke:
After the logout, user can now successfully log back in, not the prettiest solution, but it seems to work

HejdaJakub (approved):
As you said, this is probably not the best solution, but it works and I don't have a better one in my mind now, so you have my approve

sarkapalkovicova (approved):

PR #1330, opened 2023-06-07
fix: do not show sesssion expiration dialog during mfa

HejdaJakub:
* There was a bug that during step-up authentication process the session expiration dialog was shown.
* I believe that this one added condition should solve this problem without any other side effects.


bodnara (approved):
Tested, worked.

mattjoke (approved):
Looks good👍

PR #1331, opened 2023-06-07
fix: Unify member status and expiration change

bodnara:
* In all cases expiration-select component is used
* Expiration-select now allows to use Vo,Group rules for extention
* Changing expiration now always allows to choose sensible date
* Changing expiration alone can never lead to status change
* Refactored connected components, mostly removed unnecessary code


sarkapalkovicova (approved):
It seems to work correctly. Good job.

mattjoke (approved):
Looks good 👍

bodnara:
Fixed, lower bound when changing expiration for sponsorship.

sarkapalkovicova (approved):

PR #1332, opened 2023-06-08
fix: correctly handle MFA timeout

HejdaJakub:
* If the MfaTimeoutError is thrown by backend, then we should force new authentication included first factor and second factor


HejdaJakub:
Do NOT merge before: https://github.com/CESNET/perun/pull/3975

mattjoke (approved):
The code looks good 👍

bodnara (approved):

HejdaJakub:
Rebased and also changed `sessionStorage` to `localStorage` due to correct behavior in all browsers.

PR #1334, opened 2023-06-09
fix(admin): disable auto select of vo

sarkapalkovicova:
Auto select of VO in create-resource-dialog was disabled.


mattjoke (changes requested):
This solution seems right, but I would put a * or something to indicate, that the user is required to choose a VO. Also, this seems unrelated to the task, but maybe there should be a title 'Create a Resource' so it is consistent with other modals.

mattjoke (approved):
Looks good no 👍

HejdaJakub (approved):

PR #1338, opened 2023-06-22
fix(lib): handle MfaRoleTimeoutException

HejdaJakub:
* We want to catch this specific exception and according to it hide cancel button in step-up dialog and also force authentication with both factors.


bodnara (approved):

HejdaJakub:
Merge after this: https://github.com/CESNET/perun/pull/3984

PR #1343, opened 2023-06-23
fix(admin): notification display time

sarkapalkovicova:
Extended and unified time of displaying notification.

### DIFF
--- a/libs/perun/services/src/lib/notificator.service.ts
+++ b/libs/perun/services/src/lib/notificator.service.ts
@@ -10,8 +10,8 @@ export class NotificatorService {
   @Output() addNotification: EventEmitter<NotificationData> = new EventEmitter<NotificationData>();
   defaultAction: string;
   defaultRpcMessage: string;
-  defaultErrorDelayMs = 5_000;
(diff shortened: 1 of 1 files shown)


libs/perun/services/src/lib/notificator.service.ts
```diff
@@ -10,8 +10,8 @@ export class NotificatorService {
   @Output() addNotification: EventEmitter<NotificationData> = new EventEmitter<NotificationData>();
   defaultAction: string;
   defaultRpcMessage: string;
-  defaultErrorDelayMs = 5_000;
-  defaultSuccessDelayMs = 3_000;
+  defaultErrorDelayMs = 7_000;
+  defaultSuccessDelayMs = 7_000;
```
bodnara, line 14: After testing it we could put it even higher, users seem to struggle with it, so 2sec increase might not be sufficient.
```suggestion
  defaultErrorDelayMs = 10_000;
  defaultSuccessDelayMs = 10_000;
```

bodnara (approved):

PR #1353, opened 2023-06-23
feat: replace mat-flat-button by mat-stroked-button 

xkostka2:
* all occurrences of mat-flat-button have been replaced by mat-stroked-button if button does not have background


bodnara (changes requested):
Missing `NewVersionDialogComponent` and `ShowNotificationHistoryDialogComponent`.

Kuliak (commented):
What about file **`tag-section.component.html`** ?

bodnara (approved):

mattjoke (approved):
Just need to rebase, otherwise it's good

PR #1356, opened 2023-07-03
fix: don't overlap text in delete dialog

xflord:
* changed delete-entity-dialog margins to fix text overlap


sarkapalkovicova (commented):
I checked it, and looks good.

mattjoke (approved):
Looks good :+1:

PR #1358, opened 2023-07-04
fix(admin): responsive mfa window

sarkapalkovicova:
Step-up mfa window is responsive.

### DIFF
--- a/libs/perun/services/src/lib/mfa-handler.service.ts
+++ b/libs/perun/services/src/lib/mfa-handler.service.ts
@@ -155,8 +155,8 @@ export class MfaHandlerService {
 
   private setupMfaWindow(): Window {
     const url = location.pathname + location.search;
-    const width = 600;
-    const height = 600;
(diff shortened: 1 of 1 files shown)


mattjoke (approved):
Looks good :+1:

libs/perun/services/src/lib/mfa-handler.service.ts
```diff
@@ -155,8 +155,8 @@ export class MfaHandlerService {
 
   private setupMfaWindow(): Window {
     const url = location.pathname + location.search;
-    const width = 600;
-    const height = 600;
+    const width = window.outerWidth * 0.6;
```
xflord, line 158: This sets the size of the new window based on the size of the existing window, but wouldn't setting it based on the monitor size make more sense? An already small windowed window opening an even smaller window doesn't seem practical.
I think this can be achieved with `window.screen.width` and `window.screen.height`.

xflord (approved):

PR #1359, opened 2023-07-06
fix(admin): refresh group managers

sarkapalkovicova:
Correctly refresh managers when coming back from other group detail.


xflord (approved):
works

PR #1363, opened 2023-07-18
build: migrate to Angular 16

HejdaJakub:
* The whole repo was migrated from Angular 15 to Angular 16 via nx migration tool.
* Several libraries were manually upgraded too (zone.js, typescript, jest, jest-environment-jsdom, ngx-mat-select-search, ngx-charts, ngxgraph, openapi-generator-cli).
* Nx configuration via several project.json files was applied instead of previous single Angular CLI config file angular.json.
* The decorate-angular-cli script was fixed and in the next Angular version should be removed.
* Ngcc is deprecated, so the part of postinstall command was auto-removed.
* Rescoped packages from @nrwl to @nx.
* All peer dependency conflicts were removed, so the --legacy-peer-deps flag was removed.


.github/workflows/check.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: lts/*
+      - name: Install dependencies
+        run: npm ci --legacy-peer-deps
```
bodnara, line 18:
```suggestion
        run: npm ci
```

mattjoke (approved):
Looks good :+1:

PR #1368, opened 2023-07-20
fix(admin): icon button tooltip

sarkapalkovicova:
Added tooltip for every mat-icon-button.

### DIFF
--- a/apps/admin-gui/src/app/shared/components/sponsored-members-list/sponsored-members-list.component.html
+++ b/apps/admin-gui/src/app/shared/components/sponsored-members-list/sponsored-members-list.component.html
@@ -96,13 +96,12 @@
             (mouseleave)="disableRouting = routingStrategy"
             aria-label="sponsors of member"
             class="dark-hover-list-item"
+            [matTooltipPosition]="'above'"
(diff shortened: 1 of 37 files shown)


apps/admin-gui/src/app/shared/components/sponsored-members-list/sponsored-members-list.component.html
michalberky, line 1: I think it would be best to also move the tooltip from lines 106 and 107 of this file inside of the <button> tag starting on line 108. This would make the tooltip appear above the button where it should be.

HejdaJakub (changes requested):
I think that the last place where the tooltip is missing is:
`perun-web-apps/apps/consolidator/src/app/components/list-of-identities` for the `<mat-icon>clear</mat-icon>`.
There is already a tooltip for the disabled icon, but there is no tooltip for the enabled one, so can you please add also a tooltip for the enabled icon?

mattjoke (approved):
Looks good :+1:

michalberky (approved):

HejdaJakub (approved):

PR #1369, opened 2023-07-21
chore: change commitlint rules

HejdaJakub:
* The body-max-line-length rule was raised from the default 100 characters to 200 characters.
* Branches from the dependabot (with usual prefix dependabot/npm_and_yarn/) are now excluded from the commlint check.

### DIFF
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -18,6 +18,7 @@
     }
   },
   "rules": {
-    "scope-enum": [2, "always", ["admin", "profile", "pwdreset", "consolidator", "linker", "publications", "openapi", "lib", "npm", "deps", "deps-dev"]]
+    "scope-enum": [2, "always", ["admin", "profile", "pwdreset", "consolidator", "linker", "publications", "openapi", "lib", "npm", "deps", "deps-dev"]],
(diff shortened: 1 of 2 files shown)


.commitlintrc.json
```diff
@@ -18,6 +18,7 @@
     }
   },
   "rules": {
-    "scope-enum": [2, "always", ["admin", "profile", "pwdreset", "consolidator", "linker", "publications", "openapi", "lib", "npm", "deps", "deps-dev"]]
+    "scope-enum": [2, "always", ["admin", "profile", "pwdreset", "consolidator", "linker", "publications", "openapi", "lib", "npm", "deps", "deps-dev"]],
+    "body-max-line-length": [0, "always", 100]
```
bodnara, line 22: Forgot the change itself ?
```suggestion
    "body-max-line-length": [0, "always", 150]
```

HejdaJakub: Sorry, this was just testing of these rules. I changed it to 200 characters and I changed the first number to `2` (meaning of this number is the level: `0 disables the rule, 1 will be considered a warning, 2 as an error`)

bodnara (approved):

PR #1370, opened 2023-07-21
feat(admin): Displaying children of items on entity detail

bodnara:
* On overview tiles with children were replaced by expandable section that contains the children instead of new page
* In side-menu items with children will expand on click instead of navigating to new page
* Refactored side-menu and overview pages to reflect current standards
* Removed SideMenuRootItem component as it was just a duplicate of SideMenuItem


mattjoke (approved):
Looks good :+1: 
Please just fix your commit message

apps/admin-gui/src/app/users/pages/user-detail-page/user-overview/user-overview.component.ts
```diff
@@ -139,46 +139,55 @@ export class UserOverviewComponent implements OnInit {
...
+      this.items.push(
+        {
+          cssIcon: 'perun-settings2',
+          url: '/myProfile/settings/passwordReset',
+          label: 'MENU_ITEMS.USER.PASSWORD_RESET',
+          style: 'user-btn',
+        },
+        {
+          cssIcon: 'perun-settings1',
+          url: '/myProfile/settings/guiConfig',
```
HejdaJakub, line 186: Maybe we can throw away `/settings` from this URL when these items are now on the same level?

bodnara: This change was reverted. Originally I thought it would not be possible to have not entity related items expandable, but it was fixed by merging the root and "normal" side menu item component.

bodnara:
> * We used to have no border around each main item in the side menu and we also used to have a background at the "active" main item (in both pictures the `Facility management` is active).

This should now be fixed.


>  * Sections under visualiser don't work (missing `/visualizer` in the URL).

Fixed the link on overview.


>  * When I go from group to subgroup, I don't see the label and link: "This group is a subgroup of [Kuba test group](https://gui-dev.perun-aai.org/organizations/3983/groups/13994)"
>  * When I go from VO directly to subgroup and click on the link in "This group is a subgroup of [Kuba test group](http://localhost:4200/organizations/3983/groups/13994)", this link is still displayed also after the redirect to the parent group.

The check if  correct Group is loaded was changed to mimic the behavior of Vo. Since the information about parent group was only present on overview I decided to move this to the detail of Group. Now the information is present for all pages not just on overview.

apps/admin-gui/src/app/shared/side-menu/side-menu-item.service.ts
```diff
@@ -237,6 +217,24 @@ export class SideMenuItemService {
...
+          url: ['/admin/visualizer'],
+          activatedRegex: '^/admin/visualizer$',
+          children: [
+            {
+              label: 'MENU_ITEMS.VISUALIZER.ATTR_DEPENDENCIES',
+              url: ['/admin/visualizer/attrDependencies'],
+              activatedRegex: '^/admin/visualizer/attrDependencies',
+            },
+            {
+              label: this.translate.instant('MENU_ITEMS.VISUALIZER.USER_DESTINATION') as string,
```
HejdaJakub, line 231: Can you please replace it with `PerunTranslateService` and remove `as string`.

HejdaJakub (changes requested):
- If the `Home` is active, the background in the side menu is missing.
- The background is missing also for `My Profile` and `Perun Admin`, but I found out that it was broken also on the `master` branch. :eyes:.
- For the `My Profile` and `Perun Admin` is still visible the top border.
image
- For Access management and Facilities is visible the bottom border.
image

QUESTION:
- Do we want to have clickable just the button or also the text beside it?
image

apps/admin-gui/src/app/shared/side-menu/side-menu-item.service.ts
bodnara: Removed redundant translation.

HejdaJakub (approved):
I think that now it looks fine. Good job :fire:

xflord (approved):

PR #1377, opened 2023-08-01
fix(profile): authentication management

sarkapalkovicova:
Authentication management is separated into its own section.

DEPLOYMENT NOTE: In defaultConfig.json for user-profile, 'authentication', 'mfa', 'anti_phishing', and 'accountActivation' values were added into displayed_tabs.


apps/user-profile/src/app/pages/authentication-page/authentication-overview/authentication-overview.component.ts
```diff
@@ -0,0 +1,59 @@
+import { Component, OnInit, ViewChild } from '@angular/core';
+import { MatSlideToggle } from '@angular/material/slide-toggle';
+import { PerunTranslateService, StoreService } from '@perun-web-apps/perun/services';
+
+interface TabItem {
+  icon: string;
+  url: string;
+  label: string;
+  tabName: string;
+}
```
bodnara, line 10: Same interface is defined in for settings overview, please define it in one place. Probably perun-models lib.

bodnara (changes requested):
The settings overview has nice indication when item is hovered, please add this also for authentication overview. Also not sure if the "alternative passwords" page shouldn't be part of authentication ? It was not part of the assignment but maybe makes more sense there.
image

I think the authentications page should not have any content, but serve just as a overview same as settings. The SSH keys and Change password should imho each have separate page that is in the menu.

bodnara (approved):

xflord (approved):
good job

PR #1378, opened 2023-08-01
build: migrate to Angular 16

HejdaJakub:
* The whole repo was migrated from Angular 15 to Angular 16 via nx migration tool.
* Several libraries were manually upgraded too (zone.js, typescript, jest, jest-environment-jsdom, ngx-mat-select-search, ngx-charts, ngxgraph, openapi-generator-cli).
* Nx configuration via several project.json files was applied instead of previous single Angular CLI config file angular.json.
* The decorate-angular-cli script was fixed and in the next Angular version should be removed.
* Ngcc is deprecated, so the part of postinstall command was auto-removed.
* Rescoped packages from @nrwl to @nx.
* All peer dependency conflicts were removed, so the --legacy-peer-deps flag was removed.


HejdaJakub:
This commit was reviewed during the PR to the separate upstream branch `angular16`. This PR only delivers changes to the `master` branch.

PR #1380, opened 2023-08-01
fix(admin): fixed adding notification tags

xflord:


sarkapalkovicova (approved):
It works correctly.

apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.ts
```diff
@@ -177,31 +177,14 @@ export class AddEditNotificationDialogComponent implements OnInit {
...
-          tag +
-          this.applicationMail.message[language].subject.substring(position);
-      }
+    if (format === 'plain_text') {
+      format = 'plain';
     }
+    const form = this.inputFormGroup.get(
+      `${language}-${format}-${this.isTextFocused ? 'text' : 'subject'}`
+    );
+    const curValue = String(form.value);
```
bodnara, line 186: Tags are now added properly, well done. I would just like to see some technical changes.

The formats are defined inside this class and serve no other purpose it seem. So we could redefine the "plain_text" option to "plain" so the conversion is not needed.

Also please set the type of `inputFormGroup` to `FormGroup<Record<string, FormControl<string>>>` (form group of objects that have key of type string and value of type `FormControl<string>`) so we have type checking. This will require same type change to the variable `formGroupFields` in `ngOnInit`. And the conversion of value to string will not be needed anymore.

bodnara (changes requested):

apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.ts
xflord: Adjusted the typing, changed `plain_text` to `plain` analogously also in the edit-notification component + removed deprecated syntax in the add-edit-notification component

bodnara (approved):

PR #1381, opened 2023-08-02
feat(admin): attribute operation can be globally critical

HejdaJakub:
* In Perun admin Attribute definition section we can now set attribute as globally critical (critical for all objects).


apps/admin-gui/src/app/shared/components/dialogs/create-attribute-definition-dialog/create-attribute-definition-dialog.component.html
```diff
@@ -82,7 +82,9 @@ <h1 mat-dialog-title>{{'DIALOGS.CREATE_ATTRIBUTE_DEFINITION.TITLE' | translate}}
 
       <perun-web-apps-attribute-critical-operations-toggles
         (readOperationChanged)="finalReadOperations=$event"
-        (writeOperationChanged)="finalWriteOperations=$event">
+        (readGlobalChanged)="finalReadGlobal=$event"
+        (writeOperationChanged)="finalWriteOperations=$event"
```
xflord, line 86: We need to change global criticality to false when changing this to false as well. Otherwise exception is thrown (see video)
[Screencast from 3.8.2023 13:26:21.webm](https://github.com/CESNET/perun-web-apps/assets/97669007/f962fe53-cf3e-4090-97f1-ccca5d7708ca)

libs/perun/components/src/lib/attribute-critical-operations-toggles/attribute-critical-operations-toggles.component.html
```diff
@@ -1,19 +1,35 @@
 <mat-slide-toggle
-  class="toggle-font mb-25"
+  [class]="readOperation ? 'toggle-font' : 'toggle-font mb-25'"
```
bodnara, line 2: You can enable specific class (perhaps multiple not sure) like so. 
```suggestion
  class="toggle-font"
  [class.mb-25]="!readOperation"
```

bodnara (commented):
Other than problem mentioned by @xflord it looks good.

libs/perun/components/src/lib/attribute-critical-operations-toggles/attribute-critical-operations-toggles.component.ts
```diff
@@ -7,7 +7,26 @@ import { Component, EventEmitter, Input, Output } from '@angular/core';
...
+    globalChanged: EventEmitter<boolean>,
+    global: string
+  ): void {
+    operationChanged.emit(!operation);
+
+    // if the operation is changing from true to false, then we want to change also global criticality to false
+    if (operation) {
+      this[global] = false;
+      globalChanged.emit(false);
+    }
```
bodnara, line 30: I understand you wanted to keeps things DRY but I am quite against `this[global]` for multiple reasons. Firstly as a caller of this method you don't get any help from Typescript about what values of `global` are valid and need to check manually. This also means there is no warning that you are assigning to non existent property or property of different type. Lastly any change to variable names requires change to all the method calls again with no warning.
Right now I see two approaches:
- Better to repeat yourself than shoot yourself in the foot. Just create two separate methods for write and read. Straight forward option.
- Instead of having output (maybe even input) for each property have one output with all values wrapped in an object. May be a little bit less cluttered. This would however, mean that every time all values need to be reassigned (object created, ...) which is wasteful. Although here it should not be a problem.

bodnara (approved):

apps/admin-gui/src/app/shared/components/dialogs/edit-attribute-definition-dialog/edit-attribute-definition-dialog.component.ts
```diff
@@ -105,7 +115,10 @@ export class EditAttributeDefinitionDialogComponent implements OnInit {
           );
           this.dialogRef.close(true);
         },
-        error: () => (this.loading = false),
+        error: (err) => {
+          this.loading = false;
+          console.error(err);
```
xflord, line 120: Leftover from testing, no?

xflord (approved):
One detail but works fine

apps/admin-gui/src/app/shared/components/dialogs/edit-attribute-definition-dialog/edit-attribute-definition-dialog.component.ts
HejdaJakub: I wanted to leave it here for future better debugging of these several subscriptions and potentially also for E2E tests checking errors in the console. :raised_hands:

PR #1382, opened 2023-08-04
fix(admin): notification values not saved

mattjoke:
* added binding to reflect change when user inputs data in notification

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.ts
+++ b/apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.ts
@@ -109,6 +109,7 @@ export class AddEditNotificationDialogComponent implements OnInit {
       return;
     }
     this.loading = true;
+    this.validateNotification();
     if (this.data.groupId) {
       this.registrarService
         .addApplicationMailForGroup({
@@ -138,24 +139,7 @@ export class AddEditNotificationDialogComponent implements OnInit {
 
   save(): void {
     this.loading = true;
-    // Validate notification
-    for (const lang of this.languages) {
-      let escaped = this.inputEscape.escapeDangerousHtml(
-        this.inputFormGroup.get(`${lang}-html-subject`).value
-      );
-      this.applicationMail.htmlMessage[lang].subject = escaped.escapedHtml;
-      escaped = this.inputEscape.escapeDangerousHtml(
-        this.inputFormGroup.get(`${lang}-html-text`).value
-      );
-      this.applicationMail.htmlMessage[lang].text = escaped.escapedHtml;
-
-      // Update application with content from FormControl
-      this.applicationMail.message[lang].subject = this.inputFormGroup.get(
-        `${lang}-plain-subject`
-      ).value;
-      this.applicationMail.message[lang].text = this.inputFormGroup.get(`${lang}-plain-text`).value;
-    }
-
+    this.validateNotification();
     this.registrarService.updateApplicationMail({ mail: this.applicationMail }).subscribe({
       next: () => {
         this.dialogRef.close(true);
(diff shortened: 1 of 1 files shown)


sarkapalkovicova (approved):
Works.

bodnara (changes requested):
This does not seem to fix the issue. I think the problem is that the value is not updated/validated before creation. It is only done when saved, which is probably why editing works fine.

xflord:
> This does not seem to fix the issue. I think the problem is that the value is not updated/validated before creation. It is only done when saved, which is probably why editing works fine.

yep, `create()` method needs to be updated similarly to `save()`

mattjoke:
> > This does not seem to fix the issue. I think the problem is that the value is not updated/validated before creation. It is only done when saved, which is probably why editing works fine.
> 
> yep, `create()` method needs to be updated similarly to `save()`

HTML escaping is done on the backend side. The code you are referring to, is just a unused part of the code.

xflord:
> > > This does not seem to fix the issue. I think the problem is that the value is not updated/validated before creation. It is only done when saved, which is probably why editing works fine.
> > 
> > 
> > yep, `create()` method needs to be updated similarly to `save()`
> 
> HTML escaping is done on the backend side. The code you are referring to, is just a unused part of the code used before the merge. And therefore the solution in this PR does not correlate to the problem.

I see, is it safe to remove the code then? Also, since the backend validation is back up again, do we plan to uncomment/fix frontend validation?

mattjoke:
> > > > This does not seem to fix the issue. I think the problem is that the value is not updated/validated before creation. It is only done when saved, which is probably why editing works fine.
> > > 
> > > 
> > > yep, `create()` method needs to be updated similarly to `save()`
> > 
> > 
> > HTML escaping is done on the backend side. The code you are referring to, is just a unused part of the code used before the merge. And therefore the solution in this PR does not correlate to the problem.
> 
> I see, is it safe to remove the code then? Also, since the backend validation is back up again, do we plan to uncomment/fix frontend validation?

This may be needed for a more thorough discussion, as this problem/solution is rather complex. Nevertheless, I think the notifications now correctly save input data.

apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.ts
```diff
@@ -109,6 +109,24 @@ export class AddEditNotificationDialogComponent implements OnInit {
...
+        this.inputFormGroup.get(`${lang}-html-text`).value
+      );
+      this.applicationMail.htmlMessage[lang].text = escaped.escapedHtml;
+
+      // Update application with content from FormControl
+      this.applicationMail.message[lang].subject = this.inputFormGroup.get(
+        `${lang}-plain-subject`
+      ).value;
+      this.applicationMail.message[lang].text = this.inputFormGroup.get(`${lang}-plain-text`).value;
+    }
```
bodnara, line 128: Please extract this logic so it's not duplicated for both save and create.

apps/admin-gui/src/app/shared/components/dialogs/add-edit-notification-dialog/add-edit-notification-dialog.component.html
```diff
@@ -149,11 +149,15 @@ <h1 mat-dialog-title>{{'DIALOGS.NOTIFICATIONS_ADD_EDIT_MAIL.TITLE_EDIT' | transl
                             *ngIf="format === 'plain'"
                             formControlName="{{lang}}-plain-subject"
                             (focus)="isTextFocused = false"
+                            [value]="this.applicationMail.message[lang].subject"
+                            (input)="this.applicationMail.message[lang].subject = inputFormGroup.controls[lang + '-plain-subject'].value"
```
bodnara, line 153: I believe this should not be needed anymore. Also the way I understand it, is that the form control serves exactly this purpose of input-value connection (with some additional features).

mattjoke: Yes. Once I removed the duplicities, the problem no longer occurs.

bodnara (approved):

xflord (approved):
gj

PR #1386, opened 2023-08-09
fix(admin): correct button layout in edit attribute definition dialog

kofzera:
Cancel button is now on the left in the edit attribute definition dialog.


xflord (approved):
gj

PR #1387, opened 2023-08-09
feat(lib): password reset on profile opens modal window

michalberky:
Password reset, when clicked on in profile, now opens a modal window instead of redirecting to separate application.
* Created a new password reset dialog
* Modified password-reset component to use this new dialog.

### DIFF
--- a/libs/perun/components/src/lib/password-reset/password-reset.component.ts
+++ b/libs/perun/components/src/lib/password-reset/password-reset.component.ts
@@ -1,16 +1,12 @@
 import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
 import { Attribute, AttributesManagerService } from '@perun-web-apps/perun/openapi';
-import {
-  EntityStorageService,
-  OtherApplicationsService,
-  StoreService,
-} from '@perun-web-apps/perun/services';
+import { EntityStorageService, StoreService } from '@perun-web-apps/perun/services';
 import { MatTableDataSource } from '@angular/material/table';
 import { MatDialog } from '@angular/material/dialog';
 import { getDefaultDialogConfig } from '@perun-web-apps/perun/utils';
 import { ChangePasswordDialogComponent } from '@perun-web-apps/perun/dialogs';
+import { PasswordResetDialogComponent } from '@perun-web-apps/perun/dialogs';
 import { ActivatedRoute, Router } from '@angular/router';
-import { AppType } from '@perun-web-apps/perun/models';
 
 @Component({
   selector: 'perun-web-apps-password-reset',
@@ -32,7 +28,6 @@ export class PasswordResetComponent implements OnInit {
     private dialog: MatDialog,
     private route: ActivatedRoute,
     private router: Router,
-    private otherApplicationService: OtherApplicationsService,
     private entityStorageService: EntityStorageService
   ) {}
 
@@ -69,11 +64,16 @@ export class PasswordResetComponent implements OnInit {
     });
   }
 
-  resetPassword(login: string): void {
-    window.open(
-      this.otherApplicationService.getUrlForOtherApplication(AppType.PwdReset, login),
-      '_blank'
-    );
+  resetPassword(login: Attribute): void {
+    const config = getDefaultDialogConfig();
+    config.width = '600px';
+    config.data = {
+      mode: 'reset',
+      login: String(login.value),
+      namespace: login.friendlyName.split(':')[1],
+    };
+
+    this.dialog.open(PasswordResetDialogComponent, config);
   }
 
   changePassword(login: Attribute): void {
--- a/libs/perun/dialogs/src/lib/password-reset-dialog/password-reset-dialog.component.html
+++ b/libs/perun/dialogs/src/lib/password-reset-dialog/password-reset-dialog.component.html
@@ -1,0 +1,43 @@
+<ng-template #spinner>
+  <perun-web-apps-loading-dialog></perun-web-apps-loading-dialog>
+</ng-template>
+<div class="user-theme position-relative">
+  <div *perunWebAppsLoader="loading; indicator: spinner">
+    <h1 mat-dialog-title>
+      {{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.TITLE' | translate}}
+    </h1>
+    <div class="dialog-container" mat-dialog-content>
+      <p *ngIf="data.login" class="subtitle">
+        <em>
+          {{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.LOGIN' | translate}}:
+          {{data.login}}
+        </em>
+      </p>
+      <p *ngIf="data.namespace" class="subtitle">
+        <em>
+          {{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.NAMESPACE' | translate}}:
+          {{data.namespace}}
+        </em>
+      </p>
+      <perun-web-apps-password-form
+        [formGroup]="newPasswdForm"
+        [namespace]="data.namespace"
+        [language]="language">
+      </perun-web-apps-password-form>
+
+      <div mat-dialog-actions class="ms-auto justify-content-end">
+        <button (click)="close()" mat-stroked-button>
+          {{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.CANCEL' | translate}}
+        </button>
+        <button
+          (click)="onSubmit()"
+          [disabled]="newPasswdForm.invalid || newPasswdForm.pending || loading"
+          class="ms-2"
+          color="accent"
+          mat-flat-button>
+          {{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.CHANGE' | translate}}
+        </button>
+      </div>
+    </div>
+  </div>
+</div>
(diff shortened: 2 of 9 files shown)


libs/perun/components/src/lib/password-reset/password-reset.component.ts
```diff
@@ -69,11 +64,39 @@ export class PasswordResetComponent implements OnInit {
...
-    );
+  resetPassword(login: Attribute): void {
+    if (!this.authenticationPage) {
+      void this.router.navigate([], {
+        queryParams: {
+          namespace: login.friendlyNameParameter,
+        },
+        queryParamsHandling: 'merge',
+      });
+    }
```
bodnara, line 75: I believe the added query parameter is not used for anything, am I missing something ? If it is in fact not needed please remove the navigation here and after dialog is closed section.

libs/perun/dialogs/src/lib/password-reset-dialog/password-reset-dialog.component.html
```diff
@@ -0,0 +1,60 @@
...
+  <perun-web-apps-loading-dialog></perun-web-apps-loading-dialog>
+</ng-template>
+<div class="user-theme position-relative">
+  <div class="title-with-icon">
+    <h1 mat-dialog-title>
+      {{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.TITLE' | translate}}
+    </h1>
+  </div>
+  <div class="dialog-container" mat-dialog-content>
+    <p *ngIf="getLogin()" class="subtitle">
```
bodnara, line 11: Please don't use functions to preform data calculations, because of how change detection works, this will be recalculated thousand of times. It you need to transform data somehow use a pipe, otherwise just use a variable. The information is passes to the dialog and can be directly used here.

libs/perun/dialogs/src/lib/password-reset-dialog/password-reset-dialog.component.html
```diff
@@ -0,0 +1,60 @@
...
+        {{successMsg}}
+      </perun-web-apps-alert>
+    </div>
+
+    <div *ngIf="error && !loading" class="mt-5">
+      <perun-web-apps-alert alert_type="error">
+        {{errorMsg}}
+      </perun-web-apps-alert>
+    </div>
+  </div>
```
bodnara, line 41: There is no need to show any success / error message inside the dialog. Show a notification on success, error will be shown automatically if it occurs.

libs/perun/dialogs/src/lib/password-reset-dialog/password-reset-dialog.component.html
```diff
@@ -0,0 +1,60 @@
...
+
+  <div mat-dialog-actions>
+    <div class="ms-auto">
```
bodnara, line 44: The inner div is unnecessary, put `ms-auto` on the left most button.

libs/perun/dialogs/src/lib/password-reset-dialog/password-reset-dialog.component.ts
```diff
@@ -0,0 +1,109 @@
...
+    private storeService: StoreService,
+    private translate: TranslateService,
+    private apiRequestConfiguration: ApiRequestConfigurationService,
+    private usersService: UsersManagerService,
+    private formBuilder: FormBuilder,
+    private notificator: NotificatorService
+  ) {
+    translate
+      .get('SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.SUCCESS')
+      .subscribe((m: string) => (this.successMsg = m));
```
bodnara, line 50: Please use `PerunTranslateService` instead. Also we usually don't save the translation since it's not reused. You can use the `instant()` method of the translator.

libs/perun/dialogs/src/lib/password-reset-dialog/password-reset-dialog.component.html
```diff
@@ -0,0 +1,60 @@
+<ng-template #spinner>
+  <perun-web-apps-loading-dialog></perun-web-apps-loading-dialog>
+</ng-template>
+<div class="user-theme position-relative">
+  <div class="title-with-icon">
+    <h1 mat-dialog-title>
+      {{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.TITLE' | translate}}
+    </h1>
+  </div>
```
bodnara, line 9: There is no icon in the dialog title.
```suggestion
  <h1 mat-dialog-title>
    {{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.TITLE' | translate}}
  </h1>
```

bodnara (changes requested):
Some small changes are needed, but otherwise good job.

mattjoke (commented):
Looks good :+1:
However, one thing that we could discuss is how should the dialog handle longer text?
For example, if I put in a shorter password and the validation kicks in, the whole content is now a big scrollable list. This does not happen when we have the Change password dialog opened. I would say, that it shouldn't be scrollable. What do you think @bodnara ?
image
image

michalberky:
@mattjoke 
> Looks good +1 However, one thing that we could discuss is how should the dialog handle longer text? For example, if I put in a shorter password and the validation kicks in, the whole content is now a big scrollable list. This does not happen when we have the Change password dialog opened. I would say, that it shouldn't be scrollable. What do you think @bodnara ? image image

I think that both Reset password and Change password dialogs work the same in this regard. The only difference is resolution/scale. This is Change password with browser zoom of 150 %:
image

On my side, with default zoom, neither dialog transforms into scrollable list even with all validations visible.

So, if I haven't overlooked anything, the behavior is consistent between both dialogs, although whether it is also correct could be discussed.

libs/perun/dialogs/src/lib/password-reset-dialog/password-reset-dialog.component.html
```diff
@@ -0,0 +1,45 @@
...
+        >{{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.LOGIN' | translate}}:
+        {{data.login}}</em
+      >
+    </p>
+    <p *ngIf="data.namespace" class="subtitle">
+      <em
+        >{{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.NAMESPACE' | translate}}:
+        {{data.namespace}}</em
+      >
+    </p>
```
bodnara, line 20: Formated. Sometimes Prettier needs small help to format things nicely.
```suggestion
    <p *ngIf="data.login" class="subtitle">
      <em>
        {{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.LOGIN' | translate}}:
        {{data.login}}
      </em>
    </p>
    <p *ngIf="data.namespace" class="subtitle">
      <em>
        {{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.NAMESPACE' | translate}}:
        {{data.namespace}}
      </em>
    </p>
```

libs/perun/dialogs/src/lib/password-reset-dialog/password-reset-dialog.component.html
```diff
@@ -0,0 +1,45 @@
...
+    <perun-web-apps-password-form
+      class="ml-3 me-3"
```
bodnara, line 22: Seems unnecessary.
```suggestion
```

libs/perun/dialogs/src/lib/password-reset-dialog/password-reset-dialog.component.html
```diff
@@ -0,0 +1,45 @@
...
+        [disabled]="newPasswdForm.invalid || newPasswdForm.pending || loading"
+        class="ms-2"
+        color="accent"
+        mat-flat-button>
+        {{'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.CHANGE' | translate}}
+      </button>
+    </div>
+  </div>
+
+  <mat-spinner *ngIf="loading" class="ms-auto me-auto"></mat-spinner>
```
bodnara, line 44: We no longer use loading indication this way. Rather use a "non blocking" variant. You have the template set up at the start of the file it just needs to be used properly. Check out some other dialog (e.g. create vo) to see how it's done.

libs/perun/dialogs/src/lib/password-reset-dialog/password-reset-dialog.component.ts
```diff
@@ -0,0 +1,99 @@
...
+
+@Component({
+  selector: 'perun-web-apps-password-reset-dialog',
+  templateUrl: './password-reset-dialog.component.html',
+  styleUrls: ['./password-reset-dialog.component.scss'],
+})
+export class PasswordResetDialogComponent implements OnInit {
+  loading = false;
+  success = false;
+  error = false;
```
bodnara, line 28: Remove unused variables.
```suggestion
```

libs/perun/dialogs/src/lib/password-reset-dialog/password-reset-dialog.component.ts
```diff
@@ -0,0 +1,99 @@
...
+        namespace: this.data.namespace,
+        newPassword: this.newPasswdForm.value.passwordCtrl,
+      })
+      .subscribe(() => {
+        this.success = true;
+        this.notificator.showSuccess(
+          this.translate.instant(
+            'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.SUCCESS'
+          ) as string
+        );
```
bodnara, line 84: This method takes care of the translation itself and is the preferred way to show notifications.
```suggestion
        this.notificator.showInstantSuccess(
          'SHARED_LIB.PERUN.COMPONENTS.RESET_PASSWORD_DIALOG.SUCCESS'
        );
```

bodnara (changes requested):
I have just some small comments.

bodnara (approved):

mattjoke (approved):

HejdaJakub (approved):

PR #1388, opened 2023-08-09
feat: remove access for service-accounts

mattjoke:
* service accounts are now not able to log in
* a new dialog has been added, that can redirect users to /login or /service-access
* the dialog logic is evaluated when principal is loading

### DIFF
--- a/apps/admin-gui/src/assets/i18n/en.json
+++ b/apps/admin-gui/src/assets/i18n/en.json
@@ -3104,6 +3104,11 @@
         "USER_DONT_EXIST": {
           "TITLE": "The requested user (by Id or external identity) doesn't exist."
         },
+        "USER_NOT_ALLOWED": {
+          "TITLE": "Access not allowed",
+          "MESSAGE": "Service accounts are not allowed to access this application.",
+          "REDIRECT": "Redirect to login page"
+        },
         "TABLE_OPTIONS": {
           "EXPORT_TO_FILE": "Export to file",
           "ALL_DATA": "All data",
--- a/libs/general/src/lib/user-not-allowed-access/user-not-allowed-access.component.html
+++ b/libs/general/src/lib/user-not-allowed-access/user-not-allowed-access.component.html
@@ -1,0 +1,11 @@
+<h1 mat-dialog-title>
+  {{'SHARED_LIB.PERUN.COMPONENTS.USER_NOT_ALLOWED.TITLE' | translate}}
+</h1>
+<div mat-dialog-content class="dialog-container">
+  {{'SHARED_LIB.PERUN.COMPONENTS.USER_NOT_ALLOWED.MESSAGE' | translate}}
+</div>
+<div mat-dialog-actions>
+  <button mat-flat-button class="ms-auto" color="accent" (click)="redirect()">
+    {{'SHARED_LIB.PERUN.COMPONENTS.USER_NOT_ALLOWED.REDIRECT' | translate}}
+  </button>
+</div>
--- a/libs/perun/services/src/lib/init-auth.service.ts
+++ b/libs/perun/services/src/lib/init-auth.service.ts
@@ -12,6 +12,7 @@ import { OAuthInfoEvent, OAuthService } from 'angular-oauth2-oidc';
 import { filter } from 'rxjs/operators';
 import { firstValueFrom, timer } from 'rxjs';
 import { MfaHandlerService } from './mfa-handler.service';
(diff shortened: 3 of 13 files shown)


libs/perun/services/src/lib/init-auth.service.ts
```diff
@@ -109,6 +110,10 @@ export class InitAuthService {
       if (perunPrincipal.user === null) {
         const config = getDefaultDialogConfig();
         this.dialog.open(UserDontExistDialogComponent, config);
+      } else if (perunPrincipal.user.majorSpecificType === 'SERVICE') {
```
bodnara, line 113:
```suggestion
      } else if (perunPrincipal.user.serviceUser) {
```

libs/perun/services/src/lib/init-auth.service.ts
```diff
@@ -109,6 +110,10 @@ export class InitAuthService {
...
+        this.authService.logout();
```
bodnara, line 114: Depending on the proxy and configuration the logout might redirect you without the dialog ever showing. But even if the logout was performed at later stage the redirect would mess up the flow.

libs/general/src/lib/user-not-allowed-access/user-not-allowed-access.component.html
```diff
@@ -0,0 +1,16 @@
...
+</h2>
+
+<mat-dialog-actions align="end">
+  <button mat-flat-button color="primary" (click)="redirectToLogin()">
+    {{'SHARED_LIB.PERUN.COMPONENTS.USER_NOT_ALLOWED.LOGIN' | translate}}
+  </button>
+  <button mat-flat-button color="primary" (click)="redirectToServiceLogin()">
+    {{'SHARED_LIB.PERUN.COMPONENTS.USER_NOT_ALLOWED.SERVICE_LOGIN' | translate}}
+  </button>
+</mat-dialog-actions>
```
bodnara, line 16: Fixes the dialog, layout and colors.
```suggestion
<h1 mat-dialog-title>{{'SHARED_LIB.PERUN.COMPONENTS.USER_NOT_ALLOWED.TITLE' | translate}}</h1>
<div mat-dialog-content class="dialog-container">
  {{'SHARED_LIB.PERUN.COMPONENTS.USER_NOT_ALLOWED.MESSAGE' | translate}}
</div>
<div mat-dialog-actions>
  <button mat-flat-button class="ms-auto" color="accent" (click)="redirectToLogin()">
    {{'SHARED_LIB.PERUN.COMPONENTS.USER_NOT_ALLOWED.LOGIN' | translate}}
  </button>
  <button mat-flat-button color="accent" (click)="redirectToServiceLogin()">
    {{'SHARED_LIB.PERUN.COMPONENTS.USER_NOT_ALLOWED.SERVICE_LOGIN' | translate}}
  </button>
</div>
```

apps/admin-gui/src/assets/i18n/en.json
```diff
@@ -3104,6 +3104,12 @@
         "USER_DONT_EXIST": {
           "TITLE": "The requested user (by Id or external identity) doesn't exist."
         },
+        "USER_NOT_ALLOWED": {
+          "TITLE": "Access not allowed",
+          "MESSAGE": "Service accounts are not allowed to access this application.",
+          "LOGIN": "Redirect to /login",
+          "SERVICE_LOGIN": "Redirect to /service-access"
```
bodnara, line 3111: This a maybe a preference thing but this is not very elegant. I would honestly maybe put here "Change account" and redirect to login page only.

bodnara (changes requested):

libs/perun/services/src/lib/init-auth.service.ts
mattjoke: We could detect the proxy configuration and then we could do logout with a redirect, i.e. logout would happen after the dialog was closed.

apps/admin-gui/src/assets/i18n/en.json
mattjoke: What should be the correct way when someone logs in from the service-access? Should it redirect to it? Or just to the login page?

bodnara: I guess the best option for user is to redirect to login option that was used before. So for OIDC to /login and BA to /service-access.

HejdaJakub (approved):

bodnara (approved):

PR #1391, opened 2023-08-10
fix(lib): display just one alert on login page

HejdaJakub:
* When the user is logged out from service-access, he/she can see the info alter about successfull log out. Then if the user try to log in again with incorrect credentials, another alert will be displayed, but the old one used to be still presented on login page.


mattjoke (approved):

HejdaJakub:
We want to avoid this situation:
Screenshot from 2023-08-10 12-44-25

PR #1397, opened 2023-08-16
feat(admin): added support for new role

mattjoke:
* new role PERUNADMINBA
* this role is the same as PERUNADMIN, but skips MFA
* regenerated openapi
* added 'PERUNADMINBA' description field to roles in en.json


libs/perun/services/src/lib/gui-auth-resolver.service.ts
```diff
@@ -89,82 +89,101 @@ export class GuiAuthResolver {
...
   isOnlySponsor(): boolean {
     return this.hasAtLeastOne(Role.SPONSOR);
   }
 
   isThisGroupAdmin(id: number): boolean {
-    return this.editableGroups.includes(id) || this.principalRoles.has(Role.PERUNADMIN);
+    return (
+      this.editableGroups.includes(id) ||
+      this.principalRoles.has(Role.PERUNADMIN) ||
+      this.principalRoles.has(Role.PERUNADMINBA)
```
xflord, line 139: Just a thought - how about replacing all of these checks for PERUNADMIN with the call of `isPerunAdmin()`. This way should we ever remove `PERUNADMINBA` or add another similar role, the number of necessary changes will be minimal.

xflord (approved):
Looks ok

libs/perun/services/src/lib/gui-auth-resolver.service.ts
mattjoke: I replaced it as much as I could, please check, if I haven't missed anything.

xflord: I think you got all of it. Have you also considered adding some readable text for the role in `en.json`? (the roles page component uses this)

bodnara (approved):

sarkapalkovicova (approved):

PR #1398, opened 2023-08-17
feat(admin): customizable items per page in dashboard tables

kofzera:


xflord (approved):
I would consider setting the default to 5 per page, but depends. What does @bodnara think

bodnara (approved):
I think is it fine like this.

PR #1399, opened 2023-08-18
fix(admin): notifications sending enabled control changed to toggle

kofzera:


xflord (changes requested):
Looks good, but part of the task was also to add a success notification when the 'sending setting' changes (similarly as a notification pops up when you add/remove entities). You can do this using the `NotificatorService`

kofzera:
> Looks good, but part of the task was also to add a success notification when the 'sending setting' changes (similarly as a notification pops up when you add/remove entities). You can do this using the `NotificatorService`

Somehow forgot about that, should be there now :smile:

apps/admin-gui/src/app/vos/components/notification-list/notification-list.component.ts
```diff
@@ -99,12 +108,14 @@ export class NotificationListComponent implements OnChanges, AfterViewInit {
         .setSendingEnabled({ mails: [applicationMail], enabled: false })
         .subscribe(() => {
           applicationMail.send = false;
+          this.notificator.showSuccess(this.disabledSendingMessage);
```
bodnara, line 111: Preferably use the `showInstantSuccess` method of the notificator. It handles the translation itself.

sarkapalkovicova:
In `add-edit-notification-dialog` is also used a checkbox. Shouldn't be here toggle too? @bodnara

bodnara:
> In `add-edit-notification-dialog` is also used a checkbox. Shouldn't be here toggle too? @bodnara

In this case checkbox is better option, since the action requires conformation (click on save button). In case of immediate action toggle is preferable.

sarkapalkovicova (approved):

PR #1401, opened 2023-08-21
feat(admin): modify user without role message

kofzera:
After logging-in of a user whose roles are only SELF or MEMBERSHIP a notice is displayed about this fact with button leading to user profile.


apps/admin-gui/src/app/users/pages/user-detail-page/user-dashboard/user-dashboard.component.ts
```diff
@@ -25,7 +25,7 @@ export class UserDashboardComponent implements OnInit {
   roles: { [key: string]: { [key: string]: Array<number> } } = {};
   userProfileUrl = '';
   roleNames: string[];
-  isOnlySelfRole = false;
+  isOnlySelfOrMembershipRole = false;
```
bodnara, line 28: I think we should also include the role `SPONSORSHIP`. Also choose more suitable name for this variable.

bodnara (approved):

PR #1403, opened 2023-08-22
feat(admin): check disabled and hidden dependencies for application items

HejdaJakub:
* Provide a warning message when trying to change source or destination attribute for items which are marked as hidden or disabled dependency for another form item.
* Display an error in dialog when trying to delete an item which is marked as hidden or disabled dependency for another form item (it used to throw an uncatched exception).

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/edit-application-form-item-dialog/edit-application-form-item-dialog.component.ts
+++ b/apps/admin-gui/src/app/shared/components/dialogs/edit-application-form-item-dialog/edit-application-form-item-dialog.component.ts
@@ -1,6 +1,5 @@
 import { ChangeDetectorRef, Component, Inject, OnInit } from '@angular/core';
 import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
-import { TranslateService } from '@ngx-translate/core';
 import {
   ApplicationFormItem,
   AppType,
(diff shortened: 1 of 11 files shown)


apps/admin-gui/src/app/shared/components/dialogs/edit-application-form-item-dialog/edit-application-form-item-dialog.component.ts
```diff
@@ -145,6 +150,28 @@ export class EditApplicationFormItemDialogComponent implements OnInit {
...
+      this.warningMessage = this.translate.instant(
+        'DIALOGS.APPLICATION_FORM_EDIT_ITEM.DEPENDENCY_WARNING_MESSAGE',
+        {
+          dependency: hiddenDependencyForItem ? 'hidden' : 'disabled',
+          shortname: hiddenDependencyForItem
+            ? hiddenDependencyForItem.shortname
+            : disabledDependencyForItem.shortname,
+        }
+      );
+    }
```
bodnara, line 172: Little nit-picky but this way there are two unnecessary if-else conditions. Same goes for the deletion part.

bodnara (changes requested):
The way it works now when editing is that only one warning will be displayed. When both source/destination attributes are changed  both should have the warning.

bodnara (approved):

PR #1404, opened 2023-08-22
fix(admin): missaligned icons in action-buttons

kofzera:
* Fix action-buttons where the icon inside was misaligned
* Buttons in the group application form changed to be consistent with VO
* Settings button in VO application form aligned to be consistent with group


HejdaJakub (approved):
I didn't find any other misalignments, so it's fine from my side.

PR #1406, opened 2023-08-23
fix: initialize colors from instance config

bodnara:
* all color variables are now initialized by the script
* removed creation of css variables during runtime
* removed unused classes in styles.scss for all apps
* formated styles.scss for all apps


bodnara:
Seems like the real color values need to be present before build in order for the theme to be applied correctly. So the migration script is needed. All colors, are now added with the script and not during runtime. The problem was with the path to instance config on the machine, so it only worked locally.

Right now the script is capable of setting up all variables from theme and we should consider applying it to all components that need color from config. I did change the `side-nav` (admin-gui only) in this PR to have it's css class include the colors and it works fine. So there is no need to later load and "manually" add styles to the element. I would say it is better to have it in one place for debug and clarity purposes.

Another thing is that it is imho excessive to have 40+ theme variables for such details as color of text on hover in certain component. I would even say that it is a wrong approach, since we are following the material design theming guideline (https://material.angular.io/guide/theming). Which we use for material components but do not use for our components. This article shows how to apply the theme for custom components https://indepth.dev/tutorials/angular/angular-material-theming-system-complete-guide#apply-angular-material’s-theme-to-custom-component. I think it would be worth the effort to follow this practice and get rid of most of the theme variables.

HejdaJakub:
I think that after these changes the colors from `instanceConfig` are not correctly applied. :eyes:

bodnara:
To get correct color from instance config locally you have to run the script manually with a flag `node color-migration.js 1` (technically any "true" value will work) so the script looks for instance config in correct directory.

mattjoke (approved):
Looks good :+1:  I would personally leave the duplication warnings so the build is not cluttered. Otherwise it seems to be working

HejdaJakub (approved):

HejdaJakub:
The release is done, so now we can merge this PR @bodnara?

PR #1408, opened 2023-08-25
fix: open application in the new tab

HejdaJakub:
* On a ctrl (command) + left mouse click is now opened the application in the new tab and routing is stopped in the parent tab.
* This is now implemented also for side menu, tables and inner page menu in user-profile (authentication and settings).

### DIFF
--- a/apps/admin-gui/src/app/shared/side-menu/side-menu-item/side-menu-item.component.ts
+++ b/apps/admin-gui/src/app/shared/side-menu/side-menu-item/side-menu-item.component.ts
@@ -43,12 +43,18 @@ export class SideMenuItemComponent {
     });
   }
 
-  linkClicked(link: SideMenuLink): void {
-    // Navigate or expand children based on item
-    if (!link.children) {
-      void this.router.navigate(link.url);
-    } else {
-      this.expandedTilesStore.setItem(link.showChildren);
+  /**
+   * Navigate or expand children based on item if ctrl (metaKey) is not pressed
+   * @param link SideMenuLink with url and list of children
+   * @param event mouse event used for detect ctrlKey (support for Windows/Linux keyboards) or metaKey (support for Mac keyboards)
+   */
+  navigateOrExpandSideMenu(link: SideMenuLink, event: MouseEvent): void {
+    if (!event.ctrlKey && !event.metaKey) {
+      if (!link.children) {
+        void this.router.navigate(link.url);
+      } else {
+        this.expandedTilesStore.setItem(link.showChildren);
+      }
     }
   }
 }
--- a/libs/perun/directives/src/lib/middle-click-router-link.ts
+++ b/libs/perun/directives/src/lib/middle-click-router-link.ts
@@ -1,4 +1,4 @@
-import { Directive, HostListener, Input } from '@angular/core';
+import { Directive, ElementRef, HostListener, Input, Renderer2 } from '@angular/core';
 
 @Directive({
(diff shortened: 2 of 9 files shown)


HejdaJakub:
Questions:
- The side menu works for admin-gui, should it be fixed also for user-profile? (there isn't handled even middle click now)
- What about iOS, is `ctrlKey` check working the same way or not?

mattjoke (commented):
Looks good :+1:  However, I found a weird bug of some sort. If I Ctrl+ click on `Advanced Settings` the new tab opens correctly, but the Advanced Settings button/title is nowhere to be seen in the main content. The advanced settings option is, however, present in the side menu.
image

HejdaJakub:
> Looks good 👍 However, I found a weird bug of some sort. If I Ctrl+ click on `Advanced Settings` the new tab opens correctly, but the Advanced Settings button/title is nowhere to be seen in the main content. The advanced settings option is, however, present in the side menu.

Good point, it should be fixed now. I discussed it in the office and we agreed that these expandable items in the side menu are more buttons than navigators (just expands submenu), so it shouldn't be possible to open a new window via middle click / ctrl+left click on these items (it is disabled now).

apps/admin-gui/src/app/shared/side-menu/side-menu-item/side-menu-item.component.ts
```diff
@@ -23,6 +23,9 @@ export class SideMenuItemComponent {
   rippleColor = 'rgba(255, 255, 255, 0.1)';
   expandSections: Map<ExpandableSectionId, boolean>;
 
+  protected readonly window = window;
+  protected readonly location = location;
+
```
bodnara, line 28: Unused ?

libs/perun/directives/src/lib/middle-click-router-link.ts
```diff
@@ -1,20 +1,48 @@
...
+    this.renderer.listen(htmlElement, 'click', (event: MouseEvent) => {
+      // ctrl (command) + left mouse click will stop routing and open router link in the new tab
+      if (
+        // ctrlKey = support for Windows/Linux keyboards
+        // metaKey = support for Mac keyboards
+        (event.ctrlKey || event.metaKey) &&
+        event.button === 0 &&
+        this.perunWebAppsMiddleClickRouterLink !== null &&
+        this.perunWebAppsMiddleClickRouterLink !== undefined
+      ) {
```
bodnara, line 20: Nitpick, for both occurences.
```suggestion
      if (
        // ctrlKey = support for Windows/Linux keyboards
        // metaKey = support for Mac keyboards
        (event.ctrlKey || event.metaKey) &&
        event.button === 0 &&
        this.perunWebAppsMiddleClickRouterLink
      ) {
```

bodnara (approved):
Works for links, side-menu, tiles and in tables. I believe that should cover all cases.

mattjoke (approved):
Looks good :+1:

bodnara:
I have just noticed that using crtl+click in sidemenu does redirect in current window as well. Other elements/types of navigation do not navigate for current window.

HejdaJakub:
> I have just noticed that using crtl+click in sidemenu does redirect in current window as well. Other elements/types of navigation do not navigate for current window.

Yeah, I broke this behavior with the last edit, thanks for noticing. Now it should work and I also realized, that I didn't handle the `metaKey` (Mac support) for the side menu, so it is also fixed now.

bodnara (approved):

mattjoke (approved):

PR #1409, opened 2023-08-28
fix(profile): navigation working properly in mailing lists settings

kofzera:
Fixed a problem of getting stuck in the mailing lists settings. When navigating away from the component it clears the query params and replaces the old url. The query params need to be removed also from the new component. This is done in the onDestroyed method of the mailing lists.

### DIFF
--- a/libs/perun/components/src/lib/settings-mailing-lists/mailing-lists.component.ts
+++ b/libs/perun/components/src/lib/settings-mailing-lists/mailing-lists.component.ts
@@ -1,5 +1,6 @@
 import { Component, Input, OnInit, OnDestroy } from '@angular/core';
-import { ActivatedRoute, Router } from '@angular/router';
+import { Location, LocationStrategy, PathLocationStrategy } from '@angular/common';
+import { ActivatedRoute, NavigationStart, Router } from '@angular/router';
 import {
   Attribute,
   AttributesManagerService,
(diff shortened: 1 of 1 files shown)


HejdaJakub:
As we discussed at the daily call, we want to preserve other query params. :raised_hands:

xflord (commented):
the query parameters return to the url after navigating to more than 1 site from the mailing lists page. I guess the router still remembers the query params, even when URL has been modified through `location`

libs/perun/components/src/lib/settings-mailing-lists/mailing-lists.component.ts
```diff
@@ -82,6 +86,17 @@ export class MailingListsComponent implements OnInit, OnDestroy {
...
+      .pipe(filter((event): event is NavigationStart => event instanceof NavigationStart))
+      .subscribe((event) => {
+        if (!event.url.startsWith(location.pathname)) {
+          // clear the query params when navigating away
+          this.location.replaceState(
+            location.pathname,
+            this.clearParamsFromCurrUrl(['vo', 'resource'])
+          );
+        }
+      });
```
xflord, line 99: Is any of this really necessary? I tried the code only with the change to ngDestroy and it seemed to work correctly

kofzera: It is there so that the params will stay removed also when going back in the browser.

xflord (approved):

HejdaJakub:
This way the browser history is still somehow broken. If I click on e.g. 4 vos and then navigate somewhere else, for the last vo the query params are removed (when I click on the browser back button), but for the rest of vos nope. At the same time, I need to click something like 8 times to go back before mailing lists (if I clicked on 4 vos before). 

I am thinking about not navigating while triggering the `changeSelectedVo`/`changeSelectedResource`, but maybe use something like history.replaceState() (or maybe your this.location.replaceState() ?) - it will not add new data to history on this page, and with the combination of your current code, it should work properly.

HejdaJakub (approved):

PR #1412, opened 2023-08-29
feat: close session expired dialog in all tabs

bodnara:
The reauthentication of user in multiple tabs works fine. However the session expired dialog stayed present aand prevented usage of the app. This is no longer the case

### DIFF
--- a/libs/perun/services/src/lib/ApiInterceptor.ts
+++ b/libs/perun/services/src/lib/ApiInterceptor.ts
@@ -56,20 +56,23 @@ export class ApiInterceptor implements HttpInterceptor {
     ) {
       const config = getDefaultDialogConfig();
       config.width = '450px';
+      config.id = 'SessionExpirationDialog';
 
       // Skip if the dialog is already open
-      if (this.dialogRefSessionExpiration == null) {
+      if (!this.dialogRefSessionExpiration) {
         this.dialogRefSessionExpiration = this.dialog.open(
           SessionExpirationDialogComponent,
           config
         );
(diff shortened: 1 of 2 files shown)


xflord:
This works for me when I manually logout - logout of one window (or delete storage), the dialog appears on other windows, and then disappears when relogging. However, when revoking the refresh/access token manually and waiting for the session to expire, the dialog does not disappear after logging in.

bodnara:
> This works for me when I manually logout - logout of one window (or delete storage), the dialog appears on other windows, and then disappears when relogging. However, when revoking the refresh/access token manually and waiting for the session to expire, the dialog does not disappear after logging in.

I tried to make this work for revoking as well, but right now there is not a simple solution. Also after some discussion with @vyskocilpavel, revoking the token is different case from when it expires and should be handled and presented to users differently.
The main objective of this PR is to handle situation when the access token expires and cannot be refresh due to inactivity.

mattjoke (approved):
The code looks okay :+1:  I tried to manually test it and it seemed to work.

xflord:
> > This works for me when I manually logout - logout of one window (or delete storage), the dialog appears on other windows, and then disappears when relogging. However, when revoking the refresh/access token manually and waiting for the session to expire, the dialog does not disappear after logging in.
> 
> I tried to make this work for revoking as well, but right now there is not a simple solution. Also after some discussion with @vyskocilpavel, revoking the token is different case from when it expires and should be handled and presented to users differently. The main objective of this PR is to handle situation when the access token expires and cannot be refresh due to inactivity.

If that's the case then the changes seem to do what was intended

xflord (approved):

PR #1414, opened 2023-08-29
fix: do not filter attributes with empy value

bodnara:
Filtering of attributes with empty values from table was removed. It is most likely not wanted anywhere and caused problem when adding new attribute.


xflord (commented):
I'm guessing the filter was there to remove the clutter caused by attributes with empty values since such attributes do not contain useful information. An example of this is in the screenshot.
How about keeping the filter while turning it off in the add dialog?
Screenshot from 2023-08-29 18-25-08

HejdaJakub:
I agree with @xflord, it should work correctly otherwise. :raised_hands:

bodnara:
I think the use case is valid. However I would maybe add this as an option to filtering so it can be toggled on/off. Right now it may be confusing why some attributes are never showing and at the same time cannot be added.

xflord (approved):

HejdaJakub (approved):

PR #1417, opened 2023-08-31
fix(admin): indirect membership

sarkapalkovicova:
- Refactored indirect membership’s origin dialog.
- Added column for type in the members list.


bodnara (approved):
I would say it it nice improvement. One note I found a 3rd? type of icon
image

mattjoke (approved):
Looks good :+1:

PR #1418, opened 2023-09-01
test(admin): fixing the failing E2E tests

HejdaJakub:
* New path to advanced settings (and related logic) broke some tests, so it it fixed now.
* Removed skip from tests which were failing due to the bug with privileges because this bug has been already fixed IMHO (I think it was related to the refreshing roles on BE where we have done some changes to avoid PrivilegeException).
* Fixed some 'hidden' items covered by another item (like header).

### DIFF
--- a/apps/admin-gui-e2e/src/e2e/facilities/facility-admin.cy.js
+++ b/apps/admin-gui-e2e/src/e2e/facilities/facility-admin.cy.js
@@ -176,63 +176,62 @@ describe('Facility management with role Facility admin', () => {
       .should('not.exist');
   });
 
-  it('test add facility manager', () => {
-    cy.intercept('**/facilitiesManager/getEnrichedFacilities')
-      .as('getEnrichedFacilities')
-      .intercept('**/usersManager/findRichUsersWithAttributes?**')
-      .as('findRichUsers')
-      // .wait('@getEnrichedFacilities')
-      .get('[data-cy=filter-input]')
-      .type(dbFacilityName2, {force: true})
-      .get(`[data-cy=${dbFacilityName2}]`)
-      .click()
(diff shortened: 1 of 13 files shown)


HejdaJakub:
- It needs to be tested with this version of containers: https://gitlab.ics.muni.cz/perun/e2e-perun/-/merge_requests/28.
- This needs to be also included during the testing: https://github.com/CESNET/perun-web-apps/pull/1415 (I will rebase it after it will be merged).
- Until this is not released (https://github.com/CESNET/perun/pull/4062) and used as a backend, the last test (for blocked logins) in the perun-observer file will still fail.

apps/admin-gui-e2e/src/e2e/facilities/facility-admin.cy.js
```diff
@@ -13,6 +13,7 @@ describe('Facility management with role Facility admin', () => {
 
   before(() => {
     cy.login('FACILITY_MANAGER', 'facilityManager');
+    window.localStorage.setItem('settings', 'true');
```
xflord, line 16: The storage items could be set before all the tests to save code, right? I believe support/e2e.ts is the place for this

xflord (commented):
will test locally and see what happens

xflord (approved):
everything passes, good job

HejdaJakub:
Good point, it works fine. The setting of the localStorage items has been moved to support/e2e.ts file.
I noticed that vo-management/perun-admin test has incorrect title in describe block, so I changed it.
Agreed that all tests now passing.
image

mattjoke (approved):
Everything seems to pass. Good job :+1: 
image

xflord (approved):

PR #1419, opened 2023-09-04
feat(admin): add bulk operations to applications overview

michalberky:
* Added bulk selection and operations buttons to application overview page.
* Buttons are only enabled when applications with same state are selected and that operation is available for this state.
* Created a new dialog that prompts the user to confirm the bulk operation

### DIFF
--- a/apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.html
+++ b/apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.html
@@ -28,6 +28,52 @@
     </button>
   </span>
   <div>
+    <span [matTooltip]="!canApprove ? tooltipMessages.approve : ''">
+      <button
+        color="accent"
+        *ngIf="approveAuth"
+        class="me-2"
+        mat-flat-button
+        (click)="onApproveClicked()"
+        [disabled]="!canApprove">
+        {{'VO_DETAIL.APPLICATION.APPLICATION_DETAIL.APPROVE' | translate}}
+      </button>
+    </span>
+    <span [matTooltip]="!canReject ? tooltipMessages.reject : ''">
+      <button
+        color="warn"
+        *ngIf="rejectAuth"
+        class="me-2"
+        mat-flat-button
+        (click)="onRejectClicked()"
+        [disabled]="!canReject">
+        {{'VO_DETAIL.APPLICATION.APPLICATION_DETAIL.REJECT' | translate}}
+      </button>
+    </span>
+    <span [matTooltip]="!canDelete ? tooltipMessages.delete : ''">
+      <button
+        color="warn"
+        *ngIf="deleteAuth"
+        class="me-2"
+        mat-flat-button
+        (click)="onDeleteClicked()"
+        [disabled]="!canDelete">
+        {{'VO_DETAIL.APPLICATION.APPLICATION_DETAIL.DELETE' | translate}}
+      </button>
+    </span>
+    <span [matTooltip]="!canSendNotification ? tooltipMessages.sendNotifications : ''">
+      <button
+        class="me-2"
+        *ngIf="resendAuth"
+        mat-stroked-button
+        (click)="onResendClicked()"
+        [disabled]="!canSendNotification">
+        <i class="material-icons"> mail_outline </i>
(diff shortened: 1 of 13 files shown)


michalberky:
**Do not merge**
Currently waiting for backend methods for the bulk operations to be finished and used

apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.html
```diff
@@ -28,6 +28,48 @@ <h1 class="page-subtitle">{{'VO_DETAIL.APPLICATION.TITLE' | translate}}</h1>
...
   <div>
+    <span [matTooltip]="!canApprove ? tooltipMessages.approve : ''">
+      <button
+        color="accent"
+        class="me-2"
+        mat-flat-button
+        (click)="onApproveClicked()"
+        [disabled]="!canApprove">
+        {{'VO_DETAIL.APPLICATION.APPLICATION_DETAIL.APPROVE' | translate}}
+      </button>
```
bodnara, line 39: We should also check authorizations for all of the buttons. Check if principal is authorized to use bulk method.

apps/admin-gui/src/app/vos/pages/vo-detail-page/vo-applications/vo-applications.component.ts
```diff
@@ -164,4 +178,141 @@ export class VoApplicationsComponent implements OnInit {
...
+      action: selectedAction,
+      title: title,
+      description: description,
+    };
+
+    const dialogRef = this.dialog.open(ApplicationsBulkOperationDialogComponent, config);
+    return dialogRef;
+  }
+
+  private setButtonsTooltips(): void {
```
bodnara, line 291: The tooltips should also be set when initialized.

libs/perun/dialogs/src/lib/applications-bulk-operation-dialog/applications-bulk-operation-dialog.component.html
```diff
@@ -0,0 +1,52 @@
...
+    {{'VO_DETAIL.APPLICATION.DIALOG.CANCEL' | translate}}
+  </button>
+  <button
+    *ngIf="action === 'APPROVE'"
+    color="accent"
+    class="me-2"
+    mat-flat-button
+    (click)="onConfirm()">
+    {{'VO_DETAIL.APPLICATION.APPLICATION_DETAIL.APPROVE' | translate}}
+  </button>
```
bodnara, line 31: If at the end confirm will just close dialog it would be nicer to have one button just set the proper label beforehand.

libs/perun/dialogs/src/lib/applications-bulk-operation-dialog/applications-bulk-operation-dialog.component.scss
```diff
@@ -0,0 +1,12 @@
+.dialog-content {
+  padding-top: 0 !important;
+  max-height: 300px;
+  max-width: 95%;
+  overflow-y: auto;
+  overflow-x: auto;
+}
```
bodnara, line 7: I would most of this we already get by using the `mat-dialog-content`. You should set up the dialog so that all content is under one parent with `mat-dialog-content`.

bodnara (changes requested):

bodnara:
Also I just realized all this should be possible for groups as well.

libs/perun/dialogs/src/lib/applications-bulk-operation-dialog/applications-bulk-operation-dialog.component.html
michalberky: I made them 4 separate buttons since 'RESEND' is mat-stroked-button, unlike the other 3 that are mat-flat-button. And within those 3 there's also difference in color of 'APPROVE' button. So it is not just about the label.

bodnara: I see but in this case, as a confirmation dialog, it might be more in line with other dialogs to use flat accent button with label of the action for all cases.

bodnara:
Will be fixed in [another PR.](https://github.com/CESNET/perun-web-apps/pull/1421)

PR #1421, opened 2023-09-12
feat(admin): add bulk operations to applications overview

bodnara:
Application overview now allows to approve, reject, delete and resend notification in bulk. Selected applications have to have the same state. Each action will prompt a confirmation dialog. The page was refactored as it is shared for Vo and Groups.


libs/perun/dialogs/src/lib/applications-bulk-operation-dialog/applications-bulk-operation-dialog.component.html
```diff
@@ -0,0 +1,63 @@
...
+        </mat-form-field>
+      </div>
+    </div>
+
+    <perun-web-apps-alert alert_type="warn">
+      {{ data.description | translate: {count: dataSource.data.length} }}
+    </perun-web-apps-alert>
+
+    <div class="table">
+      <mat-table [dataSource]="dataSource" matSort>
```
xflord, line 45: This table only displays data for the default columns (see image).
Screenshot from 2023-09-14 09-41-58
At least group info should work IMO, the rest can be omitted.

xflord (changes requested):
Also noticed an unrelated bug, where changing column sorting while some entities are selected will break the selection. Will create a new task.
Other than that, the refactor looks good.

HejdaJakub (changes requested):
- When I click on refresh applications / show more details / show group applications, the `ExpressionChangedAfterItHasBeenCheckedError` occurs.
- The part with buttons, select items, etc. is not responsive now.
image

michalberky (changes requested):
There is also an issue with the table always trying to fit all columns into the dialog width not exceeding it. It becomes more apparent on mobile, where it easily becomes unusable.

image

bodnara:
Fixed:
- applications for group
- custom columns (attribute had to be loaded after federation attributes as @xflord suggested)
- responsiveness of buttons, etc.

Regarding the table inside confirm dialog. It would be the best to use existing table that has proper behavior and can display all data correctly. Unfortunately with how the dynamic application list is set up it is impossible to pass data to the table. It is very inconvenient as such tables can not be reused anywhere else. For example the group list can handle both paginated and "normal" data types.
For now I would be ok, with simply copying the template from dynamic app. list into confirm dialog with the logic to display things correctly needed. And rework the table to support passing data in later.

Other problem is circular dependencies. The module structure right now does not allow to use eg. "components" inside of "dialogs" and we have half of the stuff in Shared admin gui module and other half in libs. Which sometimes requires moving around multiple seemingly unrelated component. Again something for a longer discussion.
For this PR I thing this could be the easiest option:
- move all of this logic (app list, pipes for data display, shared view component and confirm dialog) to shared admin gui module

bodnara:
Did some file tetris, in the end not so many changes. Duplicated table and some logic into confirm dialog. I was considering updating the not paginated application list but both options seem equally bad and this way it was quicker. Sprinkled some `cd.detectChanges()` to make change detection error go away.

michalberky (commented):
- The table in dialog works great this way!

- Even though refreshing the table now doesn't cause ExpressionChangedAfterItHasBeenCheckedError, the error still shows up upon loading the vo/group applications page and then whenever filters (state, filter text field, date range or column selection) are changed. The table inside of the dialog doesn't cause this error in any way.

bodnara:
>   Even though refreshing the table now doesn't cause _ExpressionChangedAfterItHasBeenCheckedError_, the error still shows up upon loading the vo/group applications page and then whenever filters (state, filter text field, date range or column selection) are changed. The table inside of the dialog doesn't cause this error in any way.

Fixed some, but date and column setting change still triggers exception. The problem is with how the dynamic table is set up and how it handles loading. Honestly I do not want to spend more time on this since it makes more sense to fix dynamic tables as whole, instead of "fighting" with it every time we use it. Will create task for this.

michalberky (approved):

xflord (approved):
I agree with solving the not-so-related issues within other tasks, the bulk app operations work fine so I'm merging this

PR #1423, opened 2023-09-14
feat(admin): display also indirect managers

HejdaJakub:
* Under the group on the managers page there is now available to optionally display also indirect managers of this group.

### DIFF
--- a/apps/admin-gui/src/app/shared/components/managers-page/managers-page.component.html
+++ b/apps/admin-gui/src/app/shared/components/managers-page/managers-page.component.html
@@ -34,7 +34,7 @@
 
       <mat-form-field data-cy="role-filter" class="me-2 mt-2">
         <mat-label>{{'SHARED.COMPONENTS.MANAGERS_PAGE.SELECT_ROLE' | translate}}</mat-label>
-        <mat-select (selectionChange)="refreshUsers()" [(value)]="selectedRole">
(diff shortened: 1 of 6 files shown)


HejdaJakub:
Merge after this PR: https://github.com/CESNET/perun/pull/4076

xflord (commented):
The changes look fine, but have you made sure that RESOURCE ADMIN can view the managers page? I took a quick look and based on the policies, it seems that someone with just RESOURCEADMIN can't access the managers page for a group assigned to their resource (which was why the core issue of the task).
Also, I'm guessing someone will eventually request this feature for manager pages of the other entities as well.

HejdaJakub:
I think that for this use case is a separate task (https://perunaai.atlassian.net/browse/ST-1189). 

But as you said, that someone can want this functionality for other entities, I just realized (and tested), that maybe we can use the already existing method this.authzService.getAuthzRichAdmins(..) with property onlyDirectAdmis = false. I thought that this method didn't work properly, but now it looks, like it works fine :thinking:. If you agree that it works, I would close the PR for backend and use this method in this PR instead of the new method from groupsManager - after that it will be better prepared also for other entities.

HejdaJakub:
As mentioned before and as we discussed with @xflord, I reverted the changes related to the new method from BE, which is no longer needed, and the authzService.getAuthzRichAdmins(...) method is used instead.

xflord (approved):

PR #1425, opened 2023-09-19
feat(admin): jump onto specific page

sarkapalkovicova:
Added input for inserting a wanted page number.

### DIFF
--- a/libs/perun/utils/src/lib/table-wrapper/table-wrapper.component.ts
+++ b/libs/perun/utils/src/lib/table-wrapper/table-wrapper.component.ts
@@ -18,7 +18,7 @@ import { TableConfigService } from '@perun-web-apps/config/table-config';
 @Component({
   selector: 'perun-web-apps-table-wrapper',
   templateUrl: './table-wrapper.component.html',
-  styleUrls: ['./table-wrapper.component.css'],
+  styleUrls: ['./table-wrapper.component.scss'],
   providers: [
     { provide: MAT_PAGINATOR_DEFAULT_OPTIONS, useValue: { formFieldAppearance: 'fill' } },
   ],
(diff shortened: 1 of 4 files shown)


michalberky (changes requested):
- The input field is not vertically aligned with rest of the table controls.

Screenshot from 2023-09-19 11-07-38


- I was also thinking if we need two sets of buttons incrementing the current page by one - the 'Previous page' and 'Next page' buttons that were already there and the arrows incrementing the number (although those only change the page on change of focus)
- I feel like possibility to confirm the new page number by pressing enter as well, instead of just focus loss, would be more intuitive
- Currently when editing the number, I can move cursor with mouse but can't with arrow keys. Also removing numbers is only possible with backspace and not with delete.

sarkapalkovicova:
> * The input field is not vertically aligned with rest of the table controls.
> 
> Screenshot from 2023-09-19 11-07-38
> 
> * I was also thinking if we need two sets of buttons incrementing the current page by one - the 'Previous page' and 'Next page' buttons that were already there and the arrows incrementing the number (although those only change the page on change of focus)
> * I feel like possibility to confirm the new page number by pressing enter as well, instead of just focus loss, would be more intuitive
> * Currently when editing the number, I can move cursor with mouse but can't with arrow keys. Also removing numbers is only possible with backspace and not with delete.

Arrows in input just increment the number that is inserted. Out previous/next page buttons immediately reload data. For me, these are 2 different things, but if you insist, I can hide arrows in input.  
I  am not sure if we use confirmation by enter somewhere in the GUI. Do we want that? @bodnara

michalberky:
> Arrows in input just increment the number that is inserted. Out previous/next page buttons immediately reload data. For me, these are 2 different things, but if you insist, I can hide arrows in input.

I don't insist 😄  It was just something I wanted to point out to maybe give another thought, but I see the difference and it can definitely stay that way.

The other things I mentioned are something I thought could help intuitivity, but in the end, as you say, it depends on how they are treated elsewhere to keep consistency.

libs/perun/utils/src/lib/table-wrapper/table-wrapper.component.ts
```diff
@@ -62,4 +62,33 @@ export class TableWrapperComponent implements OnInit {
...
+    };
+    this.paginator.page.next(changePageEvent);
+
+    input.value = selectedPage.toString();
+  }
+
+  onlyValidKeys(event: KeyboardEvent): boolean {
+    const charCode = event.key.charCodeAt(0);
+    return (
+      charCode === 65 || charCode === 66 || charCode === 68 || (charCode >= 48 && charCode <= 57)
```
xflord, line 91: Either use literals or describe what this does in a comment

libs/perun/utils/src/lib/table-wrapper/table-wrapper.component.ts
```diff
@@ -62,4 +62,33 @@ export class TableWrapperComponent implements OnInit {
       }); // scroll to top of table on page changes
     }
   }
+
+  changePage(event: FocusEvent): void {
+    const input = event.target as HTMLInputElement;
+    let selectedPage = parseInt(input.value);
+
+    if (!selectedPage || selectedPage < 1) {
+      selectedPage = 1;
```
xflord, line 71: Staying on the current page makes more sense to me than going to 1, no?

libs/perun/utils/src/lib/table-wrapper/table-wrapper.component.ts
xflord: Also why are A B and D allowed? I guess for some shortcuts? In that case allow them in combination with `CTRL`? This whole check seems unnecessary, IMO resetting the input if invalid value was put in would be enough, but up to you.

xflord (approved):

xflord (changes requested):
missclick

xflord (approved):
LGTM

bodnara (commented):
With the media query for smaller displays the number of page is not visible.
image

We can do confirm with enter `(keyup.enter)="changePage($event)"`

michalberky (approved):

sarkapalkovicova:
> With the media query for smaller displays the number of page is not visible. image
> 
> We can do confirm with enter `(keyup.enter)="changePage($event)"`

It seems to be different between browsers. I extended the input a little bit.

bodnara (approved):

PR #1426, opened 2023-09-19
feat(admin): change GUI to incorporate moduleClassNames

michalberky:
* Changed all uses of moduleClassName to moduleClassNames
* Added functionality to add multiple module names in 'Change application form settings' dialog

These changes are a reaction to this https://github.com/CESNET/perun/pull/4060
Do not merge without merging the original PR.


apps/admin-gui/src/app/shared/components/dialogs/update-application-form-dialog/update-application-form-dialog.component.html
```diff
@@ -5,10 +5,30 @@
...
-        <input matInput [(ngModel)]="moduleName" />
-      </mat-form-field>
+      <div
+        *ngFor="let moduleName of moduleNames; let i = index; trackBy: trackByIndex"
+        class="d-flex flex-row align-items-center">
+        <mat-form-field class="w-100 mb-n2">
+          <mat-label>{{'DIALOGS.UPDATE_APPLICATION_FORM.MODULE_NAME' | translate}}</mat-label>
+          <input [(ngModel)]="moduleNames[i]" matInput />
+        </mat-form-field>
+        <button
```
xflord, line 15: I would prefer for the buttons to be green and red, to match how it is with policy collections 
Screenshot from 2023-09-20 13-17-17

apps/admin-gui/src/app/shared/components/dialogs/update-application-form-dialog/update-application-form-dialog.component.html
```diff
@@ -5,10 +5,30 @@
...
-      <mat-form-field>
-        <mat-label>{{'DIALOGS.UPDATE_APPLICATION_FORM.MODULE_NAME' | translate}}</mat-label>
-        <input matInput [(ngModel)]="moduleName" />
-      </mat-form-field>
+      <div
+        *ngFor="let moduleName of moduleNames; let i = index; trackBy: trackByIndex"
+        class="d-flex flex-row align-items-center">
+        <mat-form-field class="w-100 mb-n2">
+          <mat-label>{{'DIALOGS.UPDATE_APPLICATION_FORM.MODULE_NAME' | translate}}</mat-label>
+          <input [(ngModel)]="moduleNames[i]" matInput />
```
xflord, line 13: theoretically the character `,` should not be allowed in the input fields, since the backend uses it to split the module names. However it should not  be a problem since module names don't usually contain it anyways. (But it causes a slightly weird behaviour: 
Screenshot from 2023-09-20 13-21-39
Before refresh of the page 
Screenshot from 2023-09-20 13-22-02
after refresh

xflord (approved):

mattjoke (approved):
Looks good :+1:

apps/admin-gui/src/app/shared/components/dialogs/update-application-form-dialog/update-application-form-dialog.component.ts
```diff
@@ -34,21 +34,45 @@ export class UpdateApplicationFormDialogComponent implements OnInit {
...
+
+  trackByIndex(index: number): number {
+    return index;
+  }
+
+  preventComma(event: KeyboardEvent): void {
+    if (event.key === ',') {
+      event.preventDefault();
+    }
+  }
```
bodnara, line 66: Would prefer if we showed error on the mat form that says commas are not allowed.

michalberky:
`update-application-form-dialog` component has been reworked to utilize FormControl and now no longer prevents the user from typing a comma, but instead shows an error and disables the submit button

apps/admin-gui/src/app/shared/components/dialogs/update-application-form-dialog/update-application-form-dialog.component.ts
```diff
@@ -34,21 +35,50 @@ export class UpdateApplicationFormDialogComponent implements OnInit {
...
+  removeModule(index: number): void {
+    this.formControls.splice(index, 1);
+  }
+
+  trackByIndex(index: number): number {
+    return index;
+  }
+
+  formValid(): boolean {
+    return this.formControls.every((control) => control.valid);
```
xflord, line 69: I believe a more "correct" solution would be to create a FormGroup and add a new control for each module. This way you could also just use FormGroup.invalid (when disabling the confirmation button )in the template instead of a function call (which we try to avoid since it's inefficient). See https://angular.io/api/forms/FormGroup and components like password-reset-form.

michalberky:
After running into a dead end trying to use FormGroup, without making the code too chaotic, it turned out that FormArray is the better alternative for this particular case with dynamic number of FormControls. So that is what I used in the end.

apps/admin-gui/src/app/shared/components/dialogs/update-application-form-dialog/update-application-form-dialog.component.ts
```diff
@@ -17,13 +18,13 @@ export interface UpdateApplicationFormDialogData {
...
   entity: string;
   applicationForm: ApplicationForm;
-  moduleName: string;
   initialState: string;
   extensionState: string;
   embeddedState: string;
   loading = false;
   theme: string;
   autoRegistrationEnabled: boolean;
+  formArray: FormArray<FormControl>;
```
bodnara, line 27: When using  form controls always define the type itself. This way we get help and type safety from typescript.
Make sure to include type when creating form controls in this file. Also remove any casting to string (`control.value as string`) as it is no longer needed.

Otherwise good job with the `FormArray`!
```suggestion
  formArray: FormArray<FormControl<string>>;
```

bodnara (approved):

xflord (approved):

PR #1428, opened 2023-09-21
fix(lib): hide logout alert

sarkapalkovicova:
While using service access, the logout alert is hidden after a page refresh.


mattjoke:
This doesn't seem to work. Whenever I log in with service-account I get the error that I can't access service-access with the service account (this is expected). It redirects me to the login page. The logout message is there (also expected). Nevertheless when I reload page it does not remove the message.

mattjoke (approved):

bodnara:
What is the meaning and difference of `baLogout` and `baAfterLogout`? I know you didn't introduce these variables and this solution should work. However it seems to me that the whole BA logic is divided and  this change maybe shouldn't be handled here ?

HejdaJakub (approved):
We agreed on a daily meeting that this implementation is OK.

PR #1431, opened 2023-09-25
feat(admin): update User section Perun Admin menu

xflord:
* 'Organizations' and 'Groups' pages were removed from the admin version of the 'User' section since information from both of these pages is contained in 'Accounts'
* 'Resources' and 'Facilities' pages were merged into new 'Assignments' page that works similarly to 'Accounts'
* Updated two entity attribute component to allow fixed second entity and did some necessary tweaks to allow displaying Member-Resource attributes per resource in fixed facility
* Fixed a bug where the initial selected object didn't match the displayed attributes on two-entity attributes page
* Aligned loading spinner on 'Accounts' page
* Updated openapi

DEPLOYMENT NOTE: Removed 'Organizations', 'Groups', 'Resources' and 'Facilities' pages from the admin version of the User section. Refer admins to 'Accounts' page for more clear info on VOs/Groups and to new 'Assignments' page for info on Facilities/Resources

### DIFF
--- a/apps/admin-gui/src/app/users/pages/user-detail-page/user-assignments/user-assignments.component.html
+++ b/apps/admin-gui/src/app/users/pages/user-detail-page/user-assignments/user-assignments.component.html
@@ -1,0 +1,64 @@
+<h1 class="page-title">{{'USER_DETAIL.ASSIGNMENTS.TITLE' | translate}}</h1>
+<ng-template #spinner1>
+  <perun-web-apps-loading-table></perun-web-apps-loading-table>
+</ng-template>
+<div *perunWebAppsLoader="initLoading; indicator: spinner1">
+  <div class="d-flex">
+    <perun-web-apps-facility-search-select
+      *ngIf="!(initLoading || facilities.length === 0)"
+      class="w-50"
+      [facilities]="facilities"
+      (facilitySelected)="loadFacility($event)">
+    </perun-web-apps-facility-search-select>
+  </div>
+  <mat-tab-group *ngIf="!(initLoading || facilities.length === 0)" mat-stretch-tabs="false">
+    <mat-tab>
+      <ng-template matTabLabel>
+        {{'USER_DETAIL.ASSIGNMENTS.OVERVIEW' | translate}}
+      </ng-template>
+      <ng-template matTabContent>
+        <p class="mt-4 subtitle">{{'USER_DETAIL.ASSIGNMENTS.RESOURCES' | translate}}</p>
+        <perun-web-apps-resources-list
+          *perunWebAppsLoader="loading; indicator: spinner1"
+          [resources]="resources"
+          [displayedColumns]="['id', 'name', 'vo', 'description']">
+        </perun-web-apps-resources-list>
+      </ng-template>
+    </mat-tab>
+    <mat-tab>
+      <ng-template matTabLabel>
+        {{'USER_DETAIL.ASSIGNMENTS.USER_FACILITY_ATT' | translate}}
+      </ng-template>
+      <ng-template matTabContent>
+        <mat-spinner class="ms-auto me-auto" *ngIf="loading"></mat-spinner>
+        <app-two-entity-attribute-page
+          *ngIf="!loading"
+          [firstEntity]="'user'"
+          [firstEntityId]="userId"
+          [secondEntity]="'facility'"
+          [specificSecondEntity]="selectedFacility"></app-two-entity-attribute-page>
+      </ng-template>
+    </mat-tab>
+
+    <mat-tab>
+      <ng-template matTabLabel>
+        {{'USER_DETAIL.ASSIGNMENTS.MEMBER_RESOURCE_ATT' | translate}}
+      </ng-template>
+      <ng-template matTabContent>
+        <mat-spinner class="ms-auto me-auto" *ngIf="loading"></mat-spinner>
+        <app-two-entity-attribute-page
+          *ngIf="!loading"
+          [firstEntity]="'user'"
+          [firstEntityId]="userId"
+          [secondEntity]="'resource'"
+          [facilityId]="selectedFacility.id"></app-two-entity-attribute-page>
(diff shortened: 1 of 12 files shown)


apps/admin-gui/src/app/users/pages/user-detail-page/user-assignments/user-assignments.component.ts
```diff
@@ -0,0 +1,63 @@
...
+  selectedFacility: Facility = null;
+  member: Member = null;
+  resources: Resource[] = [];
+  userId: number;
+  constructor(
+    private route: ActivatedRoute,
+    private usersService: UsersManagerService,
+    private membersService: MembersManagerService,
+    private facilityService: FacilitiesManagerService,
+    private resourceService: ResourcesManagerService
```
bodnara, line 32: Unused.
```suggestion
```

apps/admin-gui/src/app/users/pages/user-detail-page/user-assignments/user-assignments.component.ts
```diff
@@ -0,0 +1,63 @@
...
+  ) {}
+
+  ngOnInit(): void {
+    this.loading = true;
```
bodnara, line 36:
```suggestion
    this.initLoading = true;
```

apps/admin-gui/src/app/users/pages/user-detail-page/user-assignments/user-assignments.component.html
```diff
@@ -0,0 +1,62 @@
+<h1 class="page-title">{{'USER_DETAIL.ASSIGNMENTS.TITLE' | translate}}</h1>
+<mat-spinner class="ms-auto me-auto" *ngIf="initLoading"></mat-spinner>
```
bodnara, line 2: Prefer non blocking loading "indication", use template and `*perunWebAppsLoader` directive.

apps/admin-gui/src/app/users/pages/user-detail-page/user-assignments/user-assignments.component.html
```diff
@@ -0,0 +1,62 @@
...
+      {{'USER_DETAIL.ASSIGNMENTS.OVERVIEW' | translate}}
+    </ng-template>
+    <ng-template matTabContent>
+      <mat-spinner class="ms-auto me-auto" *ngIf="loading"></mat-spinner>
+      <p class="mt-4 subtitle">{{'USER_DETAIL.ASSIGNMENTS.RESOURCES' | translate}}</p>
+      <perun-web-apps-resources-list
+        *ngIf="!loading"
+        [resources]="resources"
+        [displayedColumns]="['id', 'name', 'vo', 'description']"
+        (refreshTable)="loadFacility(selectedFacility)">
```
bodnara, line 23: Not valid for this use case.
```suggestion
      <perun-web-apps-resources-list
        *ngIf="!loading"
        [resources]="resources"
        [displayedColumns]="['id', 'name', 'vo', 'description']">
```

bodnara (changes requested):
Please mirror page changes for side menu as well.

bodnara:
Merge after: https://github.com/CESNET/perun/pull/4096

xflord:
Removed from side menu, added loader directives and tweaked `initLoading`

apps/admin-gui/src/app/users/pages/user-detail-page/user-assignments/user-assignments.component.html
bodnara: There is no need to define multiple templates, if the indicator is the same. Also I think we only need this for the resource table ? The attribute component handles loading for itself. One important note when using the directive parent must have position relative, so the indicator is contained within component.

apps/admin-gui/src/app/users/pages/user-detail-page/user-assignments/user-assignments.component.ts
```diff
@@ -0,0 +1,61 @@
...
+  loading = false;
+  facilities: Facility[] = [];
+  selectedFacility: Facility = null;
+  member: Member = null;
+  resources: Resource[] = [];
+  userId: number;
+  constructor(
+    private route: ActivatedRoute,
+    private usersService: UsersManagerService,
+    private membersService: MembersManagerService,
```
bodnara, line 29:
```suggestion
```

apps/admin-gui/src/app/users/pages/user-detail-page/user-assignments/user-assignments.component.html
bodnara: @xflord

xflord: Sorry, didn't notice this comment. I believe it's better to have the spinner display for every tab, since the attribute components start loading only after the facility loads (so should the user switch facility via the select and that call was slow, no loading indicator would be shown). It looks better now that I used only one template before the tab content, thanks. Also made the same changes for the accounts page (the improvements are more visible there if you want to check)

bodnara: You are right I did not think about that case. That being said right know if you switch facility while on the attributes tab it is not reloaded. Also with the loading changes on accounts page when initializing member is null error is present.

One other thing I still think we don't need spinner1/2/3/. The purpose of template is that it can be reused. You can just define one at the beginning/end of the file and use it as many times as you want. Or I am missing something here ?

xflord: You are right, I didn't quite understand how templates work. Also I assumed the loader directive worked similarly to ngIf but that's not the case, hence why the attributes didn't reload. Reverted to old spinners for attributes (and whole accounts page)

bodnara (approved):

HejdaJakub (approved):

PR #1434, opened 2023-10-01
feat(admin): attribute definition detail redesign

kofzera:
- Unique, Write critical and Read critical toggles changed to checkboxes
- All checkboxes have tooltips instead of the infobox below
- Global critical checkboxes hidden for entityless attributes


bodnara (approved):
I would prefer if the tooltips where shown on the right so they don't overlap other options.

michalberky (approved):
Looks good. Tooltips positioned _after_ work great in mobile view as well

HejdaJakub (approved):

PR #1435, opened 2023-10-02
test(admin): fix failing tests

HejdaJakub:
* Previously changed table header is now higher than before which (in my opinion) cause that some buttons are now out of visible screen for cypress - the click event must be forced for these buttons.


bodnara (approved):
Would it make sense to always do force click in tests for the future ?

xflord (approved):

HejdaJakub:
I agree that we want to avoid force-clicking, but I tried a few weeks ago to find another solution and I failed. At least the force click is not applied for disabled buttons which would be a problem if it worked this way.

PR #1438, opened 2023-10-03
feat(admin): filter groups by role

sarkapalkovicova:
Enables filter groups by role and its assignment type on vo-groups, group-subgroups, and member-groups page.


apps/admin-gui/src/app/vos/pages/group-detail-page/group-subgroups/group-subgroups.component.ts
```diff
@@ -157,4 +174,46 @@ export class GroupSubgroupsComponent implements OnInit {
...
+    if (this.selectedRoleTypes.length === this.roleTypesList.length) {
+      return 'ALL';
+    }
+    const types: string[] = this.roleTypes.value as string[];
+    if (types) {
+      return `${types[0]}  ${
+        types.length > 1
+          ? '(+' + (types.length - 1).toString() + ' ' + (types.length === 2 ? 'other)' : 'others)')
+          : ''
+      }`;
```
HejdaJakub, line 203: This is not needed when there are just 2 possible options, so it can be `ALL` or something like `types[0] + ' 1 other'`.

apps/admin-gui/src/app/vos/pages/group-detail-page/group-subgroups/group-subgroups.component.ts
```diff
@@ -48,6 +49,14 @@ export class GroupSubgroupsComponent implements OnInit {
     startWith(true)
   );
 
+  roles = new UntypedFormControl();
+  rolesList = ['GROUPADMIN', 'GROUPOBSERVER', 'GROUPMEMBERSHIPMANAGER'];
+  selectedRoles: string[] = [];
+
+  roleTypes = new UntypedFormControl();
```
HejdaJakub, line 56: Maybe it can be typed.
```suggestion
  roles = new FormControl(['']);
  rolesList = ['GROUPADMIN', 'GROUPOBSERVER', 'GROUPMEMBERSHIPMANAGER'];
  selectedRoles: string[] = [];

  roleTypes = new FormControl(['']);
```

HejdaJakub (changes requested):
The comments that I wrote for `group-subgroups` are relevant also for `member-groups` and `vo-groups` components.

xflord:
merge after https://github.com/CESNET/perun/pull/4105

xflord (changes requested):
Please extract the filter into a shared component

apps/admin-gui/src/app/vos/pages/group-detail-page/group-subgroups/group-subgroups.component.html
```diff
@@ -25,6 +25,37 @@ <h1 class="page-subtitle">{{'GROUP_DETAIL.SUBGROUPS.TITLE' | translate}}</h1>
...
+        *ngFor="let role of rolesList"
+        [value]="role"
+        >{{role | groupRole}}</mat-option
+      >
+    </mat-select>
+  </mat-form-field>
+
+  <mat-form-field class="filter-field me-2 fix-select">
+    <mat-label>{{'GROUP_DETAIL.SUBGROUPS.FILTER_ROLE_TYPES' | translate}}</mat-label>
+    <mat-select (closed)="changeRoleTypes()" [formControl]="roleTypes" multiple>
```
xflord, line 46:
```suggestion
    <mat-select (selectionChange)="changeRoleTypes()" [formControl]="roleTypes" multiple>
```
Only refresh page when the chosen roles change. This should do the job I believe

HejdaJakub (approved):
Everything works great for me.

I am now just wondering if this is maybe an ideal candidate for applying the "advanced filter" view similar for the assigned users page. Filter role, role assignment, and maybe also toggle for tree/list view can be hidden under the Filter button and it can be useful to have also the clear filters there.
image
image


We can discuss it later with other team members and we can also create a separate task for that.

mattjoke (approved):
Looks good :+1:

apps/admin-gui/src/app/shared/components/role-filter-inputs/group-roles-filter.component.html
```diff
@@ -0,0 +1,29 @@
+<mat-form-field class="me-2 fix-select">
+  <mat-label>{{'SHARED.COMPONENTS.GROUP_ROLES_FILTER.FILTER_ROLES' | translate}}</mat-label>
+  <mat-select (selectionChange)="changeRoles()" [formControl]="roles" multiple>
+    <mat-select-trigger>
+      {{displaySelectedRoles() | groupRole}}
```
bodnara, line 5: The logic of `displaySelectedRoles` should be inside of the pipe. Only pass in the list of selected roles.
Same goes for the other method + pipe combo.

apps/admin-gui/src/app/shared/components/role-filter-inputs/group-roles-filter.component.html
```diff
@@ -0,0 +1,29 @@
...
+    </mat-select-trigger>
+    <mat-option
+      class="vo-theme"
```
bodnara, line 8: Don't think it should be static vo-theme, since it is used on pages of other objects.
```suggestion
      class="{{theme}}"
```

apps/admin-gui/src/app/shared/components/role-filter-inputs/group-roles-filter.component.ts
```diff
@@ -0,0 +1,49 @@
...
+})
+export class GroupRolesFilterComponent {
+  @Output()
+  filterRoles = new EventEmitter<string[]>();
+  @Output()
+  filterRoleTypes = new EventEmitter<string[]>();
+  roles = new FormControl(['']);
+  roleTypes = new FormControl(['']);
+  rolesList = ['GROUPADMIN', 'GROUPOBSERVER', 'GROUPMEMBERSHIPMANAGER'];
+  roleTypesList = ['DIRECT', 'INDIRECT'];
```
bodnara, line 17: As mentioned would prefer if this were not strings.

bodnara (changes requested):
Nice job overall.

This is maybe more for the https://github.com/CESNET/perun/pull/4105 PR but I think it would be better if the roles and their assignment types were not just plain strings. This way we get all the good stuff from Typescript.

apps/admin-gui/src/app/shared/components/role-filter-inputs/group-roles-filter.component.ts
xflord: https://github.com/CESNET/perun/pull/4105#discussion_r1345294470

bodnara (approved):

xflord (approved):

PR #1439, opened 2023-10-04
refactor(admin): unify complementaryObject parameter

HejdaJakub:
* Regenerate openapi with new comments about supported objects for complementaryObjet(s) parameter.
* Unify a type of complementaryObject variable to PerunBean object across classes in admin-gui.
* The information about supported objects has been left just in openapi documentation to avoid future need of changing it on several places.
* Little refactor of edited files has been done.

### DIFF
--- a/apps/admin-gui/src/app/shared/components/dialogs/add-role-dialog/add-role-dialog.component.ts
+++ b/apps/admin-gui/src/app/shared/components/dialogs/add-role-dialog/add-role-dialog.component.ts
@@ -15,6 +15,7 @@ import {
 import { SelectionModel } from '@angular/cdk/collections';
 import { ToEnrichedFacilityPipe } from '@perun-web-apps/perun/pipes';
 import { DebounceFilterComponent } from '@perun-web-apps/perun/components';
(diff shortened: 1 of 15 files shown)


apps/admin-gui/src/app/shared/components/dialogs/add-role-dialog/add-role-dialog.component.ts
```diff
@@ -74,11 +75,19 @@ export class AddRoleDialogComponent implements OnInit {
...
+      },
+      error: () => (this.loading = false),
+    });
     this.resourceService.getAllResources().subscribe({
-      next: (resources) => (this.resources = resources),
+      next: (resources) => {
+        this.resources = resources;
+        this.loading = false;
+      },
+      error: () => (this.loading = false),
```
xflord, line 90: The API calls have to be nested into each other in order for loading to correctly end when the last call ends, no?

HejdaJakub: Instead of basic subscription nesting, I reworked it with rxjs `combineLatestWith` function.

PR #1441, opened 2023-10-04
fix(admin): use renamed html tag for application dynamic list

HejdaJakub:
* Due to recent move of this component from admin-gui folder to the lib folder was changed the prefix of the component, but it was not changed on member-applications page.


michalberky (approved):
Works

PR #1442, opened 2023-10-05
fix(admin): display groups on recently viewed dashboard

HejdaJakub:
* There was just a technical problem of matching type RichGroup/Group after some recent changes of loading groups.
* This matching was fixed on the dashboard and also in the related service.


mattjoke (approved):
Looks good :+1:

michalberky (approved):
Looks good to me.

PR #1443, opened 2023-10-05
fix(admin): fix application form page spinners

michalberky:
* Divided the 'loading' variable into two separate variables handling the loading of header and table individually.

### DIFF
--- a/apps/admin-gui/src/app/vos/pages/group-detail-page/group-settings/group-settings-application-form/group-settings-application-form.component.html
+++ b/apps/admin-gui/src/app/vos/pages/group-detail-page/group-settings/group-settings-application-form/group-settings-application-form.component.html
@@ -1,9 +1,9 @@
 <h1 class="page-subtitle">
   {{'GROUP_DETAIL.SETTINGS.APPLICATION_FORM.TITLE' | translate}}
 </h1>
-<mat-spinner *ngIf="loading" class="ms-auto me-auto"> </mat-spinner>
+<mat-spinner *ngIf="loadingHeader" class="ms-auto me-auto"> </mat-spinner>
 <div *ngIf="!noApplicationForm">
-  <div *ngIf="!loading" class="d-flex w-75">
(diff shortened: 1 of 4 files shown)


apps/admin-gui/src/app/vos/pages/group-detail-page/group-settings/group-settings-application-form/group-settings-application-form.component.html
```diff
@@ -1,9 +1,9 @@
 <h1 class="page-subtitle">
   {{'GROUP_DETAIL.SETTINGS.APPLICATION_FORM.TITLE' | translate}}
 </h1>
-<mat-spinner *ngIf="loading" class="ms-auto me-auto"> </mat-spinner>
+<mat-spinner *ngIf="loadingHeader" class="ms-auto me-auto"> </mat-spinner>
 <div *ngIf="!noApplicationForm">
-  <div *ngIf="!loading" class="d-flex w-75">
+  <div *ngIf="!loadingHeader" class="d-flex w-75">
```
xflord, line 6:
```suggestion
  <div **perunWebAppsLoader="loadingTable; indicator: spinner" class="d-flex w-75">
```
You can theoretically do this and reuse the spinner template (if you define it at the top instead of before the app-application-form-list. Would probably be more future-proof in order we do some global changes to the loading directive, but just a nitpick (I don't understand the directive that much)

xflord (approved):

HejdaJakub (approved):

apps/admin-gui/src/app/vos/pages/group-detail-page/group-settings/group-settings-application-form/group-settings-application-form.component.html
bodnara: In case of header of the page the directive is not well suited. The main benefit of `perunWebAppsLoader` is that it does not hide all the content from users. It just places the provided loading indicator as an overlay. But here during initial loading there is no content to show since data has not yet been loaded.

PR #1445, opened 2023-10-09
fix(admin): edit variable for sidemenu submenu active text color

HejdaJakub:
* Fix a misspelling in cssVariable name for side_submenu_active_text_color variable.


HejdaJakub:
Not all configs have been tested yet, do not merge before completed testing.

bodnara (approved):
Ideally this would not be just a plain string so typescript would catch these mistakes in the future. As a fix it it sufficient.

mattjoke (approved):
Looks good :+1:

PR #1446, opened 2023-10-09
fix(admin): disable verify button instead of hiding it

HejdaJakub:
* Diasable button for not authorized users and provide an info tooltip.


bodnara (commented):
Generally if users don't have permission to use element we don't show it at all. It there a reason to not do it in this case ?

kofzera (approved):

HejdaJakub:
Removing is already working correctly (with corresponding low privileges), so there is no more need for this PR.

PR #1447, opened 2023-10-09
feat: added type to dialog config

bodnara:
Method for retrieving default dialog config will return typed config if data type is provided.


HejdaJakub (approved):

xflord (approved):
Maybe implement one or two cases where this is used to give people in the future an example?

bodnara:
> Maybe implement one or two cases where this is used to give people in the future an example?

Provided some examples. Most dialogs already have the type for its data defined and can be used directly. In case dialog does not expect any data the `never` type can be used.

xflord (approved):
thanks

PR #1449, opened 2023-10-12
refactor: get rid of UntypedFormBuilder

HejdaJakub:
* In admin-gui, user-profile and libs we can use typed FormBuilder instead of untyped since the FormBuilder can create typed FormGroup according to the default values of defined form group items.
* Some other refactoring was made in changed files (use PerunTranslateService, use non-deprecated type of subscribe).


HejdaJakub:
I didn't want to fix all untyped things related to the forms like usage of `UntypedFormControls` etc. and I also didn't want to fix the publications application. This PR is mainly to get rid of `UntypedFormBuilder` in our currently used apps.

bodnara (approved):
lgtm

PR #1450, opened 2023-10-16
feat: merge configs on initialization

xflord:
* values from configs (default, instance, brandings) are now merged when loading instead of fetching them on each 'getProperty()' call


xflord:
Not sure whether these changes warrant a deployment note

HejdaJakub (approved):
Since this should work exactly the same on the production as up to now and no changes are needed, I think the deployment note is unnecessary.

mattjoke (approved):

PR #1452, opened 2023-10-19
feat(admin): color of block destinations

sarkapalkovicova:
Distinguish block destinations by color on the services list.


xflord (approved):
looks a bit like the table is loading when all services are blocked, but that's probably fine, couldn't think of a better color

HejdaJakub (approved):

PR #1455, opened 2023-10-25
fix(admin): fix ExpressionChangedAfterItHasBeenCheckedError on applic…

xflord:
…ations page

* correctly set loading before loading of preferences and update after changes in dynamic list


xflord:
> This fixes the error, but still, there are unnecessary multiple calls to the backend. Thanks to your solution I found out another possible one, so here is my suggestion: In the `group-applications.component.html` and `vo-applications.component.html` edit the behavior with `*ngIf` directive - it will decrease calls of `ngOnChanges` hook. image

This would also require a loading indicator, since the call for viewPreferences attribute might be stalled for some reason. In which case you cannot use *ngIf (without a wrapper which sometimes causes the elements to jump around). And using our loading directive directly on an element doesn't cause the element to disappear when the controlling variable is false (which is my biggest issue with the directive compared to previous `*ngIf=loading` approach)

PR #1456, opened 2023-10-26
fix(admin): disable the force propagation button

sarkapalkovicova:
If a blocked service is selected, then the force propagation button is disabled.


xflord (commented):
Looks to work fine, but I noticed that the table does not refresh when blocking/allowing services, which to me it makes sense it would - to be consistent with all other table actions. Could please fix this within this PR?

michalberky (approved):
Works as it should, including the refresh

mattjoke (approved):
Looks good :+1: , fix Eslint please

HejdaJakub (approved):

PR #1461, opened 2023-11-03
fix(profile): removed unused calls

mattjoke:
- removed unused calls for enitityless attributes
- fixed no-categories refresh bug

### DIFF
--- a/libs/perun/services/src/lib/mfa-api.service.ts
+++ b/libs/perun/services/src/lib/mfa-api.service.ts
@@ -19,6 +19,7 @@ import { AttributesManagerService } from '@perun-web-apps/perun/openapi';
 })
 export class MfaApiService {
   mfaApiUrl = this.store.getProperty('mfa').api_url;
+
   constructor(
     private store: StoreService,
     private oauthService: OAuthService,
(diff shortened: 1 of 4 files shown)


HejdaJakub (changes requested):
I believe this way it should work correctly (tested for e-infra instance). Just a few suggestions.

It's really confusing to still have the `enforce_mfa_attribute` config property since the attribute stored in this property does not exist anymore and we are using just the part, where the instance type is defined. I would definitely replace this property with another more clear one (if we do this, the commit shall be marked as BREAKING CHANGE then). 

The `saveSettings` method also has an unused parameter `newEnforce`.

libs/perun/services/src/lib/mfa-api.service.ts
```diff
@@ -125,9 +129,10 @@ export class MfaApiService {
   saveDetailSettings(settings: MfaSettings): void {
     let allTrue = false;
     let allFalse = true;
+    const categroriesLength = Object.keys(settings.categories).length;
```
xflord, line 132: typo

xflord (approved):

HejdaJakub (approved):

PR #1462, opened 2023-11-03
fix(admin): edit attribute definition

sarkapalkovicova:
- Enable removal of existing attribute policy collections.


HejdaJakub (commented):
It works, but it sounds a little hacky to me. First of all, I see the problem on the API layer, because there isn't defined that an empty array is not allowed - but the implementation relies on at least one object in that array (attr def ID is read on BE from that object).

In the end, it would require some changes on FE and mainly on BE (also in the current API), so from my side, we can use some workaround. But I would prefer to leave these changes and just replace an empty array of collections when the API is called (and add a comment here, why it is necessary). It can be achieved in `edit-attribute-definition-dialog` (note: just here, because for creation it is not needed) in `onSubmit` when the `setAttributePolicyCollections` method is called. We can use something like this:
```
emptyCollections: AttributePolicyCollection[] = [
   {
      id: -1,
      attributeId: this.data.attDef.id,
      action: AttributeAction.READ,
      policies: [],
   },
];
...
...
...
// COMMENT
this.attributesManager.setAttributePolicyCollections(
   collections.length > 0
   ? { policyCollections: collections }
   : { policyCollections: this.emptyCollections },
),
```

kofzera (approved):

HejdaJakub (approved):

